PR #201, opened 2012-12-04
Fix stub_const for case that const_missing defines target nested constant

sorah:
Added `const_get` before calling `const_defined` because `Module#const_missing` may
define target constant.

Representive case is a Ruby on Rails (`ActiveSupport`),
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/dependencies.rb#L176

In test in a Rails app, you'll get an error if you used `stub_const` for
controllers, models, helpers, etc before it get loaded by
`ActiveSupport::Dependencies::ModuleConstMissing#const_missing`.
1. `stub_const "FooController::X", :foo`
2. `rspec-mocks` defines `FooController` as empty module
3. call `FooController`
4. Normally, const_missing has called and Rails loads the controller,
   but `rspec-mocks` defined `FooController` in `stub_const`,
   so controller won't be loaded.
5. You'll get an error.

### DIFF
--- a/lib/rspec/mocks/mutate_const.rb
+++ b/lib/rspec/mocks/mutate_const.rb
@@ -28,6 +28,12 @@ module RSpec
       #   - C.constants(false) #=> []
       if Module.method(:const_defined?).arity == 1
         def const_defined_on?(mod, const_name)
+          begin
+            mod.const_get(const_name)
+          rescue NameError
+            return false
+          end
+
           mod.const_defined?(const_name)
         end
 
@@ -44,6 +50,12 @@ module RSpec
         end
       else
         def const_defined_on?(mod, const_name)
+          begin
+            mod.const_get(const_name)
+          rescue NameError
+            return false
+          end
+
           mod.const_defined?(const_name, false)
         end
 
--- a/spec/rspec/mocks/mutate_const_spec.rb
+++ b/spec/rspec/mocks/mutate_const_spec.rb
@@ -12,9 +12,26 @@ class TestClass
   end
 end
 
+class TestConstMissingClass
+  def self.const_missing(name)
+    if name == :Missing
+      klass = Class.new do
+        def foo
+          :bar
+        end
+      end
+      const_set :Missing, klass
+      return klass
+    else
+      super
+    end
+  end
+end
+
 class TestSubClass < TestClass
   P = :p
 end
+
 
 module RSpec
   module Mocks
@@ -309,6 +326,17 @@ module RSpec
 
         context 'for an unloaded unnested constant' do
           it_behaves_like "unloaded constant stubbing", "X"
+        end
+
+        context 'for an dynamically defined nested constant' do
+          before do
+            stub_const "TestConstMissingClass::Missing::X", :foo
+          end
+
+          it "uses dynamically defined class" do
+            TestConstMissingClass::Missing::X.should == :foo
+            TestConstMissingClass::Missing.new.foo.should == :bar
+          end
         end
 
         context 'for an unloaded nested constant' do


mockdeep:
@myronmarston et al. What about allowing a syntax that causes the autoloading to happen inline, e.g.:

```ruby
stub_const(Finder, '::PAGE_SIZE', 2)
```

myronmarston:
> @myronmarston et al. What about allowing a syntax that causes the autoloading to happen inline, e.g.:

Why complicate RSpec with a larger API when you can just load the class yourself?  In general, RSpec isn't concerned with how you load your code (that's entirely up to you) and I don't want RSpec to get in that business.

mockdeep:
I guess it seems like it would be nice to have an idiomatic way to handle stubbing autoloaded constants, as opposed to having something like this laying around our codebase:

```ruby
Finder # need to autoload because of the way RSpec stubs constants
stub_const('Finder::PAGE_SIZE', 2)
```

I haven't looked too closely at the code and I don't know your intentions with the api, but we could also pass it in as a third argument and instead require users to pass nesting in as an option:

```ruby
stub_const('PAGE_SIZE', 2, nested_under: 'Finder')
```

It seems like having the nesting and the name of the constant explicitly separate might help reduce the complexity of the implementation, but you'd know better than me on that.

I don't think it's the place of RSpec to pass judgement on the way you load your code, but I do think it should be aware of the contexts it is used and be accommodating to the more common use cases. Not a big deal one way or another. We'll probably just wrap it in a helper method on our end otherwise.

myronmarston:
If something depends upon `Finder`, why not declare that dependency at the top of the file with `require 'finder'`?

There's no need to involve autoloading at all.  Just require what you need when you need it.

mockdeep:
That's basically the same problem, it just puts a greater distance between the loading and the reason we're loading it:

```ruby
# we need to require this because of the way RSpec stubs constants
# look for `stub_const` below
require 'path/to/the/finder'

# ...later
stub_const('Finder::PAGE_SIZE', 2)
```

We can debate the pros and cons of autoloading vs manually requiring dependencies, but for better or worse a huge chunk of the ruby community uses autoloading (thanks, of course, to Rails). I can imagine as a library developer you've run into a fair share of autoloading nightmares, but as an app developer (mostly) I have to say it's been a huge convenience, and the times I've come up against it have been few and far between, not to mention easy to resolve.

xaviershay:
I'm still a Rails user, but do a lot of other non-Rails dev as well.

As a Rails user, I want to explicitly require dependencies in my spec file so it's obvious what I depend on.

I'm generally negative on autoloading since it makes it really easy to forget to care about your inter-dependencies. This library's express purpose is to help uncover and minimize dependencies, I feel like making auto-loading easier here is heading in the wrong direction. It's also potentially confusing against the verified double API (which acts different based on whether you require something or not).

I could _potentially_ imagine an API added to `rspec-rails` with the aim of supporting rails users better, so I'm probably negative on that suggestion as well.

I'd recommend if people _do_ think is a good idea, we start by providing it in a plugin. If it's popular, we can revisit bringing it into core. (Like what we did with verified doubles / RSpec fire.)

mockdeep:
I understand the motivation, though in our case we already do a pretty decent job of minimizing inter-dependencies, so manual `require`s would add process and effort without a whole lot in the way of useful architectural feedback. I'm also not sure I agree with the idea that using `require`s is the ideal way to put that sort of design pressure on an application, as opposed to more explicit static analysis checks with the likes of rubocop, or, of course, writing tests.

At any rate, as you might have imagined, I'm definitely in favor of supporting some sort of autoloading workflow, either at an individual api level, or even with a global configuration of some sort that can be toggled: `config.autoload_stuff = true`.

JonRowe:
Work around... `stub_const("#{Finder}::PAGE_SIZE", 2)`

mockdeep:
Nice thanks. I was thinking of doing something like that.

PR #467, opened 2013-11-13
Add allow(...).to receive_message_chain

fables-tales:
First pass. Comments on the back of a PR :)


stephan-nordnes-eriksen:
@myronmarston the `receive_message_chain` is really nice, but I don't see how useful it is if you can't actually, somehow, test the arguments. Would it not be possible to make the syntax something like this? `expect(obj).to receive(:method_name).with("data").receive(:method_two).with("data2").and_return("result")`

Example of usecase:

Say I want to test a single-fire object like `MessageClass.new("data").send_to("username")`. Of course you can use spy on the MessageClass and stub the `new` method, and then run an expect on the spy. But it is tedious and would be much nicer to have in a chain.

Also, there are probably some piping libraries that could benefit from this as well, think: `dataset.filter("whatever").order.first(20)`.

myronmarston:
@stephan-nordnes-eriksen -- please open a new issue.  Feature requests posted at the bottom of a merged pull request tend to get lost in the shuffle.

stephan-nordnes-eriksen:
Yea, I agree, but I don't want to create a feature request if it is something that won't be possible to implement.

PR #832, opened 2014-11-29
Add an api for customising how verifying doubles are checked

JonRowe:
This is a prerequisite for rspec/rspec-rails#1238

### DIFF
--- a/lib/rspec/mocks/object_reference.rb
+++ b/lib/rspec/mocks/object_reference.rb
@@ -57,6 +57,10 @@ module RSpec
               "Can not perform constant replacement with an object."
       end
 
+      def const
+        @object.class
+      end
+
       def defined?
         true
       end
@@ -75,6 +79,10 @@ module RSpec
         @object.name
       end
       alias description const_to_replace
+
+      def const
+        @object
+      end
     end
 
     # Used when a string is passed to `class_double`, `instance_double`
@@ -96,6 +104,10 @@ module RSpec
         @const_name
       end
       alias description const_to_replace
+
+      def const
+        object
+      end
 
       def when_loaded
         yield object if object
--- a/lib/rspec/mocks/object_reference.rb
+++ b/lib/rspec/mocks/object_reference.rb
@@ -57,7 +57,7 @@ module RSpec
               "Can not perform constant replacement with an object."
       end
 
-      def const
+      def target
         @object.class
       end
 
@@ -80,7 +80,7 @@ module RSpec
       end
       alias description const_to_replace
 
-      def const
+      def target
         @object
       end
     end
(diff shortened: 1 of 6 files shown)


JonRowe:
What does this need to finish off? I've forgotten.

myronmarston:
> What does this need to finish off? I've forgotten.

- [ ] Consider renaming `on_declaring_verifying_double` to `verifying_double_declaration_callbacks` (or at least make the ivar and method name align). (https://github.com/rspec/rspec-mocks/pull/832/files#r21209020)
- [ ] Doc the object/module reference classes that are now publicly exposed in the callback--you said you'd do this in a separate PR, which is fine as long as we get that in 3.2. (https://github.com/rspec/rspec-mocks/pull/832/files#r21211107)
- [ ] `DirectModuleReference#target` doesn't have a spec (although one fails that implicitly uses it, but I'd like an explicit one, like we have for the other `target` implementations).  To hit that case you need to pass an anonymous class or module to `instance_double` or `class_double`.  (https://github.com/rspec/rspec-mocks/pull/832#issuecomment-65351491) As discussed in that comment, we may want to rename `DirectModuleReference` to `AnonymousModuleReference` to accurately reflect when it is used.
- [ ] `construction_spec.rb:29` has `declared` misspelled as `delcared`.
- [ ] This PR is old enough that rebasing and getting new travis builds is probably worthwhile before merging.

A few little things, nothing major.

JonRowe:
I've done a few of the things, can you remind me where the specs for the other implementations are?

lib/rspec/mocks/object_reference.rb
```diff
@@ -70,11 +74,15 @@ def when_loaded
     # Represents a reference to that module.
     #
     # @private
-    class DirectModuleReference < DirectObjectReference
+    class AnonymousModuleReference < DirectObjectReference
```
myronmarston: The documentation comment for this class is still inaccurate ("Used when a module is passed to `class_double` or `instance_double`" -- it's actually only used when an anonymous module is passed) -- but that's not a merge blocker.  I'll plan to get to it post-merge if you don't get to it first.

myronmarston:
> can you remind me where the specs for the other implementations are?

It was simpler just to add the spec and push myself :).

lib/rspec/mocks/object_reference.rb
myronmarston: I took care of this while I was at it.

myronmarston:
Thanks for following through on this, @JonRowe.  I'm happy with this provided we don't forget to add docs for the reference classes that are now public in a separate PR before 3.2 ships.  I'll leave this for you to review and merge.

PR #841, opened 2014-12-16
Add failing specs for when `with` is combined with count constraints.

myronmarston:
When the message is also received with other arguments,
the other arguments are included in the failure message
rather than the invalid counts, which is confusing.

Note that when the message is _never_ received with the specified args (but is received with other args), it probably makes sense to keep the failure message how it is now (where it mentions the unexpected args), but when it is received with the specified args, I think it makes sense to mention the wrong count.  We may need to add some additional specs for this.

I discovered this when working on some rspec-core stuff.  I may come back to this later but for now I wanted to just open a PR with failing specs so we don't forget about it.  Anyone else can feel free to take it :).

### DIFF
--- a/spec/rspec/mocks/matchers/have_received_spec.rb
+++ b/spec/rspec/mocks/matchers/have_received_spec.rb
@@ -141,9 +141,38 @@ module RSpec
           let(:the_dbl) { double(:expected_method => nil) }
 
           before do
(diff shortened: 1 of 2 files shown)


myronmarston:
Closing in favor of #915.  (Hopefully someone will take it up at some point....)

PR #860, opened 2015-01-08
Fix inaccurate doc string in cucumber scenario.

myronmarston:
This was apparently a copy/paste error.

Thanks to @aruprakshit for notifying me of this problem.


JonRowe:
LGTM

PR #863, opened 2015-01-14
Docs for object references exposed by #832 callback

JonRowe:
This expands the documentation of the objects now exposed to the public via #832 /cc @myronmarston

### DIFF
--- a/lib/rspec/mocks/object_reference.rb
+++ b/lib/rspec/mocks/object_reference.rb
@@ -41,45 +41,70 @@ module RSpec
 
     # Used when an object is passed to `object_double`.
     # Represents a reference to that object.
-    #
-    # @private
     class DirectObjectReference
+      # @param object to which this refers to.
       def initialize(object)
         @object = object
       end
 
+      # The object's description (via `#inspect`).
+      # @return String
       def description
         @object.inspect
       end
 
+      # Defined for interface parity with the other object reference
+      # implementations, it has no real relevance here.
       def const_to_replace
         raise ArgumentError,
               "Can not perform constant replacement with an object."
       end
 
+      # The target of the verifying double (the object's class).
+      #
+      # @return Class
       def target
         @object.class
       end
 
+      # Always returns true for an object as the class is defined.
+      #
+      # @return true
       def defined?
         true
       end
 
+      # Yields if the reference target is loaded, providing a generic mechanism
+      # to optionally run a bit of code only when a reference's target is
+      # loaded.
+      #
+      # This specific implementation always yields because direct references
+      # are always loaded.
+      #
+      # @yield object
       def when_loaded
         yield @object
       end
     end
 
-    # Used when an anonymous module is passed to `class_double` or `instance_double`.
-    # Represents a reference to that module.
-    #
-    # @private
+    # Used when an anonymous module is passed to `class_double` or
+    # `instance_double`. Represents a reference to that module.
     class AnonymousModuleReference < DirectObjectReference
+      # Returns the constant name to replace with a double.
+      #
+      # @return String the constant name
       def const_to_replace
         @object.name
       end
+
+      # Returns the module name.
+      #
+      # @return String the constant name
       alias description const_to_replace
 
+      # The target of the verifying double (the module).
+      #
+      # @return Module
       def target
         @object
       end
@@ -89,26 +114,43 @@ module RSpec
     # or `object_double`.
     # Represents a reference to the object named (via a constant lookup)
     # by the string.
-    #
-    # @private
     class NamedObjectReference
+      # @param const_name [String] constant name
       def initialize(const_name)
         @const_name = const_name
       end
 
+      # Returns true if the named constant is defined, false otherwise.
+      #
+      # @return Boolean
       def defined?
         !!object
       end
 
+      # Returns the constant name to replace with a double.
+      #
+      # @return String the constant name
       def const_to_replace
         @const_name
       end
+
+      # Returns the object constant name.
+      #
+      # @return String the constant name
       alias description const_to_replace
 
+      # The target of the verifying double (the object).
+      #
+      # @return object
       def target
         object
       end
 
+      # Yields if the reference target is loaded, providing a generic mechanism
+      # to optionally run a bit of code only when a reference's target is
+      # loaded.
+      #
+      # @yield object
       def when_loaded
         yield object if object
       end
--- a/lib/rspec/mocks/object_reference.rb
+++ b/lib/rspec/mocks/object_reference.rb
@@ -8,7 +8,7 @@ module RSpec
         case object_module_or_name
         when Module
           if anonymous_module?(object_module_or_name)
-            AnonymousModuleReference.new(object_module_or_name)
+            DirectObjectReference.new(object_module_or_name)
           else
             # Use a `NamedObjectReference` if it has a name because this
             # will use the original value of the constant in case it has
@@ -39,22 +39,26 @@ module RSpec
       private_class_method :anonymous_module?
     end
 
-    # Used when an object is passed to `object_double`.
+    # An implementation of rspec-mocks' reference interface.
+    # Used when an object is passed to {ExampleMethods#object_double}, or
+    # an anonymous class or module is passed to {ExampleMethods#instance_double}
+    # or {ExampleMethods#class_double}.
     # Represents a reference to that object.
+    # @see NamedObjectReference
     class DirectObjectReference
-      # @param object to which this refers to.
+      # @param object [Object] the object to which this refers
       def initialize(object)
         @object = object
       end
 
-      # The object's description (via `#inspect`).
-      # @return String
+      # @return [String] the object's description (via `#inspect`).
       def description
         @object.inspect
       end
 
       # Defined for interface parity with the other object reference
-      # implementations, it has no real relevance here.
+      # implementations. Raises an `ArgumentError` to indicate that `as_stubbed_const`
+      # is invalid when passing an object argument to `object_double`.
       def const_to_replace
         raise ArgumentError,
               "Can not perform constant replacement with an object."
@@ -62,14 +66,14 @@ module RSpec
 
       # The target of the verifying double (the object itself).
       #
-      # @return Object
+      # @return [Object]
       def target
         @object
       end
 
       # Always returns true for an object as the class is defined.
       #
-      # @return true
+      # @return [true]
       def defined?
         true
       end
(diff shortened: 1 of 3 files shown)


lib/rspec/mocks/object_reference.rb
```diff
@@ -90,25 +113,41 @@ def target
...
 
+      # The target of the verifying double (the object).
+      #
+      # @return object
       def target
         object
       end
 
+      # Performs the callback if the object is loaded
+      # loaded.
```
soulcutter: two loadeds in a row

soulcutter:
Not consistent about ending documentation with a period or not.

JonRowe:
Fixed :P

lib/rspec/mocks/object_reference.rb
```diff
@@ -42,44 +42,67 @@ def self.anonymous_module?(mod)
     # Used when an object is passed to `object_double`.
     # Represents a reference to that object.
     #
-    # @private
+    # @param object
```
myronmarston: What does `@param object` do attached to a class?  I thought this was only valid for methods...

I'm running the doc server locally and not seeing this rendered (plus it would need a description).

lib/rspec/mocks/object_reference.rb
```diff
@@ -42,44 +42,67 @@ def self.anonymous_module?(mod)
...
     class DirectObjectReference
       def initialize(object)
         @object = object
       end
 
+      # The objects desription (`#inspect`).
```
myronmarston: s/desription/description/
s/objects/object's/

And I think "via" would well (e.g. `(via #inspect)`)

lib/rspec/mocks/object_reference.rb
JonRowe: I (mistakenly obviously) thought that was how you tagged an initialise attribute

lib/rspec/mocks/object_reference.rb
```diff
@@ -42,44 +42,67 @@ def self.anonymous_module?(mod)
...
+      # @return String
       def description
         @object.inspect
       end
 
+      # Not relevant to a direct object reference.
```
myronmarston: It might be worth saying, "Defined for interface parity with the other object reference implementations." to explain why we have this at all.

lib/rspec/mocks/object_reference.rb
```diff
@@ -42,44 +42,67 @@ def self.anonymous_module?(mod)
...
         raise ArgumentError,
               "Can not perform constant replacement with an object."
       end
 
+      # The target of the verifying double (the object's class).
+      #
+      # @return Class
       def target
         @object.class
       end
```
myronmarston: Should this be implemented simply as `@object` (no `.class`)? In #832, we discussed how the name `target` was chosen because it returns the target of the reference -- that is, what the reference is a reference to, but in the `DirectObjectReference` case, the reference is to `@object`, not `@object.class`.  OTOH, if the purpose of `target` is to _always_ return a class or module, that's fine, but I think the name needs to reflect that.

In the case of ActiveRecord dynamic attributes (which drove the creation of this, right?) I don't think it matters because object doubles don't have the problem you are solving since `ActiveRecord::Base#respond_to?` always supports dynamic column methods even though `ActiveRecord::Base.method_defined?` doesn't.

Sorry to bring up the naming thing we had settled before but I worry that we'll be annoyed by this inconsistency in the future and it's nice to get it right before 3.2 ships.

lib/rspec/mocks/object_reference.rb
```diff
@@ -42,44 +42,67 @@ def self.anonymous_module?(mod)
...
 
+      # Always returns true for an object as the class is defined.
+      #
+      # @return true
       def defined?
         true
       end
 
+      # Performs the callback straight away as the object is already
+      # loaded.
```
myronmarston: This doc (and the others here) describes the specific implementation of this method rather than explaining its use as part of the generic reference interface (of which there are multiple implementations).  I think it might be better to document it with something like:

``` ruby
# Yields if the reference target is loaded, providing a generic mechanism
# to optionally run a bit of code only when a reference's target is loaded.
# This specific implementation always yields because direct references
# are always loaded.
```

Thoughts?  Could be good to provide a similar treatment for the other methods, so that users who read the docs get some insight into the purpose of these methods.  Otherwise, `when_loaded` seems like a pretty pointless method...

lib/rspec/mocks/object_reference.rb
```diff
@@ -42,44 +42,67 @@ def self.anonymous_module?(mod)
...
+      #
+      # @yield object
       def when_loaded
         yield @object
       end
     end
 
-    # Used when an anonymous module is passed to `class_double` or `instance_double`.
+    # Used when a module is passed to `class_double` or `instance_double`.
```
myronmarston: The "an anonymous" part is important, because that's the only case when this is used. When a named module is passed to `class_double` or `instance_double`, this is not used, but this comment now makes it sound like it would be.

lib/rspec/mocks/object_reference.rb
```diff
@@ -42,44 +42,67 @@ def self.anonymous_module?(mod)
...
     # Represents a reference to that module.
     #
-    # @private
+    # @param module an instance of an anonymous module
     class AnonymousModuleReference < DirectObjectReference
+      # Returns the constant name to replace with a double.
+      #
+      # @return string the constant name
```
myronmarston: "String" should be capitalized to indicate the return type.

lib/rspec/mocks/object_reference.rb
```diff
@@ -42,44 +42,67 @@ def self.anonymous_module?(mod)
...
       def const_to_replace
         @object.name
       end
+
+      # Returns the module name.
+      #
+      # @return string the constant name
```
myronmarston: Same here -- "String" should be capitalized to indicate the return type.

lib/rspec/mocks/object_reference.rb
```diff
@@ -42,44 +42,67 @@ def self.anonymous_module?(mod)
...
       alias description const_to_replace
 
+      # The target of the verifying double (the module).
+      #
+      # @return Class
```
myronmarston: It could be a module, so I think this should be `@return Module`

lib/rspec/mocks/object_reference.rb
```diff
@@ -90,25 +113,41 @@ def target
     # Represents a reference to the object named (via a constant lookup)
     # by the string.
     #
-    # @private
+    # @param string constant name
```
myronmarston: Again, "String"

lib/rspec/mocks/object_reference.rb
```diff
@@ -90,25 +113,41 @@ def target
...
     class NamedObjectReference
       def initialize(const_name)
         @const_name = const_name
       end
 
+      # The definition state of the object.
```
myronmarston: I think "definition state" is a pretty vague, abstract description that might not mean anything to users.  (Plus it is using the same root -- define -- as the method name, so it's a bit like a dictionary defining a word with a different form of the same word).  What do you think of this instead?

``` ruby
# Returns true if the named constant is defined, false otherwise.
```

lib/rspec/mocks/object_reference.rb
```diff
@@ -90,25 +113,41 @@ def target
...
+      #
+      # @return Boolean
       def defined?
         !!object
       end
 
+      # Returns the constant name to replace with a double.
+      #
+      # @return string the constant name
```
myronmarston: Again, "String".

lib/rspec/mocks/object_reference.rb
myronmarston: One other thing that I think will really help clarify the purpose and use of these classes is a mention in the class docs of each implementation that each of these is an implementation of a common abstract interface, and then you can use YARD's `# @see AlternateClass` syntax to reference the other implementations.

lib/rspec/mocks/object_reference.rb
JonRowe: I don't think so, because we expect a class from it I think...

myronmarston: > I don't think so, because we expect a class from it I think...

But `NamedObjectReference#target` may not return a class.  Example:

``` ruby
class Application
  LOGGER = Logger.new(...)
end

# in a spec
object_double("Application::Logger").as_stubbed_const
```

In this case `object_double` is using a `NamedObjectReference` and `NamedObjectReference#target` will return the original value of `Application::LOGGER`, which is a `Logger` instance, not a class.  In contrast, the reference created by `object_double(Application::LOGGER)` will be a `DirectObjectReference` and will return the `Logger` class from `#target`.  Seems odd/inconsistent that calling `#target` on the references created by `object_double(Application::LOGGER)` and `object_double("Application::LOGGER")` gets you different things.

Also, `DirectObjectReference` currently provides no method to get the object it is referencing, which seems extremely odd.

myronmarston: Also, it's possible to go from an instance to a class (just call `.class`!) but it's not possible to go from a class to the original instance (unless you do gnarly object space stuff...) so converting this to its class loses information in the sense that the caller has no easy way to get the object from it.  OTOH, if we return the object from `target` and the user wants a class, then they can do:

``` ruby
target = reference.target
target = target.class unless Module === target
```

JonRowe: Modified

JonRowe:
Re-review @myronmarston ?

lib/rspec/mocks/object_reference.rb
```diff
@@ -41,45 +41,70 @@ def self.anonymous_module?(mod)
...
       end
 
+      # Defined for interface parity with the other object reference
+      # implementations, it has no real relevance here.
       def const_to_replace
         raise ArgumentError,
               "Can not perform constant replacement with an object."
       end
 
+      # The target of the verifying double (the object's class).
```
myronmarston: "The object's class" is no longer accurate.

lib/rspec/mocks/object_reference.rb
```diff
@@ -41,45 +41,70 @@ def self.anonymous_module?(mod)
...
+      #
+      # @return String the constant name
       def const_to_replace
         @object.name
       end
+
+      # Returns the module name.
+      #
+      # @return String the constant name
       alias description const_to_replace
```
myronmarston: I think YARD automatically docs aliases w/o needing doc comments (in fact, we hook into that for our matcher) aliases.  See http://rspec.github.io/documentation/3.2/rspec-expectations/RSpec/Matchers.html for an example. Each matcher lists its aliases using `(also: #some_alias)` and I believe it would do the same here but you'd have to check.

lib/rspec/mocks/object_reference.rb
JonRowe: Fixed.

myronmarston:
I've got a couple of small quibbles but it'll honestly be easier to just make the changes myself.  I'll push them to this PR shortly, if that's OK with you.

JonRowe:
Sure!

myronmarston:
OK, I pushed my changes.  A few things to note.

* I was surprised there wasn't a spec change when you changed `target` from `@object.class` to `@object`.  I discovered why in 7478671ecd12af18fb37cd0e273f7823931ad3ac...`have_attributes(:target => Object)` will pass an object that returns `Object` from `#target` or that returns an instance of `Object` from `#target`, because our composable matching logic uses `===` and `Object === any_object` is true.  In the future, if you want to specify that it must be the `Object` class (and not an instance), you could use `have_attributes(:target => eq(Object))` (or an alias of `eq`).  Regardless, I updated the spec to force it to be the object instance since that's what we want now.
* It's easy to make small mistakes with YARD that don't render properly.  For example, when you forget to wrap types in square brackets (as you did in several places) it doesn't interpret it as a type.  I make these mistakes sometimes, too.  My suggestion (and what I do) is to run `bin/yard server --reload` and to view the docs in the browser as you edit -- that way you're aware of how YARD renders it and not blindly trusting that you've written the syntax in the way that it expects.  It's basically like using TDD for docs in tems of instant feedback :).

JonRowe:
Noted; So are you happy with this now?

myronmarston:
Yes

JonRowe:
Merged.

PR #866, opened 2015-01-15
Refactor odd memoization.

myronmarston:
It is odd here because within the matcher, we call
`@expectation = expect`, and within `expect` it was
internally memoizing, which means `@expectation`
was being assigned twice. It’s also dubious to
do so because we want to minimize internal memoization
since a matcher can be re-used against a different
target.


myronmarston:
Anyone want to review this, @rspec/rspec ?

JonRowe:
LGTM

PR #868, opened 2015-01-16
Provide a clear error when received message args are mutated.

### LINKED ISSUE #865: spies don't tolerate changing arguments
```ruby
describe 'rspec-mocks' do
  it 'passes with a double' do
    d = double('d').as_null_object
    expect(d).to receive(:m).with(hash_including('offset' => 1))
    args = {}
    args['offset'] = 1
    d.m args
    args['offset'] = 2
    d.m args
  end

  it 'passes with a spy' do
    d = spy('d')
    args = {}
    args['offset'] = 1
    d.m args
    args['offset'] = 2   # it passes if you comment
    d.m args             # out these two lines
    expect(d).to have_received(:m).with(hash_including('offset' => 1))
  end
end
```

myronmarston:
...before a `have_received(...).with(...)` expectation.

Fixes #865.

### DIFF
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -15,6 +15,10 @@ module RSpec
     # invoked.
     MockExpectationAlreadyInvokedError = Class.new(Exception)
 
+    # Raised for situations that RSpec cannot support due to external mutations
+    # made externally on arguments that RSpec is holding onto to use for later comparisons.
+    CannotSupportArgMutationsError = Class.new(StandardError)
+
     # @private
     UnsupportedMatcherError  = Class.new(StandardError)
     # @private
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -163,12 +163,12 @@ module RSpec
         attr_reader :args
 
         def initialize(args)
-          @args = args
-          @original_hashes = args.map { |arg| [arg, hash_of(arg)] }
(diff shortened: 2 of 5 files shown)


lib/rspec/mocks/proxy.rb
```diff
@@ -149,7 +156,35 @@ def has_negative_expectation?(message)
...
-        @messages_received << [message, args, block]
+        @messages_received << [message, ReceivedArgList.new(args), block]
+      end
+
+      class ReceivedArgList
+        attr_reader :args
+
+        def initialize(args)
+          @args = args
+          @original_hashes = args.map { |arg| [arg, hash_of(arg)] }
```
jferris: `Array#hash` will return a `hash` based on its components, so I think you can get away with just doing `@original_hash = @args.hash` and then comparing that later on.

myronmarston: You're right.  Thanks for pointing that out!

JonRowe:
Nice work :)

lib/rspec/mocks/error_generator.rb
```diff
@@ -15,6 +15,10 @@ module Mocks
     # invoked.
     MockExpectationAlreadyInvokedError = Class.new(Exception)
 
+    # Raised for situations that RSpec cannot support due to external mutations
```
waterlink: "due to external mutations made externally" - sounds a bit odd to me: "external" part is doubled.

PR #869, opened 2015-01-16
Make and_wrap_original work with any_instance

### LINKED ISSUE #867: and_wrap_original breaks in any_instance context
Example:

``` ruby
allow_any_instance_of(String).to receive(:upcase).and_wrap_original { |method| method.call }
```

Causes the error:

```
NoMethodError: undefined method `and_wrap_original' for #<RSpec::Mocks::AnyInstance::StubChain:0x007ff9facdbdc8>
```

The following monkeypatches make my specs work as expected, but I don't know if there are other places that also need to be fixed for it to be a complete solution:

``` ruby
class RSpec::Mocks::AnyInstance::StubChain
  prepend Module.new {
    def invocation_order
      @invocation_order ||= super.merge(
        and_wrap_original: [:with, nil]
      )
    end
  }
end

RSpec::Mocks::AnyInstance::Chain::Customizations.record :and_wrap_original
```

rf-:
There were a couple of different issues preventing `and_wrap_original`
from working when used with `any_instance`. Both of these changes were
necessary to make the test case pass, and I wasn't able to find any
other code that looked like it needed to be updated (except for
`receive_message_chain`, which [apparently doesn't work] with
`and_call_original` either).

This fixes #867.

[apparently doesn't work]: https://github.com/rspec/rspec-mocks/pull/467#issuecomment-28631621

### DIFF
--- a/spec/rspec/mocks/and_wrap_original_spec.rb
+++ b/spec/rspec/mocks/and_wrap_original_spec.rb
@@ -10,54 +10,68 @@ RSpec.describe "and_wrap_original" do
 
     let(:instance) { klass.new }
 
-    it "allows us to modify the results of the original method" do
-      expect {
-        allow(instance).to receive(:results).and_wrap_original do |m|
-          m.call.first(10)
-        end
-      }.to change { instance.results.size }.from(100).to(10)
+    shared_examples "using and_wrap_original" do
+      it "allows us to modify the results of the original method" do
+        expect {
+          allow_it.to receive(:results).and_wrap_original do |m|
+            m.call.first(10)
+          end
+        }.to change { instance.results.size }.from(100).to(10)
+      end
+
+      it "raises a name error if the method does not exist" do
+        expect {
+          allow_it.to receive(:no_results).and_wrap_original { |m| m.call }
+          instance.no_results
+        }.to raise_error NameError
+      end
+
+      it "passes in the original method" do
+        value = nil
+        original_method = instance.method(:results)
+        allow_it.to receive(:results).and_wrap_original { |m| value = m }
+        instance.results
+        expect(value).to eq original_method
+      end
+
+      it "passes along the message arguments" do
+        values = nil
+        allow_it.to receive(:results).and_wrap_original { |_, *args| values  = args }
+        instance.results(1, 2, 3)
+        expect(values).to eq [1, 2, 3]
+      end
+
+      it "passes along any supplied block" do
+        value = nil
+        allow_it.to receive(:results).and_wrap_original { |&b| value = b }
+        instance.results(&(block = proc {}))
+        expect(value).to eq block
+      end
+
+      it "ignores previous stubs" do
+        allow_it.to receive(:results) { "results" }
+        allow_it.to receive(:results).and_wrap_original { |m| m.call }
+        expect(instance.results).to_not eq "results"
+      end
+
+      it "can be constrained by specific arguments" do
+        allow_it.to receive(:results) { :all }
+        allow_it.to receive(:results).with(5).and_wrap_original { |m, n| m.call.first(n) }
+        expect(instance.results 5).to eq [1,2,3,4,5]
+        expect(instance.results).to eq :all
+      end
     end
 
-    it "raises a name error if the method does not exist" do
-      expect {
-        allow(instance).to receive(:no_results).and_wrap_original { |m| m.call }
-        instance.no_results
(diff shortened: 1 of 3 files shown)


rf-:
Re: `receive_message_chain`: [code that looks like it would need changing](https://github.com/rspec/rspec-mocks/blob/master/lib/rspec/mocks/matchers/receive_message_chain.rb#L14), [pending spec for `and_call_original`](https://github.com/rspec/rspec-mocks/blob/master/spec/rspec/mocks/matchers/receive_message_chain_spec.rb#L67-71). This isn't related to `any_instance`, but it came up when I was looking for other changes that might have been missed when `and_wrap_original` was added.

myronmarston:
Thanks for rolling up your sleeves and submitting this PR for your issue, @rf-!

Re: `receive_message_chain`, if you want to tackle making that work with `any_instance`, feel free but I recall it being fiendishly difficult (I could be wrong about that, though).  `any_instance` is a feature none of the rspec core devs actively use and we generally [discourage its use](https://relishapp.com/rspec/rspec-mocks/v/3-1/docs/working-with-legacy-code/any-instance)...so it's unlikely any of us will make it a priority to address these shortcomings.  Long term, I'm thinking that I'd like to actually move it out of rspec-mocks into an external gem for RSpec 4 so that it can continue to exist for those who use it without inflicting the large amount of complexity on rspec-mocks that it currently does.

Regarding this PR, my one concern is that your spec only covers one facet of the behavior of `and_wrap_original` for `any_instance`.  As that file demonstrates, there are other facets of behavior of `and_wrap_original` that will hopefully "just work" for `any_instance` (since it winds up delegating to the same thing) but we don't know that that's the case until we test it, so I'd like to see those other facets covered by specs.  Ideally, you could use a shared example group for it.  We've done a similar thing in [receive_spec.rb](https://github.com/rspec/rspec-mocks/blob/cfaf4ef18643cbf7fbb6e0bafc93d9f6083c438d/spec/rspec/mocks/matchers/receive_spec.rb#L141).  The key abstractions that make that work for both the `allow` and `allow_any_instance_of` cases are `wrapped` (which returns `allow(object)`, `allow_any_instance_of(klass)`, `expect(object)` or `expect_any_instance_of(object)` as appropriate for the specific example group) and `receiver` (which returns the individual object the message is sent to.  See [here](https://github.com/rspec/rspec-mocks/blob/cfaf4ef18643cbf7fbb6e0bafc93d9f6083c438d/spec/rspec/mocks/matchers/receive_spec.rb#L262-L283).

rf-:
For the record, I don't think `receive_message_chain` will work with `and_wrap_original` regardless of whether `any_instance` is in play. I've never actually used `receive_message_chain`, so I'm not super worried about it personally, but I wanted to flag it as another area where `and_wrap_original` probably won't work correctly yet.

I'll work on improving the specs for this PR. Thanks for the feedback and suggested approach.

rf-:
OK, I pushed an updated version that uses shared examples to apply all of the specs with and without `any_instance`. I'm not sure about the naming though—I never really know what to call shared example groups. Let me know if you want any changes on that front.

spec/rspec/mocks/and_wrap_original_spec.rb
```diff
@@ -10,54 +10,68 @@ def results
...
+      end
     end
 
-    it "ignores previous stubs" do
-      allow(instance).to receive(:results) { "results" }
-      allow(instance).to receive(:results).and_wrap_original { |m| m.call }
-      expect(instance.results).to_not eq "results"
+    describe "allow(...).to receive(...).and_wrap_original" do
+      it_behaves_like "using and_wrap_original" do
+        let(:target) { allow(instance) }
```
myronmarston: What do you think about naming this `allow_it` rather than `target`?  Then at the call site it reads quite nicely:

``` ruby
allow_it.to receive(...)...
# rather than:
target.to receive(...)...
```

myronmarston:
Thanks, @rf-.  This is great.  I left one naming suggestion but otherwise I think we are ready to merge this.

spec/rspec/mocks/and_wrap_original_spec.rb
rf-: Sure, that sounds good.

rf-:
OK, just pushed a version using `allow_it`.

myronmarston:
Merged.  Thanks!

rf-:
No problem, thanks for the quick merge.

PR #870, opened 2015-01-17
Include details on how to run against `master`.

cupakromer:
A very common issue is when a user attempts to run against the `master`
branch. Simply pointing to `master` doesn't work. This adds the steps
necessary to use `master`.

We need all of the repos in order to also satisfy any other gems which
may have a dependency on the `rspec` gem.

[ci skip]


myronmarston:
LGTM

PR #871, opened 2015-01-19
Pr 868 followups

myronmarston:
This has a couple small fixups for #868:

* Fix comment as per @waterlink's suggestion on #868.
* I realized that the implementation in #868 would fail in a situation when it should not, and addressed that here.


JonRowe:
:+1:

PR #873, opened 2015-01-21
824 followups

myronmarston:
Deal with verified doubles for classes that redefined `name`.

    These sorts of classes worked fine before #824 was
    merged, and the switch to using a NamedObjectReference
    based on the name broke things for them. This restores
    support.


JonRowe:
LGTM

Issue #875, opened 2015-01-21
"Diff" is missing a line break


myronmarston:
I just got this error:

```
       <ShardManagement::DatabaseRepository (class)> received :current_db_for with unexpected arguments
         expected: ({:campaign=>#<struct Campaign cmoz_env="prod", user_id=2, legacy_campaign_id=3>, :cycle_type=>"week"})
              got: ({:cycle_type=>CycleType::Week, :campaign=>#<struct Campaign cmoz_env="prod", user_id=2, legacy_campaign_id=3>})Diff:

       @@ -1,4 +1,4 @@
       -[{:campaign=>
       -   #<struct Campaign cmoz_env="prod", user_id=2, legacy_campaign_id=3>,
       -  :cycle_type=>"week"}]
       +[{:cycle_type=>CycleType::Week,
       +  :campaign=>
       +   #<struct Campaign cmoz_env="prod", user_id=2, legacy_campaign_id=3>}]

        Please stub a default value first if message might be received with other args as well.
```

Notice that `Diff` is crammed at the end of line rather than on its own line.  This should be fixed before 3.2 ships.  I might get to it later if no one else does.

myronmarston:
Addressed by #876.

PR #876, opened 2015-01-22
FIX# Displaying 'Diff' in new line

sivagollapalli:
Issue #875

### DIFF
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -65,7 +65,7 @@ module RSpec
         diff = diff_message(expectation.expected_args, args)
 
         message = default_error_message(expectation, expected_args, actual_args)
-        message << "Diff:\n #{diff}" unless diff.empty?
+        message << "\n Diff:\n #{diff}" unless diff.empty?
         message << "\n Please stub a default value first if message might be received with other args as well. \n"
 
         __raise message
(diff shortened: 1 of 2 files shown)


JonRowe:
Thanks for tackling this, could you add a spec for it?

lib/rspec/mocks/error_generator.rb
```diff
@@ -65,7 +65,7 @@ def raise_missing_default_stub_error(expectation, *args)
         diff = diff_message(expectation.expected_args, args)
 
         message = default_error_message(expectation, expected_args, actual_args)
-        message << "Diff:\n #{diff}" unless diff.empty?
+        message << "\n Diff:\n #{diff}" unless diff.empty?
```
myronmarston: In `raise_unexpected_message_args_error`, we use `"\nDiff"`, which left-aligns `Diff` against the margin, whereas this has Diff indented one space.  I don't think we want that space.

imtayadeway:
I began looking at https://github.com/rspec/rspec-mocks/issues/875 before realizing there was already this open PR for it. I decided to complete it and address the test coverage as @JonRowe suggests. Not to wish to interfere here, but I am happy to help try to close this if @sivagollapalli is unable to. Thanks!

sivagollapalli:
@JonRowe Added spec for the issue. @myronmarston Removed extra space as you suggested.

myronmarston:
Thanks, @sivagollapalli!

PR #882, opened 2015-01-31
Removed rubysl from the Gemfile.

YorickPeterse:
This hasn't been required anymore since quite a while. Starting with 2.2.0 (if I
remember correctly) Rubinius takes care of ensuring the rubysl Gems are always
available, with or without Bundler.


myronmarston:
Thanks!

Issue #874, opened 2015-01-21
Negative expectation failures should be surfaced during verification


x-yuri:
Consider the following app:

`Gemfile`:

```ruby
source "https://rubygems.org"
ruby '2.1.5'

gem 'sinatra', git: 'https://github.com/sinatra/sinatra.git', ref: 'b39d72d'
gem 'sentry-raven', '~>0.12.2'
gem 'rspec', '~>3.1.0'
gem 'rack-test', '~>0.6.2'
```

`1.rb`:

```ruby
require 'rubygems'
require 'bundler/setup'

require 'sinatra'
require 'sentry-raven'

Raven.configure(true) do |config|
  config.dsn = '...'
end

use Raven::Rack

get '/' do
  raise 'test exception'
end

ENV['RACK_ENV'] = 'test'

require 'rack/test'

describe 'app' do
  include Rack::Test::Methods

  def app
    @app || Sinatra::Application
  end

  it 'passes errors to Raven' do
    expect(Raven.client).not_to receive(:send)
    # expect(Raven.client).to receive(:send)   # works the same as previous line
    get '/'
  end
end
```

Additionally, I've added some debug statements:

https://github.com/getsentry/raven-ruby/blob/0.12.3/lib/raven/base.rb#L75:

```ruby
    def send(evt)
puts 'inside Raven.send'
p client.method(:send).source_location
      client.send(evt)
```

https://github.com/getsentry/raven-ruby/blob/0.12.3/lib/raven/client.rb#L25:

```ruby
    def initialize(configuration)
puts 'inside Raven::Client.initialize'
      @configuration = configuration
```

https://github.com/getsentry/raven-ruby/blob/0.12.3/lib/raven/client.rb#L19:

```ruby
    def send(event)
puts 'inside Raven::Client#send'
      unless configuration.send_in_current_environment?
```

Then, when I run the spec:

```
$ rspec 1.rb
inside Raven::Client.initialize
D, [2015-01-21T13:15:45.492433 #19390] DEBUG -- : ** [Raven] Collecting RuntimeError: test exception
inside Raven.send
["/home/yuri/.gem/ruby/2.1.5/gems/rspec-mocks-3.1.3/lib/rspec/mocks/method_double.rb", 59]
.

Finished in 0.07989 seconds (files took 0.40847 seconds to load)
1 example, 0 failures
```

But the test should fail. Now if I comment the expectation out:

```
inside Raven::Client.initialize
[...]

PR #884, opened 2015-02-01
Ensure negative expectations raise during RSpec verification time.

samphippen:
During the execution of an example  (as noted in #874) it is possible
for a framework that wraps a mocked object to catch the
RSpec::Mocks::MockExpectationError raised by a negative expectation. This
exception is thrown immediately and so if it is caught the running
example will pass.

This patch causes message expectations to also raise the same error at
verification time.

This commit does not have a passing spec suite, that is intentional. I
want to get feedback on the implementation before I go through all the
broken tests and fix them. The relevant new test can be found on lines
228 to 239 of receive_spec.rb

### DIFF
--- a/spec/rspec/mocks/at_most_spec.rb
+++ b/spec/rspec/mocks/at_most_spec.rb
@@ -76,26 +76,27 @@ module RSpec
         expect(@double).to receive(:do_something).at_most(2).times
         @double.do_something
         @double.do_something
-        expect {
+
+        expect_fast_failure_from(@double, /expected: at most 2 times.*received: 3 times/m) do
           @double.do_something
-        }.to raise_error(/expected: at most 2 times.*received: 3 times/m)
+        end
       end
 
       it "fails fast when at_most(:once) and is called twice" do
         expect(@double).to receive(:do_something).at_most(:once)
         @double.do_something
-        expect {
+        expect_fast_failure_from(@double, /expected: at most 1 time.*received: 2 times/m) do
           @double.do_something
-        }.to raise_error(/expected: at most 1 time.*received: 2 times/m)
+        end
       end
 
       it "fails fast when at_most(:twice) and is called three times" do
         expect(@double).to receive(:do_something).at_most(:twice)
         @double.do_something
         @double.do_something
-        expect {
+        expect_fast_failure_from(@double, /expected: at most 2 times.*received: 3 times/m) do
           @double.do_something
-        }.to raise_error(/expected: at most 2 times.*received: 3 times/m)
+        end
       end
 
       it "fails fast when at_most(:thrice) and is called four times" do
@@ -103,11 +104,10 @@ module RSpec
         @double.do_something
         @double.do_something
         @double.do_something
-        expect {
+        expect_fast_failure_from(@double, /expected: at most 3 times.*received: 4 times/m) do
           @double.do_something
-        }.to raise_error(/expected: at most 3 times.*received: 4 times/m)
+        end
       end
-
     end
   end
 end
--- a/spec/rspec/mocks/matchers/receive_spec.rb
+++ b/spec/rspec/mocks/matchers/receive_spec.rb
@@ -183,17 +183,17 @@ module RSpec
         it 'sets up a negative message expectation that fails if the message is received' do
           wrapped.not_to receive(:foo)
 
-          expect {
+          expect_fast_failure_from(receiver, /expected: 0 times.*received: 1 time/m) do
             receiver.foo
-          }.to raise_error(/expected: 0 times.*received: 1 time/m)
+          end
         end
 
         it 'supports `to_not` as an alias for `not_to`' do
           wrapped.to_not receive(:foo)
 
-          expect {
+          expect_fast_failure_from(receiver, /expected: 0 times.*received: 1 time/m) do
             receiver.foo
-          }.to raise_error(/expected: 0 times.*received: 1 time/m)
+          end
         end
 
         it 'allows the caller to constrain the received arguments' do
@@ -204,9 +204,9 @@ module RSpec
             receiver.foo(:b)
           }.not_to raise_error
 
-          expect {
+          expect_fast_failure_from(receiver, /expected: 0 times.*received: 1 time/m) do
             receiver.foo(:a)
-          }.to raise_error(/expected: 0 times.*received: 1 time/m)
+          end
         end
 
         it 'prevents confusing double-negative expressions involving `never`' do
@@ -453,23 +453,23 @@ module RSpec
         end
 
         it 'supports `expect(...).not_to receive`' do
-          dbl = double
+          expect_fast_failure_from(double) do |dbl|
+            framework.new.instance_exec do
+              expect(dbl).not_to receive(:foo)
+            end
 
-          framework.new.instance_exec do
-            expect(dbl).not_to receive(:foo)
+            dbl.foo
           end
-
-          expect { dbl.foo }.to raise_error(RSpec::Mocks::MockExpectationError)
         end
 
         it 'supports `expect(...).to_not receive`' do
-          dbl = double
+          expect_fast_failure_from(double) do |dbl|
+            framework.new.instance_exec do
+              expect(dbl).to_not receive(:foo)
+            end
 
-          framework.new.instance_exec do
-            expect(dbl).to_not receive(:foo)
+            dbl.foo
           end
-
-          expect { dbl.foo }.to raise_error(RSpec::Mocks::MockExpectationError)
         end
       end
 
--- a/spec/rspec/mocks/reraising_eager_raises_spec.rb
+++ b/spec/rspec/mocks/reraising_eager_raises_spec.rb
@@ -1,0 +1,49 @@
+require "spec_helper"
+
+RSpec.describe "Reraising eager raises during the verify step" do
+  it "does not reraise when a double receives a message that hasn't been allowed/expected" do
+    with_unfulfilled_double do |dbl|
+      expect { dbl.foo }.to fail
+      expect { verify_all }.not_to raise_error
+    end
+  end
+
+  it "reraises when a negative expectation receives a call" do
+    with_unfulfilled_double do |dbl|
+      expect(dbl).not_to receive(:foo)
+      expect { dbl.foo }.to fail
+      expect { verify_all }.to fail_with(/expected: 0 times with any arguments/)
+    end
+  end
+
+  it "reraises when an expectation with a count is exceeded" do
+    with_unfulfilled_double do |dbl|
+      expect(dbl).to receive(:foo).exactly(2).times
+
+      dbl.foo
+      dbl.foo
+
+      expect { dbl.foo }.to fail
+      expect { verify_all }.to fail_with(/expected: 2 times with any arguments/)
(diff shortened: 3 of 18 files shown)


myronmarston:
Thanks for tackling this, @samphippen.  Thinking about this some more, I realized that a negative message expectation is just _one_ case where we eagerly fail rather than failing on `verify`.  There are other cases, such as when the message is received with the wrong args, or received too many times, or received out of order.  It seems to me that _all_ of these cases are prone to the same problem: if the eagerly raised error is rescued by something like Sinatra, then the example won't fail as it should.  It seems to me that we'd be better off fixing this issue for _all_ the eager cases, and not just for the negative expectation case.  For example, for the `verify_messages_received` case, I think it can be simplified too:

``` ruby
        def verify_messages_received
          return if expected_messages_received?
          InsertOntoBacktrace.line(@expected_from) { generate_error }
        end
```

Much simpler, eh?  I tried out that change and noticed that it caused about 30 failures but the handful I looked at looked correct since they were cases that failed eagerly that the spec itself caught, so a simple `reset object` will do the trick.  Before making such a change we should also catalog all the situations where eager exceptions are raised, and be sure to add specs for each for this new "should also raise on verification" behavior.

samphippen:
@myronmarston :+1: I'll continue to work on this over the coming week.

samphippen:
@myronmarston I used your implementation and started fixing the specs up. 
My understanding of the other cases required are:

* argument mismatched
* Invocation count mismatches

I think that's it?

myronmarston:
I'll try to look at this tomorrow...ran out of time today.

myronmarston:
Sorry it took me so long to get back to you; I had to do an audit to answer your question!

Here are the cases I've found that raise errors eagerly (e.g. when the message is received):

* When a non-`as_null_object` test-double receives a message that hasn't been `allowed` or `expected`,
* When a stub is made using `with` and different args are received.
* When a verified double receives a message with arguments that the method signature cannot support.
* When a message expectation configured using `with` receives unexpected args.
* When a message expectation is configured with an expected count and that count is exceeded.
* When an ordered message expectation is received out of order.
* When a stub or message expectation is configured to `and_yield` and no block has been passed.
* When a stub or message expectation is configured to `and_yield` and a block is passed with a different signature (e.g. number of args).

Initially, I was thinking we would want `verify` to re-raise errors for all of these cases but now I'm realizing that's a really bad idea.  I think we only want to re-raise on `verify` for the _expectations_ -- if it's just that wrong args were passed or some other thing raising again later is a terrible idea.  RSpec should not allow an example to pass if an expectation is violated (and thus, raising during `verifying` is a good idea, IMO for expectation cases).  But for cases where no expectation was violated raising would obviously be wrong because the user would have no way to make the example pass even though it's not an expectation failure.

So I think the cases to re-raise for are:

* When a message expectation configured using `with` receives unexpected args.
* When a message expectation is configured with an expected count and that count is exceeded.
* When an ordered message expectation is received out of order.

...which I guess is the same list you had plus the ordering one.

samphippen:
@myronmarston I started writing specs for all the cases. I think having them in their own file makes sense, do you agree? I'm aware the new spec file is a bit duplicated right now, so don't feel the need to leave a review note to that effect (I'll fix it up). I'm thinking I'm going to put the cases that don't reraise into their own context, and the ones that do also into their own context. Do you see the current implementation causing any of the cases that shouldn't raise (that you've outlined above) causing any issues?

samphippen:
@myronmarston can you take a look at this when you get a chance. Thanks in advance :)

spec/rspec/mocks/at_most_spec.rb
```diff
@@ -79,6 +79,7 @@ module Mocks
         expect {
           @double.do_something
         }.to raise_error(/expected: at most 2 times.*received: 3 times/m)
+        reset @double
```
myronmarston: So this `expect { }.to raise_error; reset double` pattern is showing up over, and over and over again through the spec suite.  I'm thinking we should come up with a common abstraction to handle that.

WDYT of this?

``` ruby
expect_fast_failure_from(double, /expected: at most 2 times.*received: 3 times/m) do
  double.do_something
end
```

`expect_fast_failure_from` would take of `expect { }.to raise_error` and the `reset`.

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -438,23 +444,23 @@ def receiver.method_missing(*a); end # a poor man's stub...
...
-            expect(dbl).not_to receive(:foo)
+            expect { dbl.foo }.to raise_error(RSpec::Mocks::MockExpectationError)
           end
-
-          expect { dbl.foo }.to raise_error(RSpec::Mocks::MockExpectationError)
         end
 
         it 'supports `expect(...).to_not receive`' do
-          dbl = double
+          with_unfulfilled_double do |dbl|
```
myronmarston: I think `expect_fast_failure_from` could take the place of `with_unfulfilled_double` in these situations as well -- you could have it yield the double to the block.

spec/rspec/mocks/reraising_eager_raises_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  it "does not reraise when a double receives a message that hasn't been allowed/expected" do
+    expect {
+      with_unfulfilled_double do |dbl|
+        begin
+          dbl.foo
+        rescue RSpec::Mocks::MockExpectationError
+          RSpec::Mocks.verify
+        end
+      end
+    }.not_to raise_error
```
myronmarston: Putting the `verify` in the `rescue` clause is a little weird -- consider that if `RSpec::Mocks::MockExpectationError` is not raised, the verify will never happen, and this spec will pass even though it's not doing what you expect.  It would be a false positive.  Instead, can you use this pattern?

``` ruby
expect { dbl.foo }.to fail
expect { verify_all }.not_to raise_error
```

It's shorter too :).  `fail` is defined in `spec/support/matchers.rb`.  `verify_all` is defined in `spec_helper`.

myronmarston:
I just kicked the build now that rspec/rspec-core#1879 has been merged...hopefully this will go green now.

JonRowe:
Seems not, :/

myronmarston:
> Seems not, :/

It's much better, though -- in the last build (https://travis-ci.org/rspec/rspec-mocks/builds/49933519) it was failing when running rspec-core's specs.  Now it's not.  It's just 1.8.7 that's failing on some rspec-mocks specs now.

myronmarston:
@samphippen, are you planning to get back to this?

samphippen:
yup. I'm gonna spend some time this weekend on this.

samphippen:
https://travis-ci.org/rspec/rspec-mocks/jobs/54355732#L2334 :skull: I will see if I can debug this locally.

samphippen:
I'm having trouble getting REE set up on my mac. @myronmarston can you look at the failure and reckon if it's legit. I'm going to see if I can get a linux box set up with REE inside it.

spec/spec_helper.rb
```diff
@@ -43,8 +43,14 @@ def reset_all
   def with_unfulfilled_double
     d = double("double")
     yield d
+  ensure
     reset d
   end
+
+  def expect_fast_failure_from(double, class_or_message=Exception, message=/.*/, &blk)
```
myronmarston: `Exception` seems like a poor choice for the default here, given that at virtually all the call sites above you pass `RSpec::Mocks::MockExpectationError` (and in the places you don't, I'm wondering if you should!).  It seems like the norm is `RSpec::Mocks::MockExpectationError` -- after all, in rspec-mocks, that's who we signal failures -- so we should make that the default and only make the caller pass a different class if it really needs something different.

spec/rspec/mocks/reraising_eager_raises_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      expect { dbl.foo }.to fail
+      expect { verify_all }.not_to raise_error
+    end
+  end
+
+  it "reraises when a negative expectation receives a call" do
+    with_unfulfilled_double do |dbl|
+      expect(dbl).not_to receive(:foo)
+      expect { dbl.foo }.to fail
+      expect { verify_all }.to raise_error(/expected: 0 times with any arguments/)
```
myronmarston: `raise_error(regex)` here and below can instead be `fail_with(regex)` (see `spec/support/matchers.rb`).  That's a bit more descriptive and is a stronger expectation since it specifies the expectation class for you.

spec/rspec/mocks/reraising_eager_raises_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  it "reraises when an expectation with a count is violated" do
```
myronmarston: I think "exceeded" would be better than "violated".  "violated" could refer to either too few invocations or too many but this really only applies to when there are too many, and "exceeded" is a better word for that.

spec/spec_helper.rb
myronmarston: Actually, I suspect that there is no need to specify a class other than `RSpec::Mocks::MockExpectationError`.  If that's the case, this can be simplified to:

``` ruby
def expect_fast_failure_from(double, *fail_with_args)
  expect { yield double }.to fail_with(*fail_with_args)
  reset double
end
```

myronmarston:
@samphippen -- The problem isn't ree-specific.  It's failing on MRI 1.8.7 for me.  I'm realizing that the failure has to do with hash ordering -- on 1.9+ it's deterministic and this spec consistently passes, but on 1.8.7 the hash ordering is non-deterministic, and as a result, in `RSpec::Mocks::Proxy#verify`, when it does `@method_doubles.each_value`, it's not deterministic whether the method double for `:foo` or the method double for `:bar` is verified first.  Both method doubles trigger failures:

* `foo` was never called but was expected, so it has a verify error.
* `bar` was called out of order (e.g. before `foo` was called) so it has a verify error.

Note, however, that the message you are expecting (`/\.foo/`) has nothing to do with the ordering; it's simply that the message was never received, which isn't an eager raise situation anyway.  It's the `bar` one that raised eagerly, and your expectation doesn't touch on that at all!

I think you want something like this instead:

``` ruby
  it "reraises when an expectation is called out of order" do
    with_unfulfilled_double do |dbl|
      expect(dbl).to receive(:foo).ordered
      expect(dbl).to receive(:bar).ordered
      expect { dbl.bar }.to fail
      dbl.foo # satisfy the `foo` expectation so that only the bar one fails below
      expect { RSpec::Mocks.verify }.to fail_with(/received :bar out of order/)
    end
  end
```

However, that doesn't pass on any ruby; instead we get:

```
#<RSpec::Mocks::MockExpectationError: (Double "double").bar(*(any args))
           expected: 1 time with any arguments
           received: 0 times with any arguments>
```

...which is simply wrong.  `bar` _was_ received 1 time, so the error message is false.  Given that this is a bug that's existing and not really part of the change you are working, it's probably OK to make the spec pending for now, although it would be great if you would come back to it once this is merged (if you have the time, of course).

samphippen:
@myronmarston how's this looking? I went through the feedback and made edits.

myronmarston:
LGTM. Can you squash it into a smaller number of commits before merging?

PR #886, opened 2015-02-05
When stubbing `MyClass.new` verify against `MyClass#initialize`.

### LINKED ISSUE #883: Idea: use the method signature of `MyClass#initialize` for verified double argument verification of `MyClass.new`
When a user mocks or stubs `.new` on a verified partial class double or on a `class_double`, we are not currently able to do any argument verification because the signature of `Class#new` is so generic.  Compare the two parameter lists:

``` irb
irb(main):001:0> class Foo; def initialize(foo:, bar:); end; end
=> :initialize
irb(main):002:0> Foo.method(:new).parameters
=> [[:rest]]
irb(main):003:0> Foo.instance_method(:initialize).parameters
=> [[:keyreq, :foo], [:keyreq, :bar]]
```

I think that for the special case of mocking or stubbing `.new` on a class, we should look at the method signature of `#initialize` instead since we know that `Class#new` delegates to that.  It is possible for users to redefine `new` so we'll probably want to only do this if `MyClass.method(:new).owner == Class`.

There may be some negative repercussions to this I'm not thinking of (can anyone think of any?).  Also, are there any other "hook" methods (`initialize`, in this case) that define the method signature for another common method (`new`, in this case) that should receive similar treatment?  I can't think of any.

/cc @xaviershay @rspec/rspec

myronmarston:
Closes #883.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,12 @@
+### 3.3.0 Development
+[Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.2.0...master)
+
+Enhancements:
+
+* When stubbing `new` on `MyClass` or `class_double(MyClass)`, use the
+  method signature from `MyClass#initialize` to verify arguments.
+  (Myron Marston, #886)
+
 ### 3.2.0 / 2015-02-03
 [Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.1.3...v3.2.0)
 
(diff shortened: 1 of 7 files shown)


JonRowe:
JRuby doesn't like this :/

myronmarston:
> JRuby doesn't like this :/

I know.  I found a bug in JRuby I'll be opening with the JRuby team later.

myronmarston:
OK, @JonRowe, the JRuby builds are passing now.

JonRowe:
Sweet :)

mikegee:
We ran into this at work last month and it was pretty annoying. We made a class-level `build` method for each class to work around it. 

I was planning to attempt a PR, I'm glad I won't need to. Thanks guys.

Issue #890, opened 2015-02-10
Matcher descriptions not used for `received` in too-many-calls failure output


myronmarston:
``` ruby
it 'does not use matcher descriptions in the `received:` description' do
  dbl = double
  expect(dbl).to receive(:foo).with(anything, hash_including(bar: anything))

  dbl.foo(1, bar: 2)
  dbl.foo(1, bar: 3)
end
```

Fails with:

```
     Failure/Error: expect(dbl).to receive(:foo).with(anything, hash_including(bar: anything))
       (Double).foo(anything, hash_including(:bar=>"anything"))
           expected: 1 time with arguments: (anything, hash_including(:bar=>"anything"))
           received: 2 times with arguments: (#<RSpec::Mocks::ArgumentMatchers::AnyArgMatcher:0x007faefc2fb7c0>, #<RSpec::Mocks::ArgumentMatchers::HashIncludingMatcher:0x007faefebf62d8 @expected={:bar=>#<RSpec::Mocks::ArgumentMatchers::AnyArgMatcher:0x007faefc2fb7c0>}>)
```

Notice the inspect output of the matchers is used for `received` but the description is properly used for `expected`.  It should be used for both.

PR #891, opened 2015-02-10
Fix argument formatting for received.

imtayadeway:
https://github.com/rspec/rspec-mocks/issues/890

### DIFF
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -129,40 +129,30 @@ module RSpec
       # @private
       def received_part_of_expectation_error(actual_received_count, *args)
         "received: #{count_message(actual_received_count)}" +
-          actual_method_call_args_description(actual_received_count, args)
+          method_call_args_description(args) do
+            actual_received_count > 0 && args.length > 0
+          end
       end
 
       # @private
       def expected_part_of_expectation_error(expected_received_count, expectation_count_type, argument_list_matcher)
         "expected: #{count_message(expected_received_count, expectation_count_type)}" +
-          expected_method_call_args_description(argument_list_matcher.expected_args)
-      end
-
-      # @private
-      def actual_method_call_args_description(count, args)
-        method_call_args_description(args) ||
-          if count > 0 && args.length > 0
-            " with arguments: #{format_args(*args)}"
-          else
-            ""
+          method_call_args_description(argument_list_matcher.expected_args) do
+            argument_list_matcher.expected_args.length > 0
           end
       end
 
       # @private
-      def expected_method_call_args_description(args)
-        method_call_args_description(args) ||
-          if args.length > 0
-            " with arguments: #{format_args(*args)}"
-          else
-            ""
-          end
-      end
-
-      # @private
-      def method_call_args_description(args)
+      def method_call_args_description(args, &block)
         case args.first
         when ArgumentMatchers::AnyArgsMatcher then " with any arguments"
         when ArgumentMatchers::NoArgsMatcher  then " with no arguments"
+        else
+          if block.call
+            " with arguments: #{format_args(*args)}"
+          else
+            ""
+          end
         end
       end
 
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -142,7 +142,7 @@ module RSpec
       def actual_method_call_args_description(count, args)
         method_call_args_description(args) ||
           if count > 0 && args.length > 0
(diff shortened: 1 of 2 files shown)


spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -298,11 +298,26 @@ def receiver.method_missing(*a); end # a poor man's stub...
...
+              )
+            end
+          end
+
+          context "when a message is received the wrong number of times" do
+            it "sets up a message expectation that formats argument matchers correctly" do
+              wrapped.to receive(:foo).with(anything, hash_including(bar: anything))
+
+              receiver.foo(1, bar: 2)
+              receiver.foo(1, bar: 3)
```
JonRowe: Our specs run against 1.8.7 so this needs to be `:bar => 3`

imtayadeway: Thanks Jon.  I'll address this and also the failing cukes which I forgot to check (!!). I'll try to get this wrapped up tonight.

imtayadeway: @JonRowe just updated the hash syntax. after fixing I was unable to reproduce the failures I was getting before in the cucumber suite (i reverted one of the lines but was still green.) maybe just a quirk on my machine. let me know if there's anything that remains to be done. thanks!

lib/rspec/mocks/error_generator.rb
```diff
@@ -142,7 +142,7 @@ def expected_part_of_expectation_error(expected_received_count, expectation_coun
       def actual_method_call_args_description(count, args)
         method_call_args_description(args) ||
           if count > 0 && args.length > 0
-            " with arguments: #{args.inspect.gsub(/\A\[(.+)\]\z/, '(\1)')}"
+            " with arguments: #{format_args(*args)}"
           else
             ""
           end
```
myronmarston: This method now looks almost identical to `expected_method_call_args_description`.  Can they be unified some how?

imtayadeway: Playing around with this now. Does it make sense for one to depend on the received count but not the other? Introducing `count` to both will squash messages like `expected: 0 times with arguments: (\"unexpected text\")\n`. I'm assuming that would not be desirable?

imtayadeway: I'm a little hesitant about the last commit - but was the best way I could find of unifying these two methods while preserving the same behavior. Let me know what you think.  Thanks!

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -298,11 +298,26 @@ def receiver.method_missing(*a); end # a poor man's stub...
...
+              wrapped.to receive(:foo).with an_instance_of Fixnum
+              expect { verify_all }.to(
+                raise_error(/expected: 1 time with arguments: \(an instance of Fixnum\)\n\s+received: 0 times$/)
+              )
+            end
+          end
+
+          context "when a message is received the wrong number of times" do
+            it "sets up a message expectation that formats argument matchers correctly" do
+              wrapped.to receive(:foo).with(anything, hash_including(bar: anything))
```
JonRowe: Introduced the new style hash syntax here which is breaking the build again :)

lib/rspec/mocks/error_generator.rb
```diff
@@ -129,40 +129,35 @@ def raise_expired_test_double_error
...
-          else
-            ""
-          end
+      def show_method_call_args_description?(type, args, count)
+        if type == :actual
+          count > 0 && args.length > 0
+        elsif type == :expected
+          args.length > 0
+        end
       end
```
myronmarston: Instead of defining this helper method and making callers of `method_call_args_description` pass the `count` and `type` args (neither of which are used in the description itself!), WDYT about making `method_call_args_description` yield to a block that would return a boolean value?  Then this logic could be moved into the appropriate caller.  I think that would be a bit cleaner.

JonRowe: :+1: on this idea, I'd like to remove the actual / expected attribute :)

spec/rspec/mocks/matchers/receive_spec.rb
imtayadeway: oops! must have missed that the first time. fixing now...

lib/rspec/mocks/error_generator.rb
imtayadeway: I like this a lot better =) let me know if there's anything else

JonRowe:
Thanks!

Issue #838, opened 2014-12-11
The "does not implement" error message from verifying doubles doesn't differentiate between instance methods and class methods


myronmarston:
This is something I realized when responding to https://groups.google.com/forum/#!topic/rspec/EQQ5nZAvKxE .  The user had used `instance_double("TestModule")` where they should have used `class_double("TestModule")`.  The failure message they got didn't help them much:

```
TestModule does not implement: config
```

...but clearly, `TestModule` does implement `config` (it just does so as a class method, not an instance method like the `instance_double` expects).  I don't yet have a suggestion for a better wording but it would be nice if the wording would indicate that `TestModule` doesn't implement `config` as an instance method as `instance_double` requires.  For bonus points, it'd be even cooler if the error message indicated that it is implemented as a class method and suggested that the user use `class_double` instead.  (That might be take way more effort than it's worth, though).

@xaviershay, do you have any ideas on wording here that can help clarify this?

imtayadeway:
> For bonus points, it'd be even cooler if the error message indicated that it is implemented as a class method and suggested that the user use class_double instead. (That might be take way more effort than it's worth, though).

This was not attempted in https://github.com/rspec/rspec-mocks/pull/896. Do you want to leave this issue open while I can investigate this?

myronmarston:
> This was not attempted in #896. Do you want to leave this issue open while I can investigate this?

Sure!

samphippen:
This can be closed because of #908

PR #896, opened 2015-02-17
Add more specific "does not implement" messages.

imtayadeway:
The unimplemented error message gives no indication of whether the
object is a class or instance double. This change passes that object to
the error generator so that it can create a more specific message
depending on whether it is an instance or class double. It defaults to
the original "<object> does not implement <method>" for anything else.

https://github.com/rspec/rspec-mocks/issues/838


imtayadeway:
Note: I deviated a little here from @samphippen's suggested wording, only to reduce my footprint on other tests that depend on the original output (they will all match `/does not implement/`).

JonRowe:
LGTM

PR #900, opened 2015-02-23
Revert arg mutation error

### LINKED ISSUE #892: Builder <hash/> with 3.2
Using rspec-mocks 3.1, I had the following double:
```ruby
double("request detail").tap do |request_detail|
  allow(request_detail).to receive(:to_xml) { |options| options[:builder].tag!(:request) }
end
```

The `options[:builder]` was an instance of Builder::XmlMarkup and it produced `<request/>`. But with rspec-mocks 3.2 it generates `"<hash/>\n<request/>"`.

myronmarston:
Stop calling methods on args passed to a stubbed method.

Any method call may mutate an object, so we can't safely call anything on a passed argument.

This reverts the following PRs and commits:

* #871:
  * 8a0962a6b8a896b1f1f2806a1282f29cde7d2baa ("Reword comment to not use “external” twice.")
  * 52c69f1ef8ac2c0de5ec08c72df709593a6a8ba5 ("Just use `Array#hash`.")
* #868:
  * ed3fb3b1ac9e1187eeb44f4a68dd9878f9ce3ec6 ("Provide a clear error when received message args are mutated.")
  * fc9e9a59904ec52fa8e26e07ee65c41029ef18d0 ("Ignore arg mutations for method calls we are not concerned with.")

This adds notes to the docs about using mutated args with `have_received(…).with(…)`.

Fixes #892.

(Also, I noticed that our `@example` formatting could use some improvements so I did that real quick.  Later realized that I should have made that a separate PR but don't feel like putting the effort into splitting it...sorry).

/cc @xaviershay

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,8 +136,7 @@ class_spy("Invitation") # => same as `class_double("Invitation").as_null_object`
 object_spy("Invitation") # => same as `object_double("Invitation").as_null_object`
 ```
 
-Stubbing and verifying messages received in this way implements the Test Spy
-pattern.
+Verifying messages received in this way implements the Test Spy pattern.
 
 ```ruby
 invitation = spy('invitation')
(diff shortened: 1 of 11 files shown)


JonRowe:
Yep, LGTM

myronmarston:
Thanks, @JonRowe.  I backported this to 3-2-maintenance and add a changlog entry there.  I'll be release 3.2.1 with this in it soon.

JonRowe:
:+1: :)

PR #906, opened 2015-03-06
Simplify `and_raise`.

myronmarston:
Passing args directly to `raise` is much simpler :).


soulcutter:
I think this is fine - although actually now that I merged I noticed the documentation is slightly outdated in that you may now pass multiple args with the class (whereas before that would have caused an `ArgumentError`

soulcutter:
Nevermind, documentation is still accurate. `raise` ignores extra arguments (I had assumed it would pass them along)

PR #908, opened 2015-03-09
Provide a message if the wrong double is used.

imtayadeway:
If an undefined method is stubbed on an instance double, the error generator
will check to see if it is defined on the class level and provide a
message to indicate that this may be what was meant. It also does the
converse for class doubles and instance methods.

https://github.com/rspec/rspec-mocks/issues/838

### DIFF
--- a/spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
+++ b/spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
@@ -28,6 +28,18 @@ module RSpec
         prevents { expect(o).to receive(:defined_instance_method) }
       end
 
+      USE_INSTANCE_DOUBLE_MSG = "Perhaps you meant to use `instance_double`"
+
+      it "suggests using `instance_double` when an instance method is stubbed" do
+        o = class_double("LoadedClass")
+        prevents(a_string_including(USE_INSTANCE_DOUBLE_MSG)) { allow(o).to receive(:defined_instance_method) }
+      end
+
+      it "doesn't suggest `instance_double` when a non-instance method is stubbed'" do
+        o = class_double("LoadedClass")
+        prevents(a_string_excluding(USE_INSTANCE_DOUBLE_MSG)) { allow(o).to receive(:undefined_instance_method) }
+      end
+
       it 'gives a descriptive error message for NoMethodError' do
         o = class_double("LoadedClass")
         expect {
--- a/spec/rspec/mocks/verifying_doubles/instance_double_with_class_loaded_spec.rb
+++ b/spec/rspec/mocks/verifying_doubles/instance_double_with_class_loaded_spec.rb
@@ -28,6 +28,18 @@ module RSpec
         prevents { expect(o).to receive(:defined_class_method) }
       end
 
+      USE_CLASS_DOUBLE_MSG = "Perhaps you meant to use `class_double`"
+
+      it "suggests using `class_double` when a class method is stubbed" do
+        o = instance_double("LoadedClass")
+        prevents(a_string_including(USE_CLASS_DOUBLE_MSG)) { allow(o).to receive(:defined_class_method) }
+      end
+
+      it "doesn't suggest `class_double` when a non-class method is stubbed" do
+        o = instance_double("LoadedClass")
+        prevents(a_string_excluding(USE_CLASS_DOUBLE_MSG)) { allow(o).to receive(:undefined_class_method) }
+      end
+
       it 'allows `send` to be stubbed if it is defined on the class' do
         o = instance_double('LoadedClass')
         allow(o).to receive(:send).and_return("received")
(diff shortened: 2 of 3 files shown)


spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
```diff
@@ -28,6 +28,12 @@ module Mocks
         prevents { expect(o).to receive(:defined_instance_method) }
       end
 
+      it "suggests using `instance_double` when an instance method is stubbed" do
+        o = class_double("LoadedClass")
+        prevents(/Perhaps you meant to use `instance_double`/) { allow(o).to receive(:defined_instance_method) }
+      end
+
+
```
myronmarston: I'd prefer we only had one blank line here.

spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
```diff
@@ -28,6 +28,12 @@ module Mocks
         prevents { expect(o).to receive(:defined_instance_method) }
       end
 
+      it "suggests using `instance_double` when an instance method is stubbed" do
+        o = class_double("LoadedClass")
+        prevents(/Perhaps you meant to use `instance_double`/) { allow(o).to receive(:defined_instance_method) }
+      end
```
myronmarston: I think we also need a negative test -- that is, a test of a method that is not an instance method, to assert that it doesn't recommend using `instance_double`.  As the specs currently exist, we could remove the `if InstanceMethodReference.for(doubled_module, method_name).implemented?` conditional and simply _always_ append the "Perhaps..." bit, and all the specs would pass.

spec/rspec/mocks/verifying_doubles/instance_double_with_class_loaded_spec.rb
```diff
@@ -28,6 +28,11 @@ module Mocks
         prevents { expect(o).to receive(:defined_class_method) }
       end
 
+      it "suggests using `class_double` when a class method is stubbed" do
+        o = instance_double("LoadedClass")
+        prevents(/Perhaps you meant to use `class_double`/) { allow(o).to receive(:defined_class_method) }
+      end
```
myronmarston: Same here -- we need a spec for the negative case.

myronmarston:
Thanks, @imtayadeway -- the implementation looks nice and simple!

spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
JonRowe: :+1:

imtayadeway: Oops! Hehe. Not sure how that crept in....

imtayadeway:
Just updated my commit. I've never used an anchor like that before - do you think that's the best way to test that an error message doesn't match a pattern?

spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
```diff
@@ -28,6 +28,16 @@ module Mocks
...
+
+      it "doesn't suggest `instance_double` when a non-instance method is stubbed'" do
+        o = class_double("LoadedClass")
+        prevents(/(?!Perhaps you meant to use `instance_double`)/) { allow(o).to receive(:undefined_instance_method) }
```
myronmarston: This doesn't actually work properly...if you break the implementation (e.g. by changing the conditional to `if true`) this still passes.  I haven't used regex look aheads much so I'm not sure why it's failing, but I think there's a better way to do this:

``` ruby
WRONG_DOUBLE_TYPE_MSG = "Perhaps you meant to use `class_double`"

it "suggests using `class_double` when a class method is stubbed" do
  o = instance_double("LoadedClass")
  prevents(a_string_including(WRONG_DOUBLE_TYPE_MSG)) { allow(o).to receive(:defined_class_method) }
end

it "doesn't suggest `class_double` when a non-class method is stubbed" do
  o = instance_double("LoadedClass")
  prevents(a_string_excluding(WRONG_DOUBLE_TYPE_MSG)) { allow(o).to receive(:undefined_class_method) }
end
```

* This uses `a_string_excluding`, which is defined as the negation of `include` [here](https://github.com/rspec/rspec-mocks/blob/f190e66066108547bff5b833e936ab838d53ffc4/spec/spec_helper.rb#L116)
* `prevents` passes the arg to `raise_error`, which like all RSpec matchers where it makes sense to do so, supports receiving a matcher as an arg.
* I also recommend extracting the message snippet into a constant; that way, the positive/negative cases are kept in sync if the wording ever changes.  In contrast, if the wording was to change, the negative case could just pass because it doesn't match the exact wording even if that part of the message was wrongly included.

imtayadeway: Wow, thanks for the feedback! This works now - the only way I departed was in extracting the constant. I used a slightly different name in each place so as to avoid generating a warning. Let me know if there's anything else I can do. Thanks!

myronmarston:
Thanks!

xaviershay:
neat

PR #911, opened 2015-03-15
Correct any_instance documentation.

### LINKED ISSUE #910: expect_any_instance_of behavior does not match Relish documentation
**summary**

If an expectation is set on a method on `any instance` of a class, rspec-mocks raises an `RSpec::Mocks::MockExpectationError` reading `The message 'bar' was received by #<Foo:1> but has already been received by #<Foo:2>` when the message is received by more than one instance of the class.  This behavior is incongruent with both the name of the method as well as the Relish documentation for `expect_any_instance_of`.

**description**

I'm in the middle of completing a conversion from mocha to rspec-mocks, and an issue I ran into is a difference between the implementations of expecting any_instance of a class to receive a message.  While I expect differences between the two mock frameworks, I found that the actual behavior of RSpec's `expect_any_instance_of` does not match the stated behavior in the Relish documentation:

https://relishapp.com/rspec/rspec-mocks/v/3-2/docs/working-with-legacy-code/any-instance
> For example, in expect_any_instance_of(Widget).to receive(:name).twice it isn't clear whether each specific instance is expected to receive name twice, or if two receives total are expected. (It's the former.)

Note especially that the 'former' case states "**each specific instance** is expected to receive".

The documented behavior is further supported by the method name `expect_any_instance_of`, which reads like `expect_one_or_more_instances_of`.

Because `any_instance` does not permit the use of spies, we cannot verify after the fact that each instance received an allowed message, nor can we set an expectation on multiple instances of a class to receive a message due to the above error.  This makes it impossible, as far as I know, using `expect_any_instance_of` to expect a specific message to be called on more than one instance of a class.

**proposed resolution**

Either the documentation should be clarified, or the behavior updated to reflect the correct behavior.  IMO the expectation error currently being thrown here would be more appropriate to apply as a conditional expectation, or using different syntax.  e.g:

```ruby
expect_an_instance_of(Foo).to receive(:bar) # 'an' instead of 'any'
expect_any_instance_of(Foo).to receive(:bar).on(1).instance # Using some kind of expectation constraint, but this is very ugly.
```

**workaround**

For others, I am using the following workaround to circumvent this issue:

```ruby
module RSpec
  module Mocks
    module ExampleMethods
      # @param [Class] The class whose instances to stub or modify in the yielded block.
      # @param [Block] A block to yield immediately after the constructor is called.
      def with_all_instances_of(klass)
        raise 'Expected a block to be supplied!' unless block_given?
        allow(klass).to(
            receive(:new).and_wrap_original do |m, *args|
              instance = m.call(*args)
              yield instance
              instance
            end)
      end
    end
  end
end
```

Usage (including a failing example demonstrating the above implementation works):

```ruby
module RSpec
  module Mocks
    module ExampleMethods
      # @param [Class] The class whose instances to stub or modify in the yielded block.
      # @param [Block] A block to yield immediately after the constructor is called.
      def with_all_instances_of(klass)
        raise 'Expected a block to be supplied!' unless block_given?
        allow(klass).to(
            receive(:new).and_wrap_original do |m, *args|
              instance = m.call(*args)
              yield instance
              instance
            end)
      end
    end
  end
end

class MyTestClass
  def name
    "FooBar"
  end
end

RSpec.describe MyTestClass do
  describe 'passing examples' do
    it 'stubs every instance' do
      allow_any_instance_of(described_class).to receive(:name).and_return("BarFood")
      expect(described_class.new.name).to eql "BarFood"
      expect(described_class.new.name).to eql "BarFood"
    end

    it 'expects on every instance' do
      with_all_instances_of(described_class) do |instance|
        expect(instance).to receive(:name).and_return("BarFood")
      end
      expect(described_class.new.name).to eql "BarFood"
      expect(described_class.new.name).to eql "BarFood"
    end
  end

  describe 'failing examples' do
    it 'expect_any_instance_of fails if more than one instance fulfill the expectation' do
      expect_any_instance_of(described_class).to receive(:name).and_return("BarFood")

      expect(described_class.new.name).to eql "BarFood"
      expect(described_class.new.name).to eql "BarFood" #This will raise
    end

    it 'expectations within with_all_instances_of block fail if not met on one or more instances' do
      with_all_instances_of(described_class) do |instance|
        expect(instance).to receive(:name).and_return("BarFood")
      end
      expect(described_class.new.name).to eql "BarFood"
      described_class.new
    end
  end
end
```

JonRowe:
Fixes #910

### DIFF
--- a/features/working_with_legacy_code/any_instance.feature
+++ b/features/working_with_legacy_code/any_instance.feature
@@ -17,7 +17,7 @@ Feature: Any Instance
   * The `rspec-mocks` API is designed for individual object instances, but this feature
     operates on entire classes of objects. As a result there are some semantically confusing
     edge cases. For example, in `expect_any_instance_of(Widget).to
-    receive(:name).twice` it isn't clear whether each specific instance is expected to
+    receive(:name).twice` it isn't clear whether a specific instance is expected to
     receive `name` twice, or if two receives total are expected. (It's the former.)
   * Using this feature is often a design smell. It may be that your test is trying to do too
     much or that the object under test is too complex.


myronmarston:
LGTM.

Issue #915, opened 2015-03-26
Confusing failure message for `have_received(...).with(...).once`


myronmarston:
Consider this spec:

``` ruby
RSpec.describe "Using `have_rececive(:message).with(:args).once`" do
  it "mentions the count discrepancy in the failure message" do
    dbl = double.as_null_object
    dbl.foo(:one)
    dbl.foo(:two)
    dbl.foo(:one)

    expect(dbl).to have_received(:foo).with(:one).once
  end
end
```

It fails, which is correct, but the failure message is really confusing:

```
  1) Using `have_rececive(:message).with(:args).once` mentions the count discrepancy in the failure message
     Failure/Error: expect(dbl).to have_received(:foo).with(:one).once
       Double received :foo with unexpected arguments
         expected: (:one)
              got: (:two)
       Diff:
       @@ -1,2 +1,2 @@
       -[:one]
       +[:two]
```

The fact that it received `foo` with `:two` is immaterial.  It's that it received the message with `:one` twice, not once, that causes the failure -- but the failure message doesn't say that.

Note: this started as #841 (which has some failing specs I sketched out) but I've not had a chance to get back to that and probably won't any time soon.  I wanted to convert it to an issue to make it clear it's a free issue for anyone to work on :).  Feel free to borrow as much of the spec changes in #841 as you like, though.

myronmarston:
Fixed by #918.

PR #918, opened 2015-03-29
Fixes issue 915 - displaying invalid counts instead of irrelevant arguments

johnceh:
# Why?
Fix for issue #915 - Confusing failure message when other arguments involved with received messages. Invalid counts more relevant in situations where multiple arguments in play. 

# What Changed?
Previously, an exception was raised immediately after a received message did not match one of items in the argument list. Now the standard invalid count message is returned unless there is no match in the entire list. Tests by @myronmarston copied from #841 added in.

# What Else Changed?
(updated 04/03/2015 with code refactor, adding in suggestions from @myronmarston and @JonRowe)
- Took out new method and rolled it into existing `raise_unexpected_message_args_error` method
- DRY'd up some error methods by adding common code to `message_error` (name to generic?)
- Rolled up `received_arg_list` into `format_received_args` and setup to handle single/multiple messages with single/multiple arguments.
- Removed a few splat operators so a more standardized messages list can be sent for single and multiple messages.

(updated 04/05/2015)
- Added fix for 2nd part of ticket. Display errors fixed for expect to receive messages.
- A couple of different options on this one, so took the path of least resistance.  Added similar_messages to SimpleMessageExpectation should allow the removal of (defined?) check. Adding check for similar_messages.blank? is not necessary, but would reduce redundant calls to unadvise method. Open to suggestions if not matching existing patterns.

(updated 04/08/2015)
- Added no-op to SimpleMessageExpectation (This removed the previous complexity warning I was getting earlier, so I removed the rubocop disable I had to add in earlier)
- Using existing `format_args` instead of using new method
- Making distinctions between args and args_with_mulitple_calls for clarity.

### DIFF
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -50,36 +50,35 @@ module RSpec
       end
 
       # @private
-      def raise_unexpected_message_args_error(expectation, *args)
+      def message_error(expectation, *args)
         expected_args = format_args(*expectation.expected_args)
         actual_args = format_received_args(*args)
-        diff = diff_message(expectation.expected_args, args)
-
         message = default_error_message(expectation, expected_args, actual_args)
-        message << "\nDiff:#{diff}" unless diff.empty?
 
-        __raise message
+        if single_message?(args)
+          diff = diff_message(expectation.expected_args, *args)
+          message << "\nDiff:#{diff}" unless diff.empty?
+        end
+
+        message
+      end
+
+      # @private
+      def raise_unexpected_message_args_error(expectation, *args)
+        __raise message_error(expectation, *args)
       end
 
       # @private
       def raise_missing_default_stub_error(expectation, *args)
-        expected_args = format_args(*expectation.expected_args)
-        actual_args = format_received_args(*args)
-        diff = diff_message(expectation.expected_args, args)
-
-        message = default_error_message(expectation, expected_args, actual_args)
-        message << "\nDiff:\n #{diff}" unless diff.empty?
+        message = message_error(expectation, *args)
         message << "\n Please stub a default value first if message might be received with other args as well. \n"
 
         __raise message
       end
 
       # @private
-      def raise_similar_message_args_error(expectation, *args_for_multiple_calls)
-        expected_args = format_args(*expectation.expected_args)
-        actual_args = args_for_multiple_calls.map { |a| format_received_args(*a) }.join(", ")
-
-        __raise(default_error_message(expectation, expected_args, actual_args))
+      def raise_similar_message_args_error(expectation, *args)
+        __raise message_error(expectation, *args)
       end
 
       def default_error_message(expectation, expected_args, actual_args)
@@ -288,11 +287,12 @@ module RSpec
       end
 
       def format_received_args(*args)
-        args.empty? ? "(no args)" : "(" + received_arg_list(*args) + ")"
-      end
-
-      def received_arg_list(*args)
-        args.map(&:inspect).join(", ")
+        return "(no args)" if args.empty? || args.first.empty?
+        return formatted_single_message(args) if single_message?(args)
+        grouped_list = grouped_args(args).map do |arg, index|
+          [formatted_single_message([arg]), group_count(index, args)].join
+        end
+        grouped_list.join("\n            ")
       end
 
       def count_message(count, expectation_count_type=nil)
@@ -303,6 +303,22 @@ module RSpec
 
       def times(count)
         "#{count} time#{count == 1 ? '' : 's'}"
+      end
+
+      def formatted_single_message(args)
+        "(" + args.first.map(&:inspect).join(", ") + ")"
+      end
+
+      def grouped_args(args)
+        Hash[args.group_by { |x| x }.map { |k, v| [k, v.count] }]
+      end
+
+      def single_message?(args)
+        args.size == 1
+      end
+
+      def group_count(index, args)
+        args.size > 1 || index > 1 ? ' (' + times(index) + ')' : ''
       end
     end
   end
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -50,35 +50,21 @@ module RSpec
       end
 
       # @private
-      def message_error(expectation, *args)
-        expected_args = format_args(*expectation.expected_args)
-        actual_args = format_received_args(*args)
-        message = default_error_message(expectation, expected_args, actual_args)
-
-        if single_message?(args)
-          diff = diff_message(expectation.expected_args, *args)
-          message << "\nDiff:#{diff}" unless diff.empty?
-        end
-
-        message
-      end
-
-      # @private
-      def raise_unexpected_message_args_error(expectation, *args)
-        __raise message_error(expectation, *args)
+      def raise_unexpected_message_args_error(expectation, *args_for_multiple_calls)
+        __raise error_message(expectation, *args_for_multiple_calls)
       end
 
       # @private
       def raise_missing_default_stub_error(expectation, *args)
-        message = message_error(expectation, *args)
+        message = error_message(expectation, *args)
         message << "\n Please stub a default value first if message might be received with other args as well. \n"
 
         __raise message
       end
 
       # @private
-      def raise_similar_message_args_error(expectation, *args)
-        __raise message_error(expectation, *args)
+      def raise_similar_message_args_error(expectation, *args_for_multiple_calls)
+        __raise error_message(expectation, *args_for_multiple_calls)
       end
 
       def default_error_message(expectation, expected_args, actual_args)
@@ -225,6 +211,19 @@ module RSpec
         "with unexpected arguments\n  expected: #{expected_args_string}\n       got: #{actual_args_string}"
       end
 
+      def error_message(expectation, *args_for_multiple_calls)
+        expected_args = format_args(*expectation.expected_args)
+        actual_args = format_received_args(*args_for_multiple_calls)
+        message = default_error_message(expectation, expected_args, actual_args)
+
+        if args_for_multiple_calls.one?
+          diff = diff_message(expectation.expected_args, *args_for_multiple_calls)
+          message << "\nDiff:#{diff}" unless diff.empty?
+        end
+
+        message
+      end
+
       def diff_message(expected_args, actual_args)
         formatted_expected_args = expected_args.map do |x|
           RSpec::Support.rspec_description_for_object(x)
@@ -286,13 +285,10 @@ module RSpec
         RSpec::Support.is_a_matcher?(arg) && arg.respond_to?(:description)
       end
 
-      def format_received_args(*args)
-        return "(no args)" if args.empty? || args.first.empty?
-        return formatted_single_message(args) if single_message?(args)
-        grouped_list = grouped_args(args).map do |arg, index|
-          [formatted_single_message([arg]), group_count(index, args)].join
-        end
-        grouped_list.join("\n            ")
+      def format_received_args(*args_for_multiple_calls)
+        grouped_args(args_for_multiple_calls).map do |args_for_one_call, index|
+          [format_args(*args_for_one_call), group_count(index, args_for_multiple_calls)].join
+        end.join("\n            ")
       end
 
       def count_message(count, expectation_count_type=nil)
@@ -305,16 +301,8 @@ module RSpec
         "#{count} time#{count == 1 ? '' : 's'}"
       end
 
-      def formatted_single_message(args)
-        "(" + args.first.map(&:inspect).join(", ") + ")"
-      end
-
       def grouped_args(args)
         Hash[args.group_by { |x| x }.map { |k, v| [k, v.count] }]
-      end
-
-      def single_message?(args)
-        args.size == 1
       end
 
       def group_count(index, args)
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -32,6 +32,9 @@ module RSpec
           end
         end
       end
+
+      def unadvise(_)
+      end
     end
 
     # Represents an individual method stub or message expectation. The methods
@@ -457,6 +460,10 @@ module RSpec
 
         def advise(*args)
           similar_messages << args
+        end
+
+        def unadvise(args)
+          similar_messages.delete_if { |message| args.include?(message) }
         end
 
         def generate_error
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -98,11 +98,15 @@ module RSpec
 
       # @private
       def check_for_unexpected_arguments(expectation)
-        @messages_received.each do |(method_name, args, _)|
-          next unless expectation.matches_name_but_not_args(method_name, *args)
+        return if @messages_received.empty?
 
-          raise_unexpected_message_args_error(expectation, *args)
-        end
+        # Check if any of the arguments were expected.
+        return if @messages_received.any? do |(method_name, args, _)|
+                    !expectation.matches_name_but_not_args(method_name, *args)
+                  end
+        unexpected_args = @messages_received.map { |_, args, _| args }
+
+        raise_unexpected_message_args_error(expectation, *unexpected_args)
       end
 
       # @private
@@ -171,11 +175,11 @@ module RSpec
           expectation.advise(*args) if null_object? unless expectation.expected_messages_received?
 
           if null_object? || !has_negative_expectation?(message)
-            raise_unexpected_message_args_error(expectation, *args)
+            raise_unexpected_message_args_error(expectation, args)
           end
         elsif (stub = find_almost_matching_stub(message, *args))
           stub.advise(*args)
-          raise_missing_default_stub_error(stub, *args)
+          raise_missing_default_stub_error(stub, args)
         elsif Class === @object
           @object.superclass.__send__(message, *args, &block)
         else
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -104,9 +104,8 @@ module RSpec
         return if @messages_received.any? do |(method_name, args, _)|
                     !expectation.matches_name_but_not_args(method_name, *args)
                   end
-        unexpected_args = @messages_received.map { |_, args, _| args }
 
-        raise_unexpected_message_args_error(expectation, *unexpected_args)
+        raise_unexpected_message_args_error(expectation, *messages_arg_list)
       end
 
       # @private
@@ -146,6 +145,11 @@ module RSpec
       end
 
       # @private
+      def messages_arg_list
+        @messages_received.map { |_, args, _| args }
+      end
+
+      # @private
       def has_negative_expectation?(message)
         method_double_for(message).expectations.find { |expectation| expectation.negative_expectation_for?(message) }
       end
@@ -170,6 +174,7 @@ module RSpec
           end
           stub.invoke(nil, *args, &block)
         elsif expectation
+          expectation.unadvise(messages_arg_list)
           expectation.invoke(stub, *args, &block)
         elsif (expectation = find_almost_matching_expectation(message, *args))
           expectation.advise(*args) if null_object? unless expectation.expected_messages_received?
--- a/spec/rspec/mocks/matchers/have_received_spec.rb
+++ b/spec/rspec/mocks/matchers/have_received_spec.rb
@@ -141,9 +141,106 @@ module RSpec
           let(:the_dbl) { double(:expected_method => nil) }
 
           before do
-            the_dbl.expected_method
-            the_dbl.expected_method
-            the_dbl.expected_method
+            the_dbl.expected_method(:one)
+            the_dbl.expected_method(:two)
+            the_dbl.expected_method(:one)
+          end
+
+          context "when constrained by `with`" do
+            it 'only considers the calls with matching args' do
+              expect(the_dbl).to have_received(:expected_method).with(:one).twice
+              expect(the_dbl).to have_received(:expected_method).with(:two).once
+            end
+
+            context "when the message is received without any args matching" do
+              it 'includes unmatched args in the error message' do
+                expect {
+                  expect(the_dbl).to have_received(:expected_method).with(:three).once
+                }.to raise_error(Expectations::ExpectationNotMetError,
+                                 a_string_including("expected: (:three)",
+                                                    "got:", "(:one) (2 times)",
+                                                            "(:two) (1 time)"))
+              end
+            end
+
+            context "when the message is received too many times" do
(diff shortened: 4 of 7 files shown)


JonRowe:
Thanks for working on this, I have some minor feedback which I'll put inline. Don't worry about the appveyor failure it's failed to start the build entirely so...

spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -141,9 +141,41 @@ module Mocks
...
+              expect(the_dbl).to have_received(:expected_method).with(:one).twice
+              expect(the_dbl).to have_received(:expected_method).with(:two).once
+            end
+
+            context "when the message is received without any args matching" do
+              it 'includes unmatched args in the error message' do
+                expect {
+                  expect(the_dbl).to have_received(:expected_method).with(:three).once
+                }.to raise_error(/#{Regexp.escape("expected: (:three)")}.*#{
+                                    Regexp.escape("got: ([:one, :two])")}/m)
```
JonRowe: This doesn't read very well, if you see @myronmarston's specs in #841, you'll see you can compose this message using `a_string_including` with the two desired fragments, I'd prefer if you refactored these specs to use that pattern

spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -141,9 +141,41 @@ module Mocks
...
+              end
+            end
+
+            context "when the message is received too many times" do
+              it 'includes the counts of calls with matching args in the error message' do
+                expect {
+                  expect(the_dbl).to have_received(:expected_method).with(:one).once
+                }.to raise_error(/expected: 1 time.*received: 2 times/m)
```
JonRowe: As above I'd prefer a composed string matcher here.

spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -141,9 +141,41 @@ module Mocks
...
+              end
+            end
+
+            context "when the message is received too few times" do
+              it 'includes the counts of calls with matching args in the error message' do
+                expect {
+                  expect(the_dbl).to have_received(:expected_method).with(:two).twice
+                }.to raise_error(/expected: 2 times.*received: 1 time/m)
```
JonRowe: And here :)

lib/rspec/mocks/proxy.rb
```diff
@@ -98,10 +98,19 @@ def replay_received_message_on(expectation, &block)
...
+        # Return unless message never received any of the specified args
+        return unless all_arguments_unexpected?(expectation)
+        unexpected_args = @messages_received.map { |_method_name, args, _| args }.uniq.flatten
 
-          raise_unexpected_message_args_error(expectation, *args)
+        raise_unexpected_message_args_error(expectation, unexpected_args)
+      end
+
+      # @private
+      def all_arguments_unexpected?(expectation)
```
JonRowe: I'd think I'd prefer we do this inline, in the method above e.g:

```Ruby
def check_for_unexpected_arguments(expectation)
  return if @messages_received.empty?
  return if @messages_received.none? do |(method_name, args, _)|
              expectation.matches_name_but_not_args(method_name, *args)
            end
  # ...
end
```

I switched the `all?` for `none?` rather than an unless, I think thats the same semantics...

myronmarston: It's not quite the same semantics.  `all?` and `none?` are not strict opposites. Consider `[].all?(&:any_predicate?)` vs `[].none?(&:any_predicate?)` -- they return the same value.

johnceh:
Good to know, the appveyor failure had me scratching my head a bit.  Thanks for the the feedback @JonRowe.  All seems like good input, so I'll touch this up and get an update in soon.

lib/rspec/mocks/proxy.rb
JonRowe: yes but I think it's equivalent here, in fact we can problem remove the return above it too

johnceh:
Updated with some of the suggested refactoring. 

- I switch form `all?` to `none?` didn't work as expected, but using `any?` did
- Additional method removed in favor of inline return
- Couldn't get fail_with to play nice with a_string_containing, but raise error specifying error type worked.

Seemed like some connection issues might have been happening during Travis CI process, so perhaps a rebuild would be needed. Some failures were because to position of `end` it seems. 

     return if @messages_received.any? do |(method_name, args, _)|
                    !expectation.matches_name_but_not_args(method_name, *args)
            end

Should I align it under the return in order to pass?

JonRowe:
> Should I align `end` under the return in order to pass? Seems that would be against what I would expect.

No it needs to align with `@messages`, `end` is expected to line up with the start of it's statement, which is what it's being passed to as a block so the start of `@messages_received.any?`

johnceh:
Looks like that did the trick @JonRowe , thanks.

lib/rspec/mocks/proxy.rb
```diff
@@ -98,11 +98,16 @@ def replay_received_message_on(expectation, &block)
...
+        return if @messages_received.empty?
 
-          raise_unexpected_message_args_error(expectation, *args)
-        end
+        # Check if any of the arguments were expected.
+        return if @messages_received.any? do |(method_name, args, _)|
+                    !expectation.matches_name_but_not_args(method_name, *args)
+                  end
+
+        unexpected_args = @messages_received.map { |_, args, _| args }.uniq.flatten
```
myronmarston: The `uniq.flatten` here concerns me.  With this in place, consider this example:

``` ruby
it 'groups the "got" arguments based on the method call that included them' do
  dbl = double(:expected_method => nil)
  dbl.expected_method(:one, :four)
  dbl.expected_method(:two, :four)
  dbl.expected_method(:one, :four)

  expect(dbl).to have_received(:expected_method).with(:three, :four).once
end
```

...which fails with:

```
     Failure/Error: expect(dbl).to have_received(:expected_method).with(:three, :four).once
       Double received :expected_method with unexpected arguments
         expected: (:three, :four)
              got: ([:one, :four, :two, :four])
       Diff:
       @@ -1,2 +1,2 @@
       -[:three, :four]
       +[[:one, :four, :two, :four]]
```

...which is quite confusing.  It makes it seem that one call was received with those four arguments.  Instead, I'd like to see something like:

```
Failure/Error: expect(dbl).to have_received(:expected_method).with(:three, :four).once
  Double received :expected_method with unexpected arguments
    expected: (:three, :four)
         got: (:one, :four) (2 times)
              (:two, :four) (1 time)
```

(And perhaps two separate diffs, one per call...or maybe no diff at all; I'm not sure yet.)

myronmarston:
In #841 there's a failing spec in `once_counts_spec.rb` that concerns this same behavior for `expect(...).to receive(...)` (rather than for `have_received` as you've dealt with here).  Any chance you'd want to tackle that in this PR as well since it's related?  If you don't want to (or don't have the time), that's fine; we'll just need to open another issue for that so we don't lose track of it.

johnceh:
Good point on the grouped arguments, didn't think of that scenario. I'll try a few different versions to see if one looks more appealing then the other.

If you won't be needing it right away, I don't mind taking a stab at the `receive`.  Just learning the project as I go.

johnceh:
I made a new method to raise the errors to since we were a little unsure of how exactly we want the formatting to be done. I also couldn't figure out how to distinguish between with/non-with errors, so things were getting messy trying to use the existing error formatting methods. New method seemed the like better option.

I took a look at the `once_counts_spec` test as well, but we might need to tackle that in a different PR if that's ok. Flow didn't seem as straight-forward on that one and I don't know when I'd be able to get back around to it.

lib/rspec/mocks/error_generator.rb
```diff
@@ -62,6 +62,18 @@ def raise_unexpected_message_args_error(expectation, *args)
...
+      def raise_unexpected_message_with_args_error(expectation, *args)
+        expected_args = format_args(*expectation.expected_args)
+        actual_args = format_unexpected_with_args(*args)
+        diff = diff_message(expectation.expected_args, args)
+
+        message = default_error_message(expectation, expected_args, actual_args)
+        message << "\nDiff:#{diff}" unless diff.empty?
+
+        __raise message
+      end
```
myronmarston: This method is nearly identical to the one above it (`raise_unexpected_message_args_error`).  From the names, it's not clear to me what the difference is intended to be.  Implementation-wise, the only difference seems to be the way the args are treated (one has `actual_args = format_received_args(*args)` and the other has `actual_args = format_unexpected_with_args(*args)`).  A few thoughts:

* Can we combine these into one method?  I think the difference is that the existing one expects `args` to be a list of the args for a single method call, where as for this method, `args` is multiple lists of args from multiple method calls...is that right?  Given that, I think we could standardize on _one_ method, where the existing callers of the old method would wrap the args in an array (`[args]`) in order to cause them to be treated as the args of a single method call.
* There's a lot of splatting and un-splatting going on here.  I know you're just following the patterns you see in this file but I think it makes sense to move away from that for new code like this -- I believe there's a cost associated with splatting and un-splatting that's not there when you just pass an array as-is.

myronmarston: Also, if `args` is an array of argument-arrays (e.g. one argument array per method call), as I understand it, I think a different name would be appropriate, something like `args_for_multiple_calls` or something similar.

lib/rspec/mocks/error_generator.rb
```diff
@@ -291,6 +303,19 @@ def format_received_args(*args)
         args.empty? ? "(no args)" : "(" + received_arg_list(*args) + ")"
       end
 
+      def format_unexpected_with_args(*args)
+        with_args = Hash[Array(args.first).group_by { |x| x }.map { |k, v| [k, v.count] }]
```
myronmarston: Why are you passing `args.first` to `Array`?  I thought that `args` is a list of arg lists (one per method call), so this looks to me like it just is the arg list from the first call...but that doesn't make sense and it works so there's something I'm not understanding here.

lib/rspec/mocks/error_generator.rb
```diff
@@ -291,6 +303,19 @@ def format_received_args(*args)
...
+        received_with_arg_list(*with_args)
+      end
+
+      def received_with_arg_list(*args)
+        list = args.map do |a, i|
+          times_count = args.size > 1 ? '(' + times(i) + ')' : ''
+          ["(#{received_arg_list(*a)})", "#{times_count}"].join(" ")
```
myronmarston: Putting these in an array only to join them seems unnecessary (and adds additional object allocations), so I'd prefer to see this just as:

``` ruby
"(#{received_arg_list(*a)}) #{times_count}"
```

myronmarston: Also, I think you want to use `format_received_args` rather than `received_arg_list`, as the former takes care of putting `"no args"` in the string when there are none.  Having a spec that covers this case would be nice, although if you do refactor things above to have only one `raise_unexpected_message...` method (as I suggested above), then I suspect that the existing specs that cover that method will cover this case and will only pass if you switch to `format_received_args` -- in which case there's no need for an explicit spec.

Given that it's easy to use `received_arg_list` when `format_received_args` should be used, I'm thinking that maybe we should roll the `received_arg_list` logic in to `format_received_args` so that it's not even available to be called.

Thoughts?

myronmarston:
> I took a look at the once_counts_spec test as well, but we might need to tackle that in a different PR if that's ok. Flow didn't seem as straight-forward on that one and I don't know when I'd be able to get back around to it.

That's fine.

lib/rspec/mocks/error_generator.rb
johnceh: Was definitely trying to take the path of least resistance by keeping with the same convention without making too many changes to existing methods.  The new method was because of the variance of arguments that were coming in made it difficult to standardize the response. If it's okay to adjust existing callers to use a new combined  method, then it should open up an easier path to make it into one method as you suggest.

lib/rspec/mocks/error_generator.rb
johnceh: I was initially trying to process all the responses through a shared method, and that was just a way to ensure it would always call group_by on a Array. Left it in after for better fault tolerance.

lib/rspec/mocks/error_generator.rb
johnceh: I initially had the code setup just like that, but then single expectations would look like this
    
    expected: (:three, :four)
         got: (:one, :four) (1 time)

vs as it is now

    expected: (:three, :four)
         got: (:one, :four)

I'm okay switching it to the one-liner, or maybe adjusting the `times` method to accept an option for ignoring excluding the text if it's 1 time.

As for `format_received_args `,  I'm usually in favor of splitting out methods more then combining, but perhaps this is a situation where we can make it into one more efficient method. I'll see where things lead after I do some of the refactoring on the raise method.

myronmarston: Maybe we can have it only display the "n times" bit only if the message was received multiple times -- so any time that `(1 time)` would be displayed, we can suppress it (since that's what a user would normally assume and it's not needed in that case).  Then there's no need for an extra option.

lib/rspec/mocks/error_generator.rb
myronmarston: Doesn't `args.first` just return the args were passed the first time the message was received, and ignore all the other times?  Or am I misunderstanding the structure of `args`?

johnceh: When I tried using the existing method, some arguments came in as `[[[:one],[:two]]]`, while others came in as `[1]` or `["wrong"]`.  `[1].first.group_by` would cause an exception, so I wrapped it up in `Array` to prevent that.   After creating a new method instead, I could have taken out the `Array` since it was always receiving consistent arguments, but I left it as is so it would still process if something unexpected came in.  

I'll re-evaluate if this will be needed after the callers get adjusted for the new method.  Might not need it anymore after that.

lib/rspec/mocks/error_generator.rb
johnceh: Sounds good to me.  The only odd scenario would probably when it's mixed in with other `(n times)` but it still makes sense I think.

    expected: (:three, :four)
         got: (:one, :four) (2 times)
              (:one, :two)
              (:one, :three) (3 times)

myronmarston: That's a good point.  I think it would be nice to include `(1 time)` in that case.  Maybe we can exclude the "n times" bit if it was exactly 1 time for each?

JonRowe:
(this was updated @myronmarston)

johnceh:
Looks like adding in the 2nd part fix is causing CI to throw a complexity warning. Any suggestions, or should I remove the 2nd part of this fix?

myronmarston:
Sorry, @johnceh, I didn't know you had pushed new commits since GitHub doesn't notify repo owners when PRs get new commits.  Next time please post a comment saying "Ready for another review" (or whatever) and that'll let me know.

> Looks like adding in the 2nd part fix is causing CI to throw a complexity warning. Any suggestions, should I remove the 2nd part of this fix, or do we tackle a complexity refactor on the `message_received` method?

To repro that failure locally, run `bundle exec rubocop lib`.

Anyhow, that method looks like it should be refactored at some point, but I consider that out of scope for this PR.  So I recommend disabling rubocop there:

``` ruby
      # rubocop:disable all
      def message_received(message, *args, &block)
        # ...
      end
      # rubocop:enable all
```

I'll try to review the rest of the PR later today or tomorrow.

johnceh:
Rubocop change is in, and it's ready for another review @myronmarston

Thanks for the tip on `bundle exec rubocop lib`, I wasn't sure how to go about testing if I tried making changes.

lib/rspec/mocks/error_generator.rb
```diff
@@ -50,36 +50,35 @@ def raise_unexpected_message_error(message, *args)
       end
 
       # @private
-      def raise_unexpected_message_args_error(expectation, *args)
+      def message_error(expectation, *args)
```
myronmarston: If I'm reading this right, `message_error` is a helper method, and not something meant to be called directly, right?  Can it be moved into the `private` section of the class to make that more explicit?

Also, `*args` is conventionally used for the list of arguments for one method call.  Here it looks like it's a list-of-argument-lists (unless I'm misreading the code), but the variable name doesn't reflect that.  If that's the case, I think `args_for_multiple_calls` is a better name.  This would apply to any other methods you're touching that similarly accept a list-of-argument-lists for multiple calls.

Thoughts?

lib/rspec/mocks/error_generator.rb
```diff
@@ -288,11 +287,12 @@ def arg_has_valid_description?(arg)
       end
 
       def format_received_args(*args)
-        args.empty? ? "(no args)" : "(" + received_arg_list(*args) + ")"
-      end
-
-      def received_arg_list(*args)
-        args.map(&:inspect).join(", ")
+        return "(no args)" if args.empty? || args.first.empty?
```
myronmarston: Why the need to check `args.empty?` and `args.first.empty?`

lib/rspec/mocks/error_generator.rb
```diff
@@ -304,6 +304,22 @@ def count_message(count, expectation_count_type=nil)
...
+        "(" + args.first.map(&:inspect).join(", ") + ")"
+      end
+
+      def grouped_args(args)
+        Hash[args.group_by { |x| x }.map { |k, v| [k, v.count] }]
+      end
+
+      def single_message?(args)
+        args.size == 1
+      end
```
myronmarston: This can be simplified to `args.one?`.

lib/rspec/mocks/error_generator.rb
```diff
@@ -304,6 +304,22 @@ def count_message(count, expectation_count_type=nil)
       def times(count)
         "#{count} time#{count == 1 ? '' : 's'}"
       end
+
+      def formatted_single_message(args)
+        "(" + args.first.map(&:inspect).join(", ") + ")"
+      end
```
myronmarston: If the args here are rspec-matchers this will wrongly use `inspect` rather than `description`.  The logic for this is already handled by `format_args` and `args_list` so it'd be best to re-use those and not create yet another method that formats an arg list.

lib/rspec/mocks/proxy.rb
```diff
@@ -166,22 +174,24 @@ def message_received(message, *args, &block)
           end
           stub.invoke(nil, *args, &block)
         elsif expectation
+          expectation.unadvise(messages_arg_list) if defined? expectation.similar_messages
```
myronmarston: The `defined?` check is a code smell (especially since the relationship between `unadvise` and `similar_messages` is not obvious).  Instead, I think you can define `unadvise` as a no-op on `RSpec::Mocks::SimpleMessageExpectation`.

lib/rspec/mocks/error_generator.rb
```diff
@@ -288,11 +287,12 @@ def arg_has_valid_description?(arg)
...
-
-      def received_arg_list(*args)
-        args.map(&:inspect).join(", ")
+        return "(no args)" if args.empty? || args.first.empty?
+        return formatted_single_message(args) if single_message?(args)
+        grouped_list = grouped_args(args).map do |arg, index|
+          [formatted_single_message([arg]), group_count(index, args)].join
+        end
+        grouped_list.join("\n            ")
       end
```
myronmarston: Maybe I'm missing something, but it looks like this method can be simplified to:

``` ruby

def format_received_args(*args_for_multiple_calls)
  grouped_args(args_for_multiple_calls).map do |args_for_one_call, index|
    [format_args(*args_for_one_call), group_count(index, args_for_multiple_calls)].join
  end.join("\n            ")
end
```

(Notice also the renaming of the arguments, which I think are more clear.  You're version also has `arg`, which is actually a list of args and not a single arg, right?)

myronmarston:
OK, I need to head to bed (I'm exhausted...) but I haven't finished my review.  I'll try to wrap it up tomorrow or the next day.

lib/rspec/mocks/error_generator.rb
johnceh: I somehow missed `format_args` completely.  `format_received_args` had the same code (basically) so I was trying to keep that working when refactoring. Using `format_args` instead should work much better.

lib/rspec/mocks/error_generator.rb
johnceh: Yup, good catch. the `# @private` must have visually threw me off on that. Making private would be best.

Since this might have arguments from a single call as well, do you think `args_for_calls` would fit better?  That would then cover single and multiple calls and still represent a collection of arguments.  I'm okay with either name, but if a single call was issued and it was called `args_for_multiple_calls`, it might not technically be correct.

lib/rspec/mocks/error_generator.rb
johnceh: Utilizing `format_args` makes your example work perfectly.  Similar to my comment above, what do you think of using `args_for_calls` and `args_for_call` vs `args_for_multiple_calls` and `args_for_one_call`.  This might be used on a collection with just one call, so something like `args_for_calls` could refer to single or multiple calls.  Thoughts?

lib/rspec/mocks/error_generator.rb
johnceh: Oops, useless function. I'll be removing.

johnceh:
Sorry @myronmarston, didn't mean to ruin your Easter weekend. No rush on my side of things. :)

johnceh:
Ready for another review @myronmarston

I think I got most of your suggestions in, and it's looking a lot better.

lib/rspec/mocks/message_expectation.rb
```diff
@@ -32,6 +32,8 @@ def verify_messages_received
           end
         end
       end
+
+      def unadvise(_) ; end
```
JonRowe: I think I'd prefer this as a two line empty method, do we use this pattern anywhere else?

johnceh: I went off this guide (https://github.com/bbatsov/ruby-style-guide#no-single-line-methods), but  consistency is key, so it would make sense to match existing style.

Something like this?

    def unadvise(*_args)
    end

JonRowe: def unadvise(_)
    end

is fine

johnceh: Change is in @JonRowe, thanks for the feedback

lib/rspec/mocks/proxy.rb
```diff
@@ -142,6 +145,11 @@ def received_message?(method_name, *args, &block)
       end
 
       # @private
+      def messages_arg_list
```
johnceh: `messages_arg_list` was the first name that came to mind, but I don't really like it.  Should I leave it, or perhaps change to `message_args`, `args_from_messages`, or perhaps just swap `for` and `from` to make `args_from_multiple_calls`. Might be too confusing with the name being so close to `args_for_multiple_calls`.

myronmarston:
Thanks @johnceh -- you did great work here!

PR #920, opened 2015-04-09
Registering doubles using ObjectInspector

sivagollapalli:


myronmarston:
rspec-mocks should register itself with rspec-support so that doubles are automatically formatted in expectation failures well.

spec/rspec/mocks/test_double_spec.rb
```diff
@@ -48,6 +48,19 @@ module Mocks
           end
         end
       end
+
+      describe ".register" do
+        pending "able to register test doubles" do
+          user = double('User')
+          allow(user).to receive(:name).and_return("FooBar")
+
+          RSpec::Mocks::TestDouble.register user.class do |user|
```
myronmarston: `RSpec::Mocks::TestDouble` does not have a `register` method.  The goal is register the it with `RSpec::Support::ObjectInspector`, right?

JonRowe: That's my reading of the situation too

sivagollapalli:
@myronmarston Spec modified as per your suggestion. Let me know if any changes. If spec is correct, then I will go ahead with implementation.

spec/rspec/mocks/test_double_spec.rb
```diff
@@ -48,6 +48,17 @@ module Mocks
...
+
+      pending "test doubles able to register themselves with ObjectInspector" do
+        allow_any_instance_of(TestDouble).to receive(:name).and_return("FooBar")
+        double('User')
+
+        RSpec::Support::ObjectInspector.register TestDouble do |user|
+          "User (#{user.name})"
+        end
+
+        expect(double.inspect).to eq("User (FooBar)")
```
JonRowe: This is still wrong for a spec, as it actually implements it, so the spec should just be:

```Ruby
it 'registers itself with the ObjectInspector' do
  expect(double('FooBar').inspect).to eq 'Double (FooBar)'
end
```
and then the implementation of `TestDouble` needs to be modified to call:

```Ruby
RSpec::Support::ObjectInspector.register TestDouble,  do |dbl|
  "Double (#{dbl.instance_variable_get('@name')})"
end
```

Although I'd argue it'd be nicer if the `ObjectInspector` could be registered with instances, then we could register specific instances when creating the doubles.

myronmarston: > This is still wrong for a spec, as it actually implements it, so the spec should just be:

> ```Ruby
it 'registers itself with the ObjectInspector' do
  expect(double('FooBar').inspect).to eq 'Double (FooBar)'
end
```

That's not right.  We're not touching the `double.inspect` method.  We're registering the type with `ObjectInspector` so that `ObjectInspector.inspect(double)` can be used to format the double in failure messages.  The spec should likewise use `ObjectInspector.inspect`.

> Although I'd argue it'd be nicer if the `ObjectInspector` could be registered with instances, then we could register specific instances when creating the doubles.

I disagree.  It would cause a memory leak.  If we registered instances, then Ruby's GC would not garbage collect those instances since there would still be a live references to then.  In addition, the `ObjectInspector` algorithm does a linear scan over each of the registered classes, and if we used instances, the number of objects to check against would grow and it would get slower.

We could of course deregister it but that seems much more complicated than just registering the class.  In addition, the object inspector logic up to this point has really been a function of what class/type the object is -- hashes are formatted one way, `Time` objects another way, etc.

Comment (JonRowe) [inline spec/rspec/mocks/test_double_spec.rb]: > That's not right.  We're not touching the `double.inspect` method.  We're registering the type with `ObjectInspector` so that `ObjectInspector.inspect(double)` can be used to format the double in failure messages.  The spec should likewise use `ObjectInspector.inspect`.

Ah yeah you're right, but still the point was that we shouldn't be registering the double in the spec, and we shouldn't be stubbing name on the double.

Comment (myronmarston) [inline spec/rspec/mocks/test_double_spec.rb]: > Ah yeah you're right, but still the point was that we shouldn't be registering the double in the spec, and we shouldn't be stubbing name on the double.

Completely agree.

**Comment by sivagollapalli** [comment]
@JonRowe I have modified the spec. Could you suggested if any changes needs to made?

@myronmarston As per above discussion we are not modifying the `inspect` method in `TestDouble` class. Instead, we use `RSpec::Support::ObjectInspector.inspect` to display failure messages. If it that case we need make a PR to rspec-expectations with this change. Am I correct?

File: spec/rspec/mocks/test_double_spec.rb [hunk 3]
```diff
@@ -48,6 +48,15 @@ module Mocks
           end
         end
       end
+
+      pending "register itself with ObjectInspector" do
+        RSpec::Support::ObjectInspector.register TestDouble do |dbl|
+          "Double (#{dbl.instance_variable_get('@name')})"
+        end
```
JonRowe: This needs to be removed, it's an implementation detail and doesn't belong in the spec.

myronmarston:
> If it that case we need make a PR to rspec-expectations with this change. Am I correct?

rspec-expectations has already been updated to use `RSpec::Support::ObjectInspector` all over the place.

sivagollapalli:
@JonRowe I just ran specs on my machine, all are passing. But on travis it fails. I couldn't able to find where I went wrong. Could check my commit?

myronmarston:
@sivagollapalli -- thanks for doing this.  I plan to take your commit and add some more to get this passing.  I think that'll be easier than going back-and-forth at this point.

myronmarston:
Superceded by #932.

myronmarston:
Thanks, @sivagollapalli.  I pulled your change into a new PR (#932) and made some further changes.

PR #922, opened 2015-04-09
Pr 918 followups

myronmarston:
This is some followups to #918 (/cc @johnceh) and some other general improvements to the error generator that I noticed could be made while reviewing that PR.

For whoever reviews this: it'll probably be easier to go commit-by-commit since it's kind of a grab bag of improvements.


JonRowe:
All looks good to me  :)

PR #924, opened 2015-04-10
Prefer `anything == object` over `object == anything`.

myronmarston:
Many users have objects that implement `==` wrongly
in a way that can raise errors when given an object
of an unexpected type.  This avoids that situation
by checking equality with `anything` instead.

See rspec/rspec-expectations#732 for one case of this.


JonRowe:
LGTM

PR #925, opened 2015-04-10
Matching arguments with RSpec matchers

samnang:
It's just an improvement in the doc. It's base on my problem that I didn't realize that we could use RSpec matcher in matching arguments on this page https://www.relishapp.com/rspec/rspec-mocks/v/3-2/docs/setting-constraints/matching-arguments

Base on conversation with @myronmarston [1], I suggest to add another scenario to make it clearer even those there is no any different from using custom matcher which is we already have scenario in term of behavior correctness. 

 https://twitter.com/samnangchhun/status/586069627207192576

### DIFF
--- a/features/setting_constraints/matching_arguments.feature
+++ b/features/setting_constraints/matching_arguments.feature
@@ -46,6 +46,31 @@ Feature: Matching arguments
       |     expected: (1, anything, /bar/)               |
       |          got: (1, nil, "other")                  |
 
+  Scenario: Using a RSpec matcher
+    Given a file named "rspec_matcher_spec.rb" with:
+      """ruby
+      RSpec.describe "Using a RSpec matcher" do
+        let(:dbl) { double }
+        before { expect(dbl).to receive(:foo).with(a_collection_containing_exactly(1, 2)) }
+
+        it "passes when the args match" do
+          dbl.foo([2, 1])
+        end
+
+        it "fails when the args do not match" do
+          dbl.foo([1, 3])
+        end
+      end
+      """
+    When I run `rspec rspec_matcher_spec.rb`
+    Then it should fail with the following output:
+      | 2 examples, 1 failure                               |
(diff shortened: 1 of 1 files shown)


features/setting_constraints/matching_arguments.feature
```diff
@@ -46,6 +46,31 @@ Feature: Matching arguments
       |     expected: (1, anything, /bar/)               |
       |          got: (1, nil, "other")                  |
 
+  Scenario: Using a RSpec matcher
+    Given a file named "rspec_matcher_spec.rb" with:
+      """ruby
+      RSpec.describe "Using a RSpec matcher" do
+        let(:dbl) { double }
+        before { expect(dbl).to receive(:foo).with(contain_exactly(1, 2)) }
```
myronmarston: Let's use `a_collection_containing_exactly` rather than `contain_exactly`.  The former is an alias rspec-expectations provides for when you need a noun form to make grammatical sense (it also changes how the matcher is described):

http://rspec.info/documentation/3.2/rspec-expectations/RSpec/Matchers.html#contain_exactly-instance_method

features/setting_constraints/matching_arguments.feature
```diff
@@ -46,6 +46,31 @@ Feature: Matching arguments
...
+
+        it "passes when the args match" do
+          dbl.foo([1, 2])
```
myronmarston: Given that `contain_exactly` allows the array to be in any order, might be worth doing `db.foo([2, 1])` here -- otherwise, the expectation above doesn't really need the additional flexibility `contain_exactly` provides.

myronmarston:
Thanks, @samnang!  I'll be happy to merge once you address the two comments above.

samnang:
Updated base on feedbacks. Thank @myronmarston for you help and explanation.

myronmarston:
Thanks!

Issue #916, opened 2015-03-26
Spies do not support ordering with interleaved calls


myronmarston:
``` ruby
RSpec.describe "Specifying ordering for calls that are interleaved with each other" do
  it 'fails when using `have_received`' do
    dbl = spy

    dbl.one
    dbl.one
    dbl.two
    dbl.one
    dbl.two

    expect(dbl).to have_received(:one).twice.ordered
    expect(dbl).to have_received(:two).once.ordered
    expect(dbl).to have_received(:one).once.ordered
    expect(dbl).to have_received(:two).once.ordered
  end

  it 'passes when using `receive`' do
    dbl = double

    expect(dbl).to receive(:one).twice.ordered
    expect(dbl).to receive(:two).once.ordered
    expect(dbl).to receive(:one).once.ordered
    expect(dbl).to receive(:two).once.ordered

    dbl.one
    dbl.one
    dbl.two
    dbl.one
    dbl.two
  end
end
```

The second spec passes, but the first fails:

```
Specifying ordering for calls that are interleaved with each other
  fails when using `have_received` (FAILED - 1)
  passes when using `receive`

Failures:

  1) Specifying ordering for calls that are interleaved with each other fails when using `have_received`
     Failure/Error: expect(dbl).to have_received(:one).twice.ordered
       (Double).one(*(any args))
           expected: 2 times with any arguments
           received: 3 times with any arguments
```

I originally put a failing spec in #881 but don't have the time to fix it so I figured I'd convert it into an issue in the hope that someone else would pick it up.

johnceh:
@myronmarston Potentially found another scenario or separate issue using a similar test.

    it 'fails when using `have_received` with unexpected message' do
      dbl = spy

      dbl.one
      dbl.two

      expect {
        expect(dbl).to have_received(:one).once.ordered
      }.to raise_error(/received unexpected message :two/m)
    end

    it 'passes when using `receive` with unexpected message' do
      expect {
        dbl = double

        expect(dbl).to receive(:one).once.ordered

        dbl.one
        dbl.two
        }.to raise_error(/received unexpected message :two/m)
    end

Second spec passes, but the first fails

Failures:

```
  1) RSpec::Mocks::Matchers::HaveReceived expect(...).to have_received ordered when used on individually allowed messages fails when using `have_received` with unexpected message
     Failure/Error: expect {
       expected Exception with message matching /received unexpected message :two/m but nothing was raised
```

[...]

PR #928, opened 2015-04-12
Issue 916/ordered spies bug

johnceh:
<del># Need some discussion (Info Below)</del>

# Why?
Fix for Issue #916 - Spies were not tracking order properly and raised error incorrectly for alternating groups. It was including all messages/expectations/invocations when evaluating order on each expectation.

# What Changed?
Added limiting range for messages/expectations/invocations for order checks so just current expectations from group are evaluated. Tests by @myronmarston in issue added in.

<del>
# Discussions Needed
1) By limiting range to running expectation, the error message now focuses on what it's expecting vs what happened.  Changing this will cause failures to occur on previously successful tests.
Before, the error would say that one is out of order, but it will change to say two is out of order since the range limits what messages/expectations it checks for. Previous code left commented in for now for visibility.

    allow(the_dbl).to receive(:one)
    allow(the_dbl).to receive(:two)

    the_dbl.one
    the_dbl.two

    expect(the_dbl).to have_received(:two).ordered
    expect(the_dbl).to have_received(:one).ordered
              
    # Raised error before - received :one out of order
    # Raised error after - received :two out of order

Before I dedicate any more time to this, I thought I'd check if I'm even on the right track with this fix. Also I didn't want to make assumptions on if we want to make this change or not, so if it needs to be backwards compatible, then I will keep looking into it after some feedback.

2) Had to swap checks in `expected_messages_received_in_order` for expected messages and order so order is checked first.  This was needed so that if an unexpected message comes in first, it doesn't error on the fact that 0 expected messages were received, but instead that the expectation was for a different message. It's a similar problem as above with breaking existing tests since an order error will be displayed now vs a count error.  The travis ci failure is due to this change, and it's use case makes sense for count I think. Anyway, I thought I'd push it up for now and see what everybody thought before I made to many more changes.  Previous code left commented in for now for visibility.
    

3) This did bring to light a potential issue for `receive` messages as well (which had the passing test in the issue). I added a pending test that is failing and was hoping someone can verify what the expected result should be and if a similar change will be needed where order is checked before message expectations.
</del>

I found some other inconsistent responses with varied scenarios, so just ignore my comments for now and I'll try to get this to respond more like the `receive`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -26,6 +26,9 @@ Bug Fixes:
   arguments. Previously it confusingly listed the arguments as being
   mis-matched (even when the double was allowed to receive with any
   args) rather than listing the count. (John Ceh, #918)
+* Fix `any_args`/`anything` support so that we avoid calling `obj == anything`
+  on user objects that may have improperly implemented `==` in a way that
+  raises errors. (Myron Marston, #924)
 
 ### 3.2.1 / 2015-02-23
 [Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.2.0...v3.2.1)
--- a/features/basics/spies.feature
+++ b/features/basics/spies.feature
@@ -125,5 +125,5 @@ Feature: Spies
       |           received: 2 times with any arguments                                                   |
       |                                                                                                  |
       |  2) An invitiation fails when an order constraint is not satisifed                               |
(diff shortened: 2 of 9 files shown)


johnceh:
Better to leave open while I work out a few things, or close and reopen when ready?

myronmarston:
It's fine to leave this open unless you expect it to be weeks before you make more progress.

PR #929, opened 2015-04-12
Avoid using bare raise_error

JonRowe:
We (or are going to) recommend people don't use a bare `raise_error` so we shouldn't either :D

### DIFF
--- a/spec/rspec/mocks/any_instance_spec.rb
+++ b/spec/rspec/mocks/any_instance_spec.rb
@@ -160,7 +160,7 @@ module RSpec
           it "fails the spec with an expectation error when the arguments do not match" do
             expect do
               klass.new.foo(:param_one, :param_three)
-            end.to(raise_error(RSpec::Mocks::MockExpectationError))
+            end.to fail
           end
         end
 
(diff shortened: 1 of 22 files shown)


myronmarston:
These could all be `expect { }.to fail` given we've defined a `fail` helper method:

https://github.com/rspec/rspec-mocks/blob/40911d52d71fbc6283568356113789ef85b39343/spec/support/matchers.rb#L3-L5

JonRowe:
I changed mine and made them all consistent :P

spec/rspec/mocks/any_instance_spec.rb
```diff
@@ -637,7 +637,7 @@ def inspect
                 expect(klass).to receive(:woot)
                 klass.new.foo
                 verify_all
-              end.to(raise_error(RSpec::Mocks::MockExpectationError) do |error|
+              end.to(fail_with do |error|
```
myronmarston: It seems odd to not pass any args to `fail_with`, given that the "with" is intended to read together with the arguments.  Instead, I think this would read better as `fail` w/ a block.

spec/rspec/mocks/double_spec.rb
```diff
@@ -78,7 +78,7 @@ module Mocks
         expected_error_line = __LINE__; expect(@double).to receive(:wont_happen).with("x", 3)
         expect {
           verify @double
-        }.to raise_error(RSpec::Mocks::MockExpectationError) { |e|
+        }.to fail_with { |e|
```
myronmarston: Same here and below.

spec/rspec/mocks/any_instance_spec.rb
JonRowe: There is an argument, there's a block...

JonRowe:
Decided I didn't care enough about `fail_with(&block)` vs `fail(&block)` and changed them

myronmarston:
Thanks, merged.

PR #930, opened 2015-04-13
Issue 916 - Fixes bug for ordered spies and alternating calls

johnceh:
# Why?
Fix for Issue #916 - Spies were not tracking order properly and raised error incorrectly for alternating groups. It was including all messages/expectations/invocations when evaluating order on each expectation.

# What Changed?
Added limiting range for messages/expectations/invocations for order checks so just current expectations from group are evaluated. Tests by @myronmarston in issue added in.

### DIFF
--- a/features/basics/spies.feature
+++ b/features/basics/spies.feature
@@ -125,5 +125,5 @@ Feature: Spies
       |           received: 2 times with any arguments                                                   |
       |                                                                                                  |
       |  2) An invitiation fails when an order constraint is not satisifed                               |
-      |     Failure/Error: expect(invitation).to have_received(:deliver).with("foo@example.com").ordered |
+      |     Failure/Error: expect(invitation).to have_received(:deliver).with("bar@example.com").ordered |
       |       Double "invitation" received :deliver out of order                                         |
--- a/lib/rspec/mocks/matchers/have_received.rb
+++ b/lib/rspec/mocks/matchers/have_received.rb
@@ -107,7 +107,11 @@ module RSpec
 
         def expected_messages_received_in_order?
           mock_proxy.replay_received_message_on @expectation, &@block
-          @expectation.expected_messages_received? && @expectation.ensure_expected_ordering_received!
+          if @expectation.ordered?
+            @expectation.ensure_expected_ordering_received! && @expectation.expected_messages_received?
+          else
+            @expectation.expected_messages_received? && @expectation.ensure_expected_ordering_received!
+          end
         end
 
         def mock_proxy
--- a/lib/rspec/mocks/order_group.rb
+++ b/lib/rspec/mocks/order_group.rb
@@ -41,7 +41,8 @@ module RSpec
       end
 
       def verify_invocation_order(expectation)
-        expectation.raise_out_of_order_error unless expectations_invoked_in_order?
+        message = @invocation_order[@index] ? @invocation_order[@index].message : nil
+        expectation.raise_out_of_order_error(message) unless expectations_invoked_in_order?
         true
       end
 
@@ -55,6 +56,10 @@ module RSpec
         @expectations.empty?
       end
 
+      def message_range
+        (@index..end_range)
+      end
+
     private
 
       def remaining_expectations
@@ -66,15 +71,27 @@ module RSpec
       end
 
       def invoked_expectations
-        @expectations.select { |e| e.ordered? && @invocation_order.include?(e) }
+        @expectations[order_range].select { |e| e.ordered? && @invocation_order.include?(e) }
       end
 
       def expected_invocations
-        @invocation_order.map { |invocation| expectation_for(invocation) }.compact
+        @invocation_order[order_range].map { |invocation| expectation_for(invocation) }.compact
       end
 
       def expectation_for(message)
-        @expectations.find { |e| message == e }
+        @expectations[order_range].find { |e| message == e }
+      end
+
+      def order_range
+        (start_range..end_range)
+      end
+
+      def start_range
+        @index - (@index.zero? ? 0 : 1)
+      end
+
+      def end_range
+        @expectations.size - 1
       end
     end
   end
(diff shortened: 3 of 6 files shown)


features/basics/spies.feature
```diff
@@ -125,5 +125,5 @@ Feature: Spies
       |           received: 2 times with any arguments                                                   |
       |                                                                                                  |
       |  2) An invitiation fails when an order constraint is not satisifed                               |
-      |     Failure/Error: expect(invitation).to have_received(:deliver).with("foo@example.com").ordered |
+      |     Failure/Error: expect(invitation).to have_received(:deliver).with("bar@example.com").ordered |
```
JonRowe: Why?

johnceh: I thought the first expectation (with bar) was the one that should be throwing the error, but I realize now that the first expectation should pass since it's a spy. It's the second expectation that should actually fail. I'll need to adjust the pull request a bit to adjust for this.

lib/rspec/mocks/matchers/have_received.rb
```diff
@@ -107,7 +107,11 @@ def generate_failure_message
 
         def expected_messages_received_in_order?
           mock_proxy.replay_received_message_on @expectation, &@block
-          @expectation.expected_messages_received? && @expectation.ensure_expected_ordering_received!
+          if @expectation.ordered?
+            @expectation.ensure_expected_ordering_received! && @expectation.expected_messages_received?
```
JonRowe: Why do these flip around when ordered, why are we even checking the ordering when an expectation isn't ordered!

lib/rspec/mocks/order_group.rb
```diff
@@ -41,7 +41,8 @@ def handle_order_constraint(expectation)
       end
 
       def verify_invocation_order(expectation)
-        expectation.raise_out_of_order_error unless expectations_invoked_in_order?
+        message = @invocation_order[@index] ? @invocation_order[@index].message : nil
```
JonRowe: I think this would read better as `message = @invocation_order[@index].message if @invocation_order[@index]`

lib/rspec/mocks/proxy.rb
```diff
@@ -266,6 +266,10 @@ def find_matching_method_stub(method_name, *args)
       def find_almost_matching_stub(method_name, *args)
         method_double_for(method_name).stubs.find { |stub| stub.matches_name_but_not_args(method_name, *args) }
       end
+
+      def with_range(expectation)
+        expectation.expectation_count_type || !expectation.ordered? ? (0..-1) : @order_group.message_range
```
JonRowe: I feel this would read better as guard clauses rather than a compound ternary...

```Ruby
return expectation.expectation_count_type if expectation.expectation_count_type
return @order_group.message_range if expectation.ordered?
0..-1
```

lib/rspec/mocks/order_group.rb
```diff
@@ -66,15 +71,27 @@ def expectations_invoked_in_order?
...
-        @expectations.find { |e| message == e }
+        @expectations[order_range].find { |e| message == e }
+      end
+
+      def order_range
+        (start_range..end_range)
+      end
+
+      def start_range
+        @index - (@index.zero? ? 0 : 1)
```
JonRowe: It's not immediately clear to me why this is so

spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -459,6 +459,73 @@ module Mocks
             }.to raise_error(/received :two out of order/m)
           end
 
+          context "Specifying ordering for calls that are interleaved with each other" do
+            it 'passes when using `have_received`' do
```
myronmarston: This docstring suggests that it always passes when using `have_received`, which is not true.  (Likewise for the similar docstring for `receive` below).  When stating something like this, the doc string should state under what conditions it passes since it does not always pass.

Instead, WDYT about this structure:

``` ruby
context "Specifying ordering for calls that are interleaved with each other" do
  it "passes when the actual and expected orderings match"
  it "fails when the actual and expected orderings do not match"
end
```

Notice also that we're not mentioning `have_received` in the doc string -- this file is `have_received_spec.rb` and is entirely about `have_received` so there's no need to mention it.  The `receive` examples you added below should find another home (probably `mock_ordering_spec.rb`).

spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -459,6 +459,73 @@ module Mocks
...
+              dbl.two
+              dbl.one
+              dbl.two
+
+              expect {
+                expect(dbl).to have_received(:two).once.ordered
+                expect(dbl).to have_received(:one).twice.ordered
+                expect(dbl).to have_received(:one).once.ordered
+                expect(dbl).to have_received(:two).once.ordered
+              }.to raise_error(/received :one out of order/m)
```
myronmarston: Can this expectation be more specific about which of the ordering expectations raises the error?

(Also, this can use `fail_with`, rather than `raise_error`, I believe, which specifies the exception type as well).

spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -459,6 +459,73 @@ module Mocks
...
+                expect(dbl).to receive(:one).twice.ordered
+                expect(dbl).to receive(:one).once.ordered
+                expect(dbl).to receive(:two).once.ordered
+
+                dbl.one
+                dbl.one
+                dbl.two
+                dbl.one
+                dbl.two
+              }.to raise_error(/received :one out of order/m)
```
myronmarston: Ditto here -- would be nice for this to specify where the failure is raised, and for it to use `fail_with`.

lib/rspec/mocks/order_group.rb
```diff
@@ -66,15 +71,27 @@ def expectations_invoked_in_order?
...
+      end
+
+      def end_range
```
myronmarston: I'd normally expect any method that ends in `_range` to return a range.  Could this be `range_start` and `range_end` as those names make it more clear that these methods don't return ranges but rather the start and end of one?

johnceh:
Closing this for now since there are some open questions on issue #916 and I won't be able to get to it right away

myronmarston:
Thanks for giving this a shot, @johnceh!

PR #932, opened 2015-04-17
Improve how doubles are represented in failures

myronmarston:
Supersedes #920 and fixes rspec/rspec-expectations#688.


JonRowe:
LGTM

PR #935, opened 2015-04-22
Fix typo in receive example description.

imtayadeway:
Just remembered spotting this when researching an issue that I think is closed now.


myronmarston:
Thanks!

PR #936, opened 2015-04-26
Benchmark and optimise allocations

samphippen:
This was allocating linear hashes per mock. Now it allocates at most
one in the execution of the whole test suite, the default hash is a
constant.

WIP, but here's an initial optimisation. I'll do as many as I can and get this cleaned up by the end of next weekend.

### DIFF
--- a/benchmarks/allocations/1_object_1000_mocks.rb
+++ b/benchmarks/allocations/1_object_1000_mocks.rb
@@ -1,0 +1,76 @@
+require_relative "helper"
+symbols = (1..1000).map {|x| :"#{x}"}
+
+benchmark_allocations do
+  o = Object.new
+  symbols.each do |sym|
+    expect(o).to receive(sym)
+  end
+
+  symbols.each do |sym|
+    o.send(sym)
+  end
+
+  RSpec::Mocks.space.verify_all
+  RSpec::Mocks.space.reset_all
+end
+
+__END__
+As of commit 9ee3e3adb529113bf1cc75cc4424d014f880dc47:
+
+                     class_plus                       count
+----------------------------------------------------  -----
+Array                                                 22003
+Proc                                                   3001
+RubyVM::Env                                            3001
+Array<Symbol>                                          2000
+String                                                 2000
+Hash                                                   1002
+RSpec::Mocks::ExpectationTarget                        1000
+Enumerator                                             1000
+RSpec::Mocks::Matchers::Receive                        1000
+Array<Fixnum>                                          1000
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -28,3 +28,4 @@ platforms :jruby do
 end
 
 eval File.read('Gemfile-custom') if File.exist?('Gemfile-custom')
+gem 'allocation_stats'
```
JonRowe: The build fails because of this, prehaps it's better off in a platform directive or even just in `Gemfile-custom` as we've done similarly before for benchmarking extras (like ips stuff) before :)

benchmarks/allocations/1_object_1000_mocks.rb
```diff
@@ -0,0 +1,16 @@
...
+    expect(o).to receive(sym)
+  end
+
+  symbols.each do |sym|
+    o.send(sym)
+  end
+
+  RSpec::Mocks.space.verify_all
+  RSpec::Mocks.space.reset_all
+end
```
myronmarston: Can you add before/after output after `__END__` to this file?

lib/rspec/mocks/message_expectation.rb
```diff
@@ -176,6 +176,7 @@ def and_throw(*args)
       def and_yield(*args, &block)
         raise_already_invoked_error_if_necessary(__method__)
         yield @eval_context = Object.new if block
+        @args_to_yield ||= []
```
myronmarston: Given that it's common to refactor from this to initializing it in `initialize`, it might be worth leaving a comment explaining the reason we're not doing that here so we don't accidentally refactor back.

lib/rspec/mocks/proxy.rb
```diff
@@ -45,8 +45,10 @@ def original_method_handle_for(_message)
         nil
       end
 
+      DEFAULT_MESSAGE_EXPECTATION_OPTS = {}
```
myronmarston: Should this have a `.freeze` to ensure it's never mutated?

JonRowe: I think that would be a wise precaution...

benchmarks/allocations/1_object_1000_mocks.rb
samphippen: I've added this, check the diff :)

samphippen:
HT @mrgilman for giving me a hand with this.

myronmarston:
LGTM. Mind squashing it?

Issue #646, opened 2014-03-30
AnyInstance::Recorder should use ErrorGenerator


xaviershay:
Currently just raises its own messages, which is inconsistent with the rest of the code base.

PR #946, opened 2015-05-13
Aggregate failures

myronmarston:
This is the necessary changes for rspec/rspec-expectations#776 so that mock expectation failures are aggregated, too.

* Error raising was scattered around rspec-mocks so I spent some effort centralizing it in the error generator, which solved #646.
* With the error raising centralized it was pretty easy to use the RSpec::Support failure notifier.


myronmarston:
This is ready for review.

/cc @rspec/rspec @xaviershay

xaviershay:
LGTM

PR #955, opened 2015-05-23
Update to object formatter

### LINKED ISSUE #898: Expectations for `with`, not adding display precision for Time and DateTime
Based on [#331](https://github.com/rspec/rspec-expectations/issues/331), [#416](https://github.com/rspec/rspec-expectations/pull/416), and [f9ee8f9](https://github.com/rspec/rspec-expectations/commit/f9ee8f9cae0696b8aae97492035d31b866b636d6), I would expect that a call made to `with` comparing a `Time` or `DateTime` to output the same display precision as a non-`with` test would. Unfortunately it doesn't. :disappointed: 

Here's an example:

```ruby
context 'test' do
  class Fake; def blah(time); end; end

  it 'fails nicely' do
    expect(Time.now).to eq Time.now
  end

  it 'fails poorly' do
    expect_any_instance_of(Fake).to receive(:blah).with(Time.now)
    Fake.new().blah(Time.now)
  end
end
```

And the output:

```ruby
1) test fails nicely
   Failure/Error: expect(Time.now).to eq Time.now

     expected: 2015-02-19 00:09:28.123861000 +0000
          got: 2015-02-19 00:09:28.123845000 +0000

     (compared using ==)

     Diff:

   # ./spec/fake_spec.rb:10:in `block (5 levels) in <top (required)>'

2) test fails poorly
   Failure/Error: Fake.new().blah(Time.now)
     #<Fake:0x007fb3f8f395b8> received :blah with unexpected arguments
       expected: (2015-02-19 00:09:28 +0000)
            got: (2015-02-19 00:09:28 +0000)
   # ./spec/fake_spec.rb:20:in `block (5 levels) in <top (required)>'
```

I'd love to provide a PR for this, but can't find the seam to insert this code. If someone could point me in the right direction, that'd be awesome. :smile:

myronmarston:
Fixes #898.  Depends upon rspec/rspec-support#205.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ Gemfile-custom
 bundle
 .rbx
 .rspec-local
+spec/examples.txt
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,9 @@ Enhancements:
 * Rename `RSpec::Mocks::Configuration#when_declaring_verifying_double` to
   `RSpec::Mocks::Configuration#before_verifying_doubles` and utilise when
   verifying partial doubles. (Jon Rowe, #940)
+* Use rspec-support's `ObjectFormatter` for improved formatting of
+  arguments in failure messages so that, for example, full time
+  precisions is displayed for time objets. (Gavin Miller, Myron Marston, #955)
 
 Bug Fixes:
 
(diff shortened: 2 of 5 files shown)


JonRowe:
LGTM

PR #957, opened 2015-05-27
Prevent received count from being double-incremented.

### LINKED ISSUE #956: Negative message expectations interact weirdly with failure aggregation
``` ruby
RSpec.describe "Using `aggregate_failures` with a negative message expectation" do
  it "should report only one failure but reports two", :aggregate_failures do
    dbl = double
    expect(dbl).not_to receive(:foo)
    dbl.foo
  end
end
```

This fails with:

```
  1) Using `aggregate_failures` with a negative message expectation should report only one failure but reports two
     Got 2 failures:

     1.1) Failure/Error: dbl.foo
            (Double (anonymous)).foo(no args)
                expected: 0 times with any arguments
                received: 1 time
          # ./spec/unit/foo_spec.rb:5:in `block (2 levels) in <top (required)>'

     1.2) Failure/Error: expect(dbl).not_to receive(:foo)
            (Double (anonymous)).foo(*(any args))
                expected: 0 times with any arguments
                received: 2 times with any arguments
          # ./spec/unit/foo_spec.rb:4:in `block (2 levels) in <top (required)>'
```

Two failures are reported when it's really the same failure.  This is really confusing.

I believe the source of the problem is #884 -- with that change we notify the error twice.

I think this should be fixed before 3.3 is released.  There's also something weird going on with it claiming to have received `foo` 2 times...we should look into that as well.

myronmarston:
For #956.

This doesn't completely fix #956 (I'm still mulling over solutions) but addresses the weirdness with the failure saying it received the message 2 times when it only received it once.

### DIFF
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -526,24 +526,26 @@ module RSpec
           args.unshift(orig_object) if yield_receiver_to_implementation_block?
 
           if negative? || (allowed_to_fail && (@exactly || @at_most) && (@actual_received_count == @expected_received_count))
-            @actual_received_count += increment
             @failed_fast = true
             # args are the args we actually received, @argument_list_matcher is the
             # list of args we were expecting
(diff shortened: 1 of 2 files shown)


JonRowe:
LGTM

PR #960, opened 2015-05-30
Show the right message when using spies and calling many methods

### LINKED ISSUE #949: Confusing failure message when a spy receives multiple messages
Consider this spec:

``` ruby
RSpec.describe "Expecting that a spy received a message with the wrong args" do
  def use_spy(call_bar)
    my_spy = spy

    my_spy.foo(1)
    my_spy.bar(2) if call_bar

    expect(my_spy).to have_received(:foo).with(3)
  end

  example "when not calling another method" do
    use_spy false
  end

  example "when calling another method" do
    use_spy true
  end
end
```

The first example fails with a good error message:

```
  1) Expecting that a spy received a message with the wrong args when not calling another method
     Failure/Error: expect(my_spy).to have_received(:foo).with(3)
       #<Double (anonymous)> received :foo with unexpected arguments
         expected: (3)
              got: (1)
```

The second fails with a really inaccurate and confusing error message:

```
  1) Expecting that a spy received a message with the wrong args when calling another method
     Failure/Error: expect(my_spy).to have_received(:foo).with(3)
       (Double (anonymous)).foo(3)
           expected: 1 time with arguments: (3)
           received: 0 times
```

It says `foo` was not received but it was.  Somehow the act of calling `bar` on the spy causes the confusing failure message.  I have no idea why.  We should fix it.

mauricio:
This fixes the issue where when you called a method expected using `have_received` with an unexpected argument and called some other bogus method the spec error message wouldn't say the expected method was called with wrong arguments.

Fixes #949.

### DIFF
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -101,12 +101,15 @@ module RSpec
       def check_for_unexpected_arguments(expectation)
         return if @messages_received.empty?
 
-        # Check if any of the arguments were expected.
-        return if @messages_received.any? do |(method_name, args, _)|
-                    !expectation.matches_name_but_not_args(method_name, *args)
-                  end
+        return if @messages_received.any? { |method_name, args, _| expectation.matches?(method_name, *args) }
 
-        raise_unexpected_message_args_error(expectation, messages_arg_list)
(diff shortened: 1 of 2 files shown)


myronmarston:
:+1: Great work, @mauricio!

PR #962, opened 2015-06-02
Pass `source_id` when notifying failures

### LINKED ISSUE #956: Negative message expectations interact weirdly with failure aggregation

myronmarston:
Fixes #956 (in combination with rspec/rspec-expectations#799, which should be merged first).

### DIFF
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -477,7 +477,7 @@ module RSpec
             @error_generator.raise_expectation_error(
               @message, @expected_received_count, @argument_list_matcher,
               @actual_received_count, expectation_count_type, expected_args,
-              @expected_from
+              @expected_from, exception_source_id
             )
           else
             @error_generator.raise_similar_message_args_error(
@@ -487,7 +487,7 @@ module RSpec
         end
 
         def raise_unexpected_message_args_error(args_for_multiple_calls)
-          @error_generator.raise_unexpected_message_args_error(self, args_for_multiple_calls)
+          @error_generator.raise_unexpected_message_args_error(self, args_for_multiple_calls, exception_source_id)
         end
 
         def expectation_count_type
@@ -530,6 +530,10 @@ module RSpec
 
       private
 
+        def exception_source_id
+          @exception_source_id ||= "#{self.class.name} #{__id__}"
+        end
+
         def invoke_incrementing_actual_calls_by(increment, allowed_to_fail, parent_stub, *args, &block)
           args.unshift(orig_object) if yield_receiver_to_implementation_block?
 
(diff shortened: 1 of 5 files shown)


myronmarston:
This one is green and ready for review, although it can't be merged until the rspec-expectations PR is merged and the temp commit here removed. /cc @rspec/rspec

lib/rspec/mocks/message_expectation.rb
```diff
@@ -378,7 +378,6 @@ def initialize(error_generator, expectation_ordering, expected_from, method_doub
           # Initialized to nil so that we don't allocate an array for every
           # mock or stub. See also comment in `and_yield`.
           @args_to_yield = nil
-          @failed_fast = nil
```
JonRowe: Why did this go away?

lib/rspec/mocks/message_expectation.rb
```diff
@@ -563,10 +569,6 @@ def raise_already_invoked_error_if_necessary(calling_customization)
           error_generator.raise_already_invoked_error(message, calling_customization)
         end
 
-        def failed_fast?
-          @failed_fast
-        end
```
JonRowe: As above really, why has this gone away?

myronmarston: I realized that it is no longer used.  It used to be used but the usage was removed in #884.  `failed_fast?` and `@failed_fast` were vestiges of old logic that we no longer have a reason to keep around.

JonRowe: Fair enough, merge when temp commit removed and green ;)

myronmarston:
Pushed one more commit to remove the last use of the deprecated notifier interface, which should allow rspec/rspec-support#212 to pass once this is merged.

myronmarston:
Merging even though appveyor isn't done yet.  (I'm too impatient to wait).

Issue #964, opened 2015-06-04
expect and_call_original messing up with modules build as jruby extensions.


purbon:
While doing some test to upgrade https://github.com/elastic/logstash to use the last rspec set of gems, I found out an issue with the way expectation with the and_call_orignal call work in the 3.x branch. 

*Situation*

In logstash we're using the jruby wrapper of jrjackson [1] as a json lib, so we get a really fast json processing. So to build the test we use calls like:

```
 expect(JrJackson::Raw).to receive(:parse_raw).with(json_hash).and_call_original
 expect(LogStash::Json.load(json_hash)).to eql(hash)
```
during the upgrade to use rspec 3.x I discovered that this kind of expectation were not playing well anymore, as they did with versions < 2.99. There seems to be a regression with using the ```and_call_original``` in modules defined as jruby/java exceptions.

To make a bit more easy to reproduce this I made a simple example that you can find at https://github.com/purbon/playground/tree/master/call_origin_with_module, running the test in this dummy gem will:

* If using the last rspec 3.2.x:

```
CallOriginWithModule
  calls the generate once
  calls the generate twice (FAILED - 1)
  calls the dump once (FAILED - 2)
  calls the dump twice (FAILED - 3)
  without using call original
    calls the generate once
    calls the generate twice

Failures:

  1) CallOriginWithModule calls the generate twice
     Failure/Error: expect(AClass.bar(3)).to eq("{\"source\":3,\"target\":3}")
     NoMethodError:
       undefined method `generate' for JrJackson::Raw:Module
     # ./lib/call_origin_with_module.rb:8:in `bar'
     # ./spec/call_origin_with_module_spec.rb:12:in `(root)'
```

if using a version <= 2.99:

```
╰─$ bundle exec rspec

CallOriginWithModule
  calls the generate once
  calls the generate twice
  calls the dump once
  calls the dump twice
  without using call original
    calls the generate once
    calls the generate twice

Finished in 0.322 seconds
6 examples, 0 failures
```

[1] https://github.com/guyboertje/jrjackson

JonRowe:
Can you try something for me? Can you run your build with: 

```
 export JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
```

purbon:
Hi @JonRowe I tried to run https://github.com/purbon/playground/tree/master/call_origin_with_module with your proposed options and I see the same error:

```
[...]

PR #965, opened 2015-06-05
Spike checking #964

JonRowe:
WIP


myronmarston:
@JonRowe -- remind me what this is for?

JonRowe:
It was an attempt to trigger a bug from another issue, so I could debug it (without having to worry about getting JRuby running locally)

PR #969, opened 2015-06-09
Silence whitespace-only diffs.

myronmarston:


JonRowe:
Merging because as mentioned on #956 this is failing due to an unrelated issue.

PR #970, opened 2015-06-10
Deprecate using `with` without arguments.

yousuketto:
RSpec3.0 prohibit using `with` without arguments.
- https://github.com/rspec/rspec-mocks/blob/v3.0.4/lib/rspec/mocks/message_expectation.rb#L336-L339

### example spec
```ruby
class Dummy
end

RSpec.describe Dummy do
  let(:dummy){Dummy.new}
  it "use 'with()'" do
    expect(dummy).to receive(:dummy_method).with().and_return("this is dummy method.")

    expect(dummy.dummy_method).to eq("this is dummy method.")
  end
end
```

### execute by rspec2.99
```
$ bundle exec rspec example_spec.rb 
.

Finished in 0.00211 seconds
1 example, 0 failures
```

### execute by rspec3.0
```
$ bundle exec rspec example_spec.rb 
F

Failures:

  1) Dummy use 'with()'
     Failure/Error: expect(dummy).to receive(:dummy_method).with().and_return("this is dummy method.")
     ArgumentError:
       `with` must have at least one argument. Use `no_args` matcher to set the expectation of receiving no arguments.
     # ./example_spec.rb:7:in `block (2 levels) in <top (required)>'

Finished in 0.02599 seconds (files took 0.21393 seconds to load)
1 example, 1 failure
```

### DIFF
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -358,6 +358,8 @@ module RSpec
           else
             self.inner_implementation_action = block
           end
+        elsif args.empty?
+          RSpec.deprecate "Using `with` without arguments", :replacement => "`with(no_args)`"
         end
 
         @argument_list_matcher = ArgumentListMatcher.new(*args, &block)
--- a/spec/rspec/mocks/mock_spec.rb
+++ b/spec/rspec/mocks/mock_spec.rb
@@ -776,6 +776,22 @@ module RSpec
             @double.should_receive(:foo).with('bar')
             @double.foo('bar')
           end
+
+          it "warns of deprecation when called without arguments" do
+            expect(RSpec).to receive(:deprecate) do |message, opts|
+              expect(message).to match(/Using `with` without arguments/)
+            end
+            obj = Object.new
+            obj.stub(:foo).with().and_return('bar')
+            expect(obj.foo).to eq('bar')
+          end
+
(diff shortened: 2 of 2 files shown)


JonRowe:
Seems reasonable but for some reason our build is failing here, any chance you feel like investigating?

myronmarston:
It's because aruba had a breaking change between 0.6.1 and 0.6.2.  See rspec/rspec-core@294fc7fa59195a61269c034e2eb682921ab70994 and rspec/rspec-core#1850.  We had to make changes to each of the repos at that point but didn't backport them to 2.99.

We'll have to make a change to each repo to get the builds back to green.  Simplest is to tighten up the aruba version constraint to exactly 0.6.1 to avoid this issue.  I'll go commits changes to the 2-99-maintenance branch of each repo so the build here can pass.

myronmarston:
@yousuketto - can you add a spec for the deprecation?  See 7349928c5a94cb594bea30b1c2d7dd6984bf9db8 for an example of a similar deprecation and spec.

myronmarston:
OK, I pushed commits to the repos to pin aruba to 0.6.1, before the `regexp` method was removed.

If you rebase against `2-99-maintenance`, I believe your PR will pass now.

yousuketto:
I add a spec for the deprecation. Is it fine like that?

JonRowe:
Thanks @yousuketto

Issue #972, opened 2015-06-12
allow no longer works with ActiveRecord classes


tisba:
While updating a rails 4.2.1 project from rspec 3.2.0 to 3.3.0 I got a lot of failing specs.

I noticed that the issue can be reproduced simply by the following (pointless spec):

```ruby
it "will fail" do
  allow(User).to receive(:confirmed).and_return(:something)
end
```

myronmarston:
Thanks for reporting this.

Can you provide a backtrace for the error you are getting?  We haven't seen that problem and can't guess what the problem is.  Also, if you're able to provide a reproducible example, that would help us get to the bottom of this quickly.

Thanks!

baburdick:
+1

```ruby
    context "error occurred" do
      before do
        allow(Account).to receive(:new).and_raise("Oops")
      end

      it "returns 'error'" do
        post :create, params

        expect(response                ).to be_ok
        expect(json_response["status"] ).to eq "error"
        expect(json_response["message"]).to eq "Oops"
      end
    end
```
```
Failures:

  1) AccountsController#create error occurred returns 'error'
     Failure/Error: allow(Account).to receive(:new).and_raise("Oops")
     ActiveRecord::ActiveRecordError:
       ActiveRecord::Base doesn't belong in a hierarchy descending from ActiveRecord
     # /Users/bruceb/.rvm/gems/ruby-2.1.2@account_service/gems/activerecord-4.1.10/lib/active_record/inheritance.rb:71:in `base_class'
     # /Users/bruceb/.rvm/gems/ruby-2.1.2@account_service/gems/activerecord-4.1.10/lib/active_record/attribute_methods.rb:78:in `block in define_attribute_methods'
     # /Users/bruceb/.rvm/gems/ruby-2.1.2@account_service/gems/activerecord-4.1.10/lib/active_record/attribute_methods.rb:76:in `define_attribute_methods'
     # ./spec/controllers/accounts_controller_spec.rb:3:in `block (4 levels) in <top (required)>'
```

myronmarston:
Thanks, @baburdick, that's helpful.  Can you re-run it with `-b` to enable the full backtrace so we can see the RSpec stack frames?

baburdick:
```
Failures:

  1) CustomerAccountsController#create error occurred returns 'error'
     Failure/Error: allow(AccountCore::CustomerAccount).to receive(:new).and_raise("Oops")
     ActiveRecord::ActiveRecordError:
[...]

PR #973, opened 2015-06-13
Add failing spec for #972.

myronmarston:
Not sure how to fix it yet.  Figured I'd open this, though.

### DIFF
--- a/spec/rspec/mocks/partial_double_spec.rb
+++ b/spec/rspec/mocks/partial_double_spec.rb
@@ -383,6 +383,22 @@ module RSpec
         object.implemented
       end
 
+      context "for a class" do
+        it "only runs the `before_verifying_doubles` callback for the class (not for superclasses)" do
+          subclass = Class.new(klass)
+
+          expect { |probe|
+            RSpec.configuration.mock_with(:rspec) do |config|
+              config.before_verifying_doubles(&probe)
+            end
+
+            allow(subclass).to receive(:new)
+          }.to yield_successive_args(
+            an_object_having_attributes(:target => subclass),
+          )
+        end
+      end
+
       it 'does not allow a non-existing method to be expected' do
         prevents { expect(object).to receive(:unimplemented) }
       end


cupakromer:
Note, while the spec added is attempting to stub `new`, which comes from `BasicObject` in Ruby 2, the issue happens for any class method, even a singleton method defined directly on the object and not in the ancestor chain.

I think I narrowed this all down to here: [`PartialClassDoubleProxyMethods#original_method_handle_for`](https://github.com/rspec/rspec-mocks/blob/c3716a07e4/lib/rspec/mocks/proxy.rb#L348-L374)

This method seems to _always_ build the entire proxy chain. It is the building of this proxy chain which causes [
`RSpec::Mocks.configuration.verifying_double_callbacks.each`](https://github.com/rspec/rspec-mocks/commit/e5bfbcb75b#diff-0bf4a04898f16be2b40df1fd25fb2ac6R102) to provide all ancestors of `ActiceRecord`, including `ActiveRecord::Base`, to the block.

In regards to the rspec-rails issue, I'm not sure there is anything rspec-mocks can, or should do, which would have prevented it from occurring. It would always have been possible for someone to attempt to stub `ActiveRecord::Base` causing the issue.

I do not know enough about this process to understand if building the entire proxy chain is a :bug:, preferable, or simply "The Right Thing"™.

All I can guess at is, we could attempt to short circuit the proxy chain creation as soon as we found the first un-stubbed method definition in the ancestor chain. However, that will not prevent the entire chain being generated for the cause of stubbing `new`.

/cc @myronmarston @samphippen

samphippen:
I'm gonna have a crack at this tomorrow. Thanks for your notes @cupakromer

myronmarston:
> I think I narrowed this all down to here: `PartialClassDoubleProxyMethods#original_method_handle_for`

> This method seems to always build the entire proxy chain.

Yep.  You've correctly identified the source of the issue.

> It would always have been possible for someone to attempt to stub `ActiveRecord::Base` causing the issue.

I came to the same conclusion, which is why I forged ahead with the rspec-rails fix.  Users generally don't stub `ActiveRecord::Base` directly, but if they did, the error would have occurred even if rspec-mocks didn't invoke the callback for every ancestor.

> I do not know enough about this process to understand if building the entire proxy chain is a :bug:, preferable, or simply "The Right Thing"™.

I'd say that building the entire proxy chain is an implementation detail that specifically allows stubbing a class and its superclass (or a subclass) to work.  That said, I think that invoking the verifying double callbacks with each ancestor is a bug.  It's purely accidental due to internal implementation details of the proxies, and not something a user of the callback API would ever expect.  (Consider, for example, that when you do `instance_double(SomeModelClass)`, we do not call the callback with each ancestor of `SomeModelClass`).

> All I can guess at is, we could attempt to short circuit the proxy chain creation as soon as we found the first un-stubbed method definition in the ancestor chain. However, that will not prevent the entire chain being generated for the case of stubbing new.

IMO, having the callback invoked for _some_ of the ancestors but not all, would be the worst result, because then it's unpredictable from the user point of view which ancestors the callback will be invoked with.

Here's what I'm thinking we want to do (although I haven't figured out the details of how to do it, and I'm glad @samphippen has volunteered to step up to the plate!): refactor the internals to no-longer instantiate proxies for each ancestor while maintaining the same external behavior enabled by the logic @cupakromer linked to.

A good starting point would be to comment out all of the methods of `RSpec::Mocks::PartialClassDoubleProxyMethods` except `initialize`, then see what fails, then try to come up with an alternate implementation that makes them pass.  Maybe that logic could be pulled into some kind of a helper module that would allow you to recurse up the ancestor chain w/o instantiating proxies for each?  Apart from fixing this issue, avoiding instantiating unnecessary/unused proxies sounds like a win for fewer object allocations, anyway :).

PR #974, opened 2015-06-14
A fix for #972

samphippen:
It seemed to me that if the problem is creating superclass proxies, then we could just add a method and put a flag in to the constructors to fix this. The current fix is a bit shonky, but demonstrates the solution.

I have some questions/thoughts:

1. Perhaps, this behaviour change should be extracted out in to an object, as opposed to a conditional. We'd inject a callbacks collaborator into the proxies which they would then invoke
2. Perhaps this should be flag on proxy_for instead of adding a new method?
3. I think the verifying class proxy is the only place we need to put the flag, but is that correct?

This fix is *really simple*, but I'm not sure it's architecturally brilliant. Thoughts?

### DIFF
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -391,7 +391,7 @@ module RSpec
         return @superclass_proxy if defined?(@superclass_proxy)
 
         if (superclass = object.superclass)
-          @superclass_proxy = @source_space.proxy_for(superclass)
+          @superclass_proxy = @source_space.superclass_proxy_for(superclass)
         else
           @superclass_proxy = nil
         end
--- a/lib/rspec/mocks/space.rb
+++ b/lib/rspec/mocks/space.rb
@@ -35,6 +35,10 @@ module RSpec
         false
       end
 
+      def superclass_proxy_for(*_args)
+        raise_lifecycle_message
+      end
+
       def new_scope
         Space.new
       end
@@ -109,6 +113,13 @@ module RSpec
         end
       end
 
+      def superclass_proxy_for(klass)
+        proxy_mutex.synchronize do
+          id = id_for(klass)
+          proxies.fetch(id) { superclass_proxy_not_found_for(id, klass) }
+        end
+      end
+
       alias ensure_registered proxy_for
 
       def registered?(object)
@@ -150,11 +161,7 @@ module RSpec
                       when NilClass   then ProxyForNil.new(@expectation_ordering)
                       when TestDouble then object.__build_mock_proxy_unless_expired(@expectation_ordering)
                       when Class
-                        if RSpec::Mocks.configuration.verify_partial_doubles?
-                          VerifyingPartialClassDoubleProxy.new(self, object, @expectation_ordering)
-                        else
-                          PartialClassDoubleProxy.new(self, object, @expectation_ordering)
-                        end
+                        class_proxy_with_callback_verification_strategy(object, CallbackInvocationStrategy.new)
                       else
                         if RSpec::Mocks.configuration.verify_partial_doubles?
                           VerifyingPartialDoubleProxy.new(object, @expectation_ordering)
@@ -162,6 +169,24 @@ module RSpec
                           PartialDoubleProxy.new(object, @expectation_ordering)
                         end
                       end
+      end
+
+      def superclass_proxy_not_found_for(id, object)
+        raise "superclass_proxy_not_found_for called with something that is not a class" unless Class === object
+        proxies[id] = class_proxy_with_callback_verification_strategy(object, NoCallbackInvocationStrategy.new)
+      end
+
+      def class_proxy_with_callback_verification_strategy(object, strategy)
+        if RSpec::Mocks.configuration.verify_partial_doubles?
+          VerifyingPartialClassDoubleProxy.new(
+            self,
+            object,
+            @expectation_ordering,
+            strategy
+          )
+        else
+          PartialClassDoubleProxy.new(self, object, @expectation_ordering)
+        end
       end
 
       def any_instance_recorder_not_found_for(id, klass)
--- a/lib/rspec/mocks/verifying_proxy.rb
+++ b/lib/rspec/mocks/verifying_proxy.rb
@@ -4,6 +4,21 @@ RSpec::Support.require_rspec_mocks 'method_reference'
 module RSpec
   module Mocks
     # @private
+    class CallbackInvocationStrategy
+      def call(doubled_module)
+        RSpec::Mocks.configuration.verifying_double_callbacks.each do |block|
+          block.call doubled_module
+        end
+      end
+    end
+
+    # @private
+    class NoCallbackInvocationStrategy
+      def call(_doubled_module)
+      end
+    end
+
+    # @private
     module VerifyingProxyMethods
       def add_stub(method_name, opts={}, &implementation)
         ensure_implemented(method_name)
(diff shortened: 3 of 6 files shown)


lib/rspec/mocks/proxy.rb
```diff
@@ -376,22 +376,22 @@ def original_method_handle_for(message)
...
+        # looking up the ancestor chain.
+        if @method_doubles.include?(message)
+          @method_doubles.fetch(message).original_method.unbind
+        else
+          superclass_proxy\
+            && superclass_proxy.original_unbound_method_handle_from_ancestor_for(message)
         end
-
-        method_double.original_method.unbind
       end
```
myronmarston: Why the change to this method?  It looks to me like what you have is semantically equivalent but slightly less efficient.  After all, for the case where `if @method_doubles.include?(message); @method_doubles.fetch(message)` is going to be less efficient than just `@method_doubles.fetch(message)` like we had before.

myronmarston: Also, your version is potentially subject to race conditions--e.g. if `@method_doubles` changes between the `include?` check and the `fetch`.

samphippen: it's a left over from when I was doing some debugging, I'll undo it.

samphippen:
@myronmarston I'm happy with the state of this PR now. Any other comments, or shall I squash and merge on green?

myronmarston:
Haven't had a chancery do a full review yet. I'll take a look later today.

lib/rspec/mocks/space.rb
```diff
@@ -164,6 +171,24 @@ def proxy_not_found_for(id, object)
...
+          VerifyingPartialClassDoubleProxy.new(
+            self,
+            object,
+            @expectation_ordering,
+            strategy
+          )
+        else
+          PartialClassDoubleProxy.new(self, object, @expectation_ordering)
+        end
+      end
```
cupakromer: Could we remove this check? Instead we set the proxy adapter on the configuration when `verify_partial_doubles` is configured. Something like:

```ruby
def verify_partial_doubles=(flag)
  # set the state
  if flag
    self.partial_class_double_proxy = VerifyingPartialClassDoubleProxy
  else
    self.partial_class_double_proxy = PartialClassDoubleProxy
  end
end
```
Turning section into:

```ruby
def class_proxy_with_callback_verification_strategy(object, strategy)
  RSpec::Mocks.configuration.partial_class_double_proxy.new(
    self,
    object,
    @expectation_ordering,
    strategy
  )
end
```

Or even making that a factory method on the configuration so we do not have to call `new`.

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -383,6 +383,22 @@ def defined_private_method
...
+            RSpec.configuration.mock_with(:rspec) do |config|
+              config.before_verifying_doubles(&probe)
+            end
+
+            allow(subclass).to receive(:new)
+          }.to yield_successive_args(
+            an_object_having_attributes(:target => subclass)
+          )
+        end
+      end
```
cupakromer: Since we have not grouped the other specs into states of object instances, and given we only have the one spec, could we flatten this (e.g. remove the `context`)?

cupakromer:
LGTM :smiley_cat: I like the refactored approach. Left some minor comments, neither are merge blockers.

lib/rspec/mocks/space.rb
samphippen: I think the setter method also setting up the state is a bit weird, but I think we could have a method:

```
def partial_class_double_proxy_class
  {
    true => VerifyingPartialClassDoubleProxy,
    false => PartialClassdoubleProxy
  }.fetch(RSpec::Mocks.configuration.verify_partial_doubles?)
end
```
which would do the trick. Thoughts?

cupakromer: I was hoping to avoid needing to do the lookup / check on every call. Instead we just send the message to an adapter. However, my main goal, was simplifying this part of the `Space` code.

How the configuration manages the adapter I view as a "private implementation detail". I don't have too strong an opinion on how that happens.

cupakromer: > I think the setter method also setting up the state is a bit weird

Can you expand on this? I'm not sure I follow. Isn't that what a "setter" is intended to do?

lib/rspec/mocks/verifying_proxy.rb
```diff
@@ -4,6 +4,21 @@
...
+          block.call doubled_module
+        end
+      end
+    end
+
+    # @private
+    class NoCallbackInvocationStrategy
+      def call(_doubled_module)
+      end
+    end
```
myronmarston: Personally, using the strategy pattern to conditionally do something feels like overkill.  The strategy pattern has its uses but here it feels to me like it's being used just to avoid an `if` conditional, and IMO the `if` conditional would actually be simpler and better.  Anytime you're adding a conditional I think it's worth considering what a polymorphic solution would look like...but not always worth choosing it over the conditional.

Anyhow, if you do decide to keep using the strategy pattern here, I'd advocate for making these singleton modules instead of classes you instantiate since they are stateless.  Fewer object instantiates means less garbage for the GC to cleanup :).

myronmarston:
Thanks for coming up with a solution quickly, @samphippen!

I'm a bit concerned that this increases the complexity of proxy instantiation considerably.  As I mentioned in #972, I don't think it's desirable for us to instantiate the superclass proxies at all.  It's merely an implementation detail of getting `original_method_handle_for` to look up the superclass chain--and now it's an implementation detail that is growing even more complex in this PR.  The fact that the instantiating superclass proxies caused a bug is, IMO, a hint that maybe we should look into an alternate implementation that avoids instantiating superclass proxies entirely.  I expect such a solution would _decrease_ complexity rather than increase it.

That said, this PR definitely fixes a bug that I'd like to get squashed soon, and merging this doesn't preclude us from doing the alternate solution in the future.  It may make such a future refactoring more difficult, though, and increase maintenance costs (as complexity tends to attract more complexity).  That's the main thing that's making me reticent to merge this as-is.

Thoughts?

samphippen:
@myronmarston given this represents a real bug fix, I'd like to see it merged. I agree that aiming at reducing complexity is a good idea, but I don't think it should come at the cost of fixes. This also has behaviour that should be preserved after any refactorings.

Regarding the strategy pattern situation: I'd rather not introduce a conditional if it can be avoided, I find conditionals breed like complexity. I'm happy to store individual instances of both strategy objects so that we don't have to make new ones each time.

myronmarston:
Let's go with this fix.

myronmarston:
There's another issue we've discovered with the way we invoke the callbacks for this case:

https://github.com/rspec/rspec-core/issues/1994#issuecomment-112325603

This can still be merged but there may be more changes we want in this area.

/cc @samphippen

myronmarston:
@samphippen -- any reason for the delay on squashing and merging this?  Would be good to get a patch release out with the fix.

samphippen:
Done, sorry. @myronmarston what's a good changelog entry for this while we're waiting for CI?

samphippen:
Trying to debug that jruby failure.

samphippen:
Cannot reproduce that failure locally.

samphippen:
@myronmarston any thoughts on that one CI failure?

myronmarston:
I can reproduce it.  The source of the failure is rspec/rspec-support#215.  Have you pulled the latest rspec-support?  I'm guessing your local checkout is out of date.

The source of the problem is that JRuby 1.8 mode does `pp` kinda weird, apparently, so the output you generate in the test here:

https://github.com/rspec/rspec-mocks/blob/f7837f522072a890b7be7ad8f71b66eca3f18de9/spec/rspec/mocks/diffing_spec.rb#L148

...differs from the output the differ generates via this line of code:

https://github.com/rspec/rspec-support/blob/40e434d4b77065ea3f3fd98bcde10c4ed9a6f30c/lib/rspec/support/object_formatter.rb#L103

I pushed a fix that makes `inspect` more determinstic.

samphippen:
@myronmarston is this good to merge?

myronmarston:
> @myronmarston is this good to merge?

Yes, after we add a changelog entry (forgot to respond to your question about that yesterday...sorry!).  Maybe this?

> Fix bug in `before_verifying_double` callback logic that caused it to be called once for class
> in the ancestor list when mocking or stubbing a class. Now it is only called for the mocked or
> stubbed class, as you would expect. (Sam Phippen, #974)

myronmarston:
You can add that post-merge, though.

Also, please backport to 3-3-maintenance.

PR #976, opened 2015-06-19
Backport #974 to 3.3

samphippen:


Changelog.md
```diff
@@ -1,3 +1,12 @@
+### 3.3.1 Development
+
+Bug Fixes:
+
+* Fix bug in before_verifying_double callback logic that caused it to be called
+  once for class in the ancestor list when mocking or stubbing a class. Now it
```
myronmarston: `before_verifying_double` should have backticks around it since it's a method name.

Also, "one for each class" instead of "once for class".

(Sorry if my original suggestion included these typos!).

BTW, don't worry about updating this in `master` -- I always update the changelog ever each patch release so that the `master` changelog includes the release info.

myronmarston:
LGTM.

PR #978, opened 2015-06-20
Make `have_received` work without relying upon rspec-expectations.

### LINKED ISSUE #967: have_received matcher fails with Minitest

myronmarston:
Fixes #967 and #968.

@randycoulman -- want to try this out?

### DIFF
--- a/spec/rspec/mocks/matchers/have_received_spec.rb
+++ b/spec/rspec/mocks/matchers/have_received_spec.rb
@@ -1,6 +1,26 @@
 module RSpec
   module Mocks
-    RSpec.describe Matchers::HaveReceived do
+    # This shared example group is highly unusual as it is used to test how
+    # `have_received` works in two situations:
+    #
+    # * With rspec-mocks as a standalone library.
+    # * Together with rspec-expectations.
+    #
+    # To simulate the former, we've had to hack things a bit. Special care must be taken:
+    #
+    # * Only define examples with `it`, (not `fit`, `xit`, `specify`, etc). We redefine
+    #   `it` below to make it support our needs here but that definition isn't applied to
+    #   the other forms.
+    # * All normal expectations must use `_expect`, not `expect`. Expectations
+    #   for `have_received` should use `expect`.
+    RSpec.shared_examples_for Matchers::HaveReceived do
+      # Make rspec-expectations' `expect` available via an alias so that when
+      # this group is included below in a context that simulates rspec-expectations
+      # not being loaded by using rspec-mocks' `expect` instead of rspec-expectations'
+      # `expect`, we still have a way to use the expectations one for normal expectations.
+      # In this group, all normal expectations should use `_expect` instead of `expect`.
+      alias _expect expect
+
       describe "expect(...).to have_received" do
         it 'passes when the double has received the given message' do
           dbl = double_with_met_expectation(:expected_method)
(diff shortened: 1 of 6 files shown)


spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -1,6 +1,26 @@
 module RSpec
   module Mocks
-    RSpec.describe Matchers::HaveReceived do
+    # This shared exampe group is highly unusual as it is used to test how
```
randycoulman: `exampe` => `example`

randycoulman:
I've tested this with my example case, and it seems to work just fine.  Note the one typo I pointed out.

Making rspec-mocks fully work without rspec-expectations is a much better solution than just making it clear that parts of it don't work, so I much prefer this solution to my workaround.

I missed the `outside_rspec` features when I did my PR, so I'm glad that's there.  The way you wrote the RSpec specs for this is impressive and a little bit scary, but I'm glad you figured out a way to do it.  The rest of the code seems relatively straightforward.

spec/rspec/mocks/matchers/have_received_spec.rb
myronmarston: Thanks, fixed.

myronmarston:
> The way you wrote the RSpec specs for this is impressive and a little bit scary, but I'm glad you figured out a way to do it.

Yep.  Felt like a hack, and it's probably not the way I would do it if I was writing all the tests from scratch...but it was the easiest way I could find to re-use the existing specs.

Thanks for trying this out :).

spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -1,6 +1,26 @@
...
+    #   the other forms.
+    # * All normal expectations must use `_expect`, not `expect`. Expectations
+    #   for `have_received` should use `expect`.
+    RSpec.shared_examples_for Matchers::HaveReceived do
+      # Make rspec-expectations' `expect` available via an alias so that when
+      # this group is included below in a context that simulates rspec-expectations
+      # not being loaded by using rspec-mocks' `expect` instead of rspec-expectations'
+      # `expect`, we still have a way to use the expectations one for normal expectations.
+      # In this group, all normal expectations should use `_expect` instead of `expect`.
+      alias _expect expect
```
samphippen: Would making this name longer (e.g. `rspec_expectations_expect`) or changing the other one to be "mocks_expect" add clarity here?

samphippen:
One minor quibble but I'm otherwise happy :+1:

spec/rspec/mocks/matchers/have_received_spec.rb
myronmarston: > Would making this name longer (e.g. `rspec_expectations_expect`)

I think I'd want a longer name if it wasn't localized to this one file.  `rspec_expectations_expect` is soo much longer that I prefer the terser `_expect`.

> or changing the other one to be "mocks_expect" add clarity here?

I thought about that, but there's a downside to that approach: if we wrote a new spec, and forgot to use `mocks_expect` for `have_received`, the spec would pass even though it wasn't testing what we intend it to.  Contrast that to when we write a new spec and forget to use `_expect` over `expect`: we'll get a failure and we'll be forced to fix it.

I much prefer things setup so that when we forget it fails rather than getting a false positive.

PR #979, opened 2015-06-24
Switch to using reentrant mutex to avoid deadlocks

JonRowe:
Part of a solution for rspec/rspec-core#1994

### DIFF
--- a/lib/rspec/mocks/space.rb
+++ b/lib/rspec/mocks/space.rb
@@ -1,3 +1,5 @@
+RSpec::Support.require_rspec_support 'reentrant_mutex'
+
 module RSpec
   module Mocks
     # @private
@@ -142,18 +144,8 @@ module RSpec
 
     private
 
-      # We don't want to depend on the stdlib ourselves, but if the user is
-      # using threads then a Mutex will be available to us. If not, we don't
-      # need to synchronize anyway.
       def new_mutex
-        defined?(::Mutex) ? ::Mutex.new : FakeMutex
-      end
-
-      # @private
-      module FakeMutex
-        def self.synchronize
-          yield
-        end
+        Support::ReentrantMutex.new
       end
 
       def proxy_not_found_for(id, object)
--- a/spec/rspec/mocks/partial_double_spec.rb
+++ b/spec/rspec/mocks/partial_double_spec.rb
@@ -383,6 +383,22 @@ module RSpec
         object.implemented
       end
 
+      it 'avoids deadlocks when a proxy is accessed from within a `before_verifying_doubles` callback' do
+        _klass = Class.new { def message; end; }
+        called_for = []
+
+        RSpec.configuration.mock_with(:rspec) do |config|
+          config.before_verifying_doubles do |ref|
+            unless called_for.include? ref.target
+              called_for << ref.target
+              ::RSpec::Mocks.space.proxy_for(ref.target)
+            end
+          end
+        end
+
+        expect { allow(_klass.new).to receive(:message) }.to_not raise_error
+      end
+
       context "for a class" do
         it "only runs the `before_verifying_doubles` callback for the class (not for superclasses)" do
           subclass = Class.new(klass)


JonRowe:
^^ @myronmarston

myronmarston:
You mind adding a spec for this?  Basically, if you write a spec that registers a `before_verified_double` callback that does something that causes a proxy to be accessed (e.g. stubbing a method) it should fail w/o this with the deadlock error and pass with it.

lib/rspec/mocks/space.rb
```diff
@@ -1,3 +1,4 @@
+require 'rspec/support/reentrant_mutex'
```
myronmarston: Should be `RSpec::Support.require_rspec_support "reentrant_mutex"`

(And a blank line between this and `module RSpec` would be nice).

lib/rspec/mocks/space.rb
```diff
@@ -143,10 +144,11 @@ def any_instance_recorders_from_ancestry_of(object)
...
 
       # We don't want to depend on the stdlib ourselves, but if the user is
-      # using threads then a Mutex will be available to us. If not, we don't
-      # need to synchronize anyway.
+      # using threads then a Mutex will be available to us (which we use via
+      # (RSpec::Support::ReentrantMutex). If not, we don't need to synchronize
+      # anyway.
       def new_mutex
-        defined?(::Mutex) ? ::Mutex.new : FakeMutex
+        defined?(::Mutex) ? Support::ReentrantMutex.new : FakeMutex
```
myronmarston: I don't think the `define?(::Mutex)` bit or `FakeMutex` makes sense here anymore.

JonRowe: I think it does, as it still prevents unnecessarily loading `mutex`

lib/rspec/mocks/space.rb
JonRowe: I actually moved this below anyway

lib/rspec/mocks/space.rb
```diff
@@ -143,10 +143,16 @@ def any_instance_recorders_from_ancestry_of(object)
...
 
       # We don't want to depend on the stdlib ourselves, but if the user is
-      # using threads then a Mutex will be available to us. If not, we don't
-      # need to synchronize anyway.
+      # using threads then a Mutex will be available to us (which we use via
+      # (RSpec::Support::ReentrantMutex). If not, we don't need to synchronize
+      # anyway.
       def new_mutex
-        defined?(::Mutex) ? ::Mutex.new : FakeMutex
+        if defined?(::Mutex)
```
myronmarston: I still think this should be removed because on all rubies besides 1.8.7, `Mutex` is _always_ defined.  On 1.8.7 it's in the stdlib (made available by `require 'thread'`) but on 1.9+ it's a built-in that's always available:

```
➜  ~  ruby --disable=gems -e 'puts Mutex.new'
#<Mutex:0x007fe5940a9048>
```

I (wrongly) thought it would only be available if a user had required `thread` but learned that it's always there as part of rspec/rspec-core#1858.  So now this check is just adding unnecessary overhead and should be removed, IMO.

JonRowe:
I'm not quite sure why 2.2 doesn't like this, it doesn't fail for me locally...

myronmarston:
> I'm not quite sure why 2.2 doesn't like this, it doesn't fail for me locally...

It fails only on ruby 2.2.0.  On later versions of 2.2.x it does not fail.  There was a new warning in ruby 2.2.0 that I [complained about](https://bugs.ruby-lang.org/issues/10661) and which was removed in later 2.2.x releases.

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -383,6 +383,22 @@ def defined_private_method
...
+        called_for = []
+        RSpec.configuration.mock_with(:rspec) do |config|
+          config.before_verifying_doubles do |ref|
+            unless called_for.include? ref.target
+              called_for << ref.target
+              ::RSpec::Mocks.space.proxy_for(ref.target)
+            end
+          end
+          expect { allow(klass.new).to receive(:message) }.to_not raise_error
+          called_for.clear
```
myronmarston: Why is the `clear` necessary?

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -383,6 +383,22 @@ def defined_private_method
...
+          config.verifying_double_callbacks.clear
+        end
+      end
```
myronmarston: It's a bit hard to read this spec with it having no blank lines separating the parts.  Can you reformat it a bit?

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -383,6 +383,22 @@ def defined_private_method
         object.implemented
       end
 
+      it 'wraps proxies in a reentrant mutex so we dont trigger deadlocks' do
```
myronmarston: Our use of a reentrant mutex is an implementation detail, IMO, and isn't part of the public facing behavior of RSpec.  I'd phrase this as:

``` ruby
it 'avoids deadlocks when a proxy is accessed from within a `before_verifying_doubles` callback'
```

myronmarston: Also, I feel like this spec would better belong in `space_spec.rb`.  After all, that's where the bug was and the fix has nothing to do with partial doubles.

spec/rspec/mocks/partial_double_spec.rb
JonRowe: I was attempting to resolve the Ruby 2.0.0 warning

spec/rspec/mocks/partial_double_spec.rb
JonRowe: We don't have any specs along these lines there, I felt it was similar in spirit to these specs and this is surfaced by this call back, if its moved elsewhere I'll need another way to trigger it.

myronmarston: We don't have any specs that access `RSpec::Mocks.space.proxy_for` in this file.  This bug kinda straddles things but since we're fixing it by making the space proxy mutex re-entrant it feels to me like it belongs there.

That's subjective, though.  If you feel strongly about it staying here let's go with it.

spec/rspec/mocks/partial_double_spec.rb
myronmarston: I don't see the connection with the ruby 2.2.0 warning.  The 2.2.0 warning I see is this:

```
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:404: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:439: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:444: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:449: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:457: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:461: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:469: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:475: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:484: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:511: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:515: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:516: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:521: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:532: warning: possible reference to past scope - klass
/Users/myron/code/rspec-dev/repos/rspec-mocks/spec/rspec/mocks/partial_double_spec.rb:546: warning: possible reference to past scope - klass
```

...because there's a `let(:klass)` below and a local variable `klass` in this spec.

I also tried commenting out `called_for.clear` and it didn't have any affect, so we can do away with it, I believe.

JonRowe: Hey I said attempting ;)

spec/rspec/mocks/partial_double_spec.rb
JonRowe: > We don't have any specs that access RSpec::Mocks.space.proxy_for in this file. This bug kinda straddles things but since we're fixing it by making the space proxy mutex re-entrant it feels to me like it belongs there.

I'm trying to write a spec that will trigger the deadlock without the callback so we can do away with this, but struggling currently.

>  If you feel strongly about it staying here let's go with it.

I don't feel that strongly about it, but the deadlock error is trigger by use of the callback

JonRowe: I'm only using `RSpec::Mocks.space.proxy_for ` here because normal mocks usage triggers the thread deadlock, but then triggers an infinite loop anyway so i needed something more direct.

JonRowe:
Updated to hopefully avoid the 2.2 bug

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -383,6 +383,22 @@ def defined_private_method
...
 
+      it 'avoids deadlocks when a proxy is accessed from within a `before_verifying_doubles` callback' do
+        _klass = Class.new { def message; end; }
+        called_for = []
+
+        RSpec.configuration.mock_with(:rspec) do |config|
+          config.before_verifying_doubles do |ref|
+            unless called_for.include? ref.target
+              called_for << ref.target
+              ::RSpec::Mocks.space.proxy_for(ref.target)
```
JonRowe: @myronmarston are you happy with this for now?

myronmarston:
LGTM

JonRowe:
Hmm, does this need a changelog, I mean it's technically an internal refactoring but also a bug fix...

myronmarston:
It is a bug fix, so I think so.

JonRowe:
Hm, I think that #980 should get the changelog entry, and this is actually a refactoring of that for 3.4...

PR #980, opened 2015-06-25
Switch to reentrant mutex 3-3-maintenance

JonRowe:
This is #979 ported to 3-3-maintenance

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,11 @@
+### 3.3.2 Development
+[Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.3.1...3-3-maintenance)
+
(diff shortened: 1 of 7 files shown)


JonRowe:
Just awaiting a build on this one @myronmarston

myronmarston:
LGTM...but it looks like it didn't build.  Do you want to kick the build?

JonRowe:
It did build, its on the commit above (theres a weird bug/feature with builds these days whereby if the last commit doesn't have a build status (e.g. due to [skip ci]) then the PR ignores any other status).

JonRowe:
Turns out my changelog entry broke appveyor because of a whitespace eol, oops.

PR #983, opened 2015-06-30
Add means to fail test for expectations set on nil

lizrush:
This PR adds a configuration option for setting expectations on nil. Previously there was an option to `allow_expectations_on_nil` that would either warn by default or suppress the warnings when an expectation was set on nil.

Now, we have the following options that can be set inside an RSpec configuration:
* Do nothing with the config flag and retain the default behavior:
  * RSpec will warn the user when an expectation has been set on nil, but the test will not fail
* Set `allow_expectations_on_nil` to `true`:
  * RSpec will suppress the warning messages
* Set `allow_expectations_on_nil` to `false`:
  * RSpec will fail the test when the expectation has been set on nil.

This will help catch false positives and help prevent users from thinking their tests are fine when green but warnings that were given were not read.

### DIFF
--- a/lib/rspec/mocks/configuration.rb
+++ b/lib/rspec/mocks/configuration.rb
@@ -3,11 +3,28 @@ module RSpec
     # Provides configuration options for rspec-mocks.
     class Configuration
       def initialize
+        @allow_message_expectations_on_nil = nil
         @yield_receiver_to_any_instance_implementation_blocks = true
         @verify_doubled_constant_names = false
         @transfer_nested_constants = false
         @verify_partial_doubles = false
       end
+
+      # Sets whether RSpec will warn, ignore, or fail a test when
+      # expectations are set on nil.
+      # By default, when this flag is not set, warning messages are issued when
+      # expectations are set on nil. This is to prevent false-positives and to
+      # catch potential bugs early on.
+      # When set to `true`, warning messages are suppressed.
+      # When set to `false`, it will raise an error.
+      #
+      # @example
+      #   RSpec.configure do |config|
+      #     config.mock_with :rspec do |mocks|
+      #       mocks.allow_message_expectations_on_nil = false
+      #     end
+      #   end
+      attr_accessor :allow_message_expectations_on_nil
 
       def yield_receiver_to_any_instance_implementation_blocks?
         @yield_receiver_to_any_instance_implementation_blocks
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -213,6 +213,16 @@ module RSpec
         notify MockExpectationAlreadyInvokedError.new(error_message)
       end
 
+      def raise_expectation_on_nil_error(method_name)
+        __raise expectation_on_nil_message(method_name)
+      end
+
+      def expectation_on_nil_message(method_name)
+        "An expectation of :#{method_name} was set on nil. " \
+          "To allow expectations on nil & suppress this message, set allow_expectations_on_nil to true. " \
+          "To disallow expectations on nil, set allow_expectations_on_nil to false"
+      end
+
     private
 
       def received_part_of_expectation_error(actual_received_count, args)
--- a/lib/rspec/mocks/example_methods.rb
+++ b/lib/rspec/mocks/example_methods.rb
@@ -197,16 +197,9 @@ module RSpec
       # By default warning messages are issued when expectations are set on
       # nil.  This is to prevent false-positives and to catch potential bugs
       # early on.
+      # @deprecated Use {RSpec::Mocks::Configuration#allow_message_expectations_on_nil} instead.
       def allow_message_expectations_on_nil
         RSpec::Mocks.space.proxy_for(nil).warn_about_expectations = false
-      end
-
-      # Raise an error when setting expectations on nil.
-      #
-      # By default warning messages are issued when expectations are set on
-      # nil.  This is to force test failures when an expectation is set on nil.
-      def disallow_message_expectations_on_nil
-        RSpec::Mocks.space.proxy_for(nil).disallow_expectations = true
       end
 
       # Stubs the named constant with the given value.
--- a/lib/rspec/mocks/example_methods.rb
+++ b/lib/rspec/mocks/example_methods.rb
@@ -201,6 +201,14 @@ module RSpec
         RSpec::Mocks.space.proxy_for(nil).warn_about_expectations = false
       end
 
+      # Raise an error when setting expectations on nil.
+      #
+      # By default warning messages are issued when expectations are set on
+      # nil.  This is to force test failures when an expectation is set on nil.
+      def disallow_message_expectations_on_nil
+        RSpec::Mocks.space.proxy_for(nil).disallow_expectations = true
+      end
+
       # Stubs the named constant with the given value.
       # Like method stubs, the constant will be restored
       # to its original value (or lack of one, if it was
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -407,23 +407,29 @@ module RSpec
     class ProxyForNil < PartialDoubleProxy
       def initialize(order_group)
         @warn_about_expectations = true
+        @disallow_expectations = false
         super(nil, order_group)
       end
 
       attr_accessor :warn_about_expectations
+      attr_accessor :disallow_expectations
       alias warn_about_expectations? warn_about_expectations
+      alias disallow_expectations? disallow_expectations
 
       def add_message_expectation(method_name, opts={}, &block)
+        raise_error(method_name) if disallow_expectations?
         warn(method_name) if warn_about_expectations?
         super
       end
 
       def add_negative_message_expectation(location, method_name, &implementation)
+        raise_error(method_name) if disallow_expectations?
         warn(method_name) if warn_about_expectations?
         super
       end
 
       def add_stub(method_name, opts={}, &implementation)
+        raise_error(method_name) if disallow_expectations?
         warn(method_name) if warn_about_expectations?
         super
       end
@@ -433,6 +439,11 @@ module RSpec
       def warn(method_name)
         source = CallerFilter.first_non_rspec_line
         Kernel.warn("An expectation of :#{method_name} was set on nil. Called from #{source}. Use allow_message_expectations_on_nil to disable warnings.")
+      end
+
+      def raise_error(method_name)
+        source = CallerFilter.first_non_rspec_line
+        raise("An expectation of :#{method_name} was set on nil. Called from #{source}. To allow expectations on nil, set disallow_expectations_on_nil to false.")
       end
     end
   end
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -406,44 +406,62 @@ module RSpec
     # @private
     class ProxyForNil < PartialDoubleProxy
       def initialize(order_group)
-        @warn_about_expectations = true
-        @disallow_expectations = false
+        set_expectation_behavior
         super(nil, order_group)
       end
 
-      attr_accessor :warn_about_expectations
       attr_accessor :disallow_expectations
-      alias warn_about_expectations? warn_about_expectations
-      alias disallow_expectations? disallow_expectations
+      attr_accessor :warn_about_expectations
 
       def add_message_expectation(method_name, opts={}, &block)
-        raise_error(method_name) if disallow_expectations?
-        warn(method_name) if warn_about_expectations?
+        warn_or_raise!(method_name)
         super
       end
 
       def add_negative_message_expectation(location, method_name, &implementation)
-        raise_error(method_name) if disallow_expectations?
-        warn(method_name) if warn_about_expectations?
+        warn_or_raise!(method_name)
         super
       end
 
       def add_stub(method_name, opts={}, &implementation)
-        raise_error(method_name) if disallow_expectations?
-        warn(method_name) if warn_about_expectations?
+        warn_or_raise!(method_name)
         super
       end
 
     private
 
+      def set_expectation_behavior
+        case RSpec::Mocks.configuration.allow_message_expectations_on_nil
+        when false
+          @warn_about_expectations = false
+          @disallow_expectations = true
+        when true
+          @warn_about_expectations = false
+          @disallow_expectations = false
+        else
+          @warn_about_expectations = true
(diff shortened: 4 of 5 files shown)


lib/rspec/mocks/proxy.rb
```diff
@@ -407,23 +407,29 @@ class PartialClassDoubleProxy < PartialDoubleProxy
     class ProxyForNil < PartialDoubleProxy
       def initialize(order_group)
         @warn_about_expectations = true
+        @disallow_expectations_on_nil = false
```
JonRowe: As this has no real ties on `nil` and is already on a class called `ProxyForNil`, this and all corresponding methods should really just be `@disallow_expectations`

lib/rspec/mocks/proxy.rb
```diff
@@ -407,23 +407,29 @@ class PartialClassDoubleProxy < PartialDoubleProxy
...
 
       def add_message_expectation(method_name, opts={}, &block)
+        raise_error(method_name) if disallow_expectations_on_nil?
         warn(method_name) if warn_about_expectations?
         super
       end
 
       def add_negative_message_expectation(location, method_name, &implementation)
+        raise_error(method_name) if disallow_expectations_on_nil?
         warn(method_name) if warn_about_expectations?
```
JonRowe: Can we collapse these into one if check for performance reasons? e.g:

```Ruby
...
warn_or_raise(method_name) if warn_or_disallow_expectations?
...

def warn_or_raise(method_name)
  if disallow_expectations?
    raise_error(method_name)
  else
    warn(method_name)
  end
end
```

We'd also need to refactor the accessors to make that work but it means that a happy path will only encounter 1 if check.

lib/rspec/mocks/example_methods.rb
```diff
@@ -201,6 +201,14 @@ def allow_message_expectations_on_nil
         RSpec::Mocks.space.proxy_for(nil).warn_about_expectations = false
       end
 
+      # Raise an error when setting expectations on nil.
+      #
+      # By default warning messages are issued when expectations are set on
+      # nil.  This is to force test failures when an expectation is set on nil.
+      def disallow_message_expectations_on_nil
```
myronmarston: This mirrors the existing API (`allow_message_expectations_on_nil`) being something you set for a particular example...but that's always felt odd to me given it's the only config we handle that way.  I think many people are going to want  this set for all examples and to not have to set it in each example to achieve that.

WDYT about moving this to a global config option?

``` ruby
RSpec.configure do |config|
  config.mock_with :rspec do |mocks|
    mocks.allow_message_expectations_on_nil = false
  end
end
```

lizrush: @myronmarston That is exactly where I was expecting to find this when I was rooting around looking to mimic the existing code for `warn_about_expectations`, myself. I would like it as a global config, as well as with `warn`. Seems to make more sense as a config option than to have to set it for each example.

lib/rspec/mocks/proxy.rb
lizrush: :ok_hand:

lib/rspec/mocks/proxy.rb
lizrush: Will do.

lib/rspec/mocks/example_methods.rb
myronmarston: I think `allow_message_expectations_on_nil` originated in RSpec 1 (before rspec-mocks had its own config) and it just never got moved to global config once we added that.

I'm thinking a single config option (`allow_message_expectations_on_nil`) can suffice. If not set (e.g. `nil`) it will warn as it normally does.  If set to `true` it silences the warning.  If set to `false` it will raise an error instead of warning.

For backwards compatibility we'll have to keep the `allow_message_expectations_on_nil` method for now.

lizrush: Sounds good to me! I'll try to update this PR shortly.

xaviershay: I like the global config approach.

spec/rspec/mocks/nil_expectation_warning_spec.rb
```diff
@@ -24,6 +24,13 @@ module Mocks
         nil.foo
       end
 
+      it "raises an error when expectations on nil are disallowed" do
+        disallow_message_expectations_on_nil
+
+        expect { expect(nil).to receive(:foo)     }.to raise_error(StandardError)
```
xaviershay: nit: particularly with such a generic error type, I'd typically do `raise_error(/foo/)` to make sure the error is a) relevant, and b) contains relevant debugging information (tests the interpolation)

lib/rspec/mocks/proxy.rb
```diff
@@ -434,6 +440,11 @@ def warn(method_name)
         source = CallerFilter.first_non_rspec_line
         Kernel.warn("An expectation of :#{method_name} was set on nil. Called from #{source}. Use allow_message_expectations_on_nil to disable warnings.")
       end
+
+      def raise_error(method_name)
+        source = CallerFilter.first_non_rspec_line
+        raise("An expectation of :#{method_name} was set on nil. Called from #{source}. To allow expectations on nil, set disallow_expectations_on_nil to false.")
```
xaviershay: I wonder if we have an appropriate exception type to either raise or subclass here rather than `StandardError`? I tend to like errors coming from the framework to be of a framework type, but can't remember what we do elsewhere in this project.

lizrush: @xaviershay Who would I ask about this to find out?

xaviershay: @myronmarston or @JonRowe  will chime in :)

Grepping around I found https://github.com/rspec/rspec-mocks/blob/master/lib/rspec/mocks/error_generator.rb which suggests two things:

1. We do use custom exception classes, but they inherit from `StandardError` rather than a common mocks type. However, we also typically use `ArgumentError` and `NoMethodError` (ruby built-ins), either of which is almost applicable here. Also found a single example of raising `StandardError` at https://github.com/rspec/rspec-mocks/blob/master/lib/rspec/mocks/mutate_const.rb#L329 so...
2. We may want to use `ErrorGenerator`, which delegates to `RSpec::Support.notify_failure`. We use that method to support things like aggregate failures (I think). [We (rspec team) outside of this PR should probably add some class doc to explain why that class exists and when it is/isn't appropriate to use.] This isn't used everywhere, but I think it's a fairly recent addition and just not everything is ported to it yet maybe? It appears to handle similar types of failures to the one you're adding here.

myronmarston: Please use `ErrorGenerator`.  It is intended to be the central place for errors to be generated (as the name implies) although that hasn't always been followed.  Having it delegate to `RSpec::Support.notify_failure` will be good to make this work with `aggregate_failures` and be treated as a normal rspec expectation failure.

spec/rspec/mocks/nil_expectation_warning_spec.rb
JonRowe: It will eventually be a `RSpec::Mocks::MockExpectationError` if you use the error generator like Myron suggests!

lizrush:
Hey @xaviershay, @myronmarston, @JonRowe! Can I get a second look :eyes: at this guy when you've got the chance? Thanks.

lib/rspec/mocks/configuration.rb
```diff
@@ -3,12 +3,29 @@ module Mocks
...
       end
 
+      # Sets whether RSpec will warn, ignore, or fail a test when
+      # expectations are set on nil.
+      # By default, when this flag is not set, warning messages are issued when
+      # expectations are set on nil. This is to prevent false-positives and to
+      # catch potential bugs early on.
+      # When set to `true`, warning messages are suppressed.
+      # When set to `false`, it will raise an error.
+
```
myronmarston: Thanks for taking the time to write up good docs for this!

This blank line causes YARD to not understand that this command block above is about `allow_message_expectations _on_nil`.  Here's how it renders right now:

screen shot 2015-07-07 at 2 26 45 pm

I think if you add a line with only a `#` like this:

``` ruby
# Some comments
#
# Some more comments
```

...then YARD will render it properly.  You can check how YARD renders the docs by running `yard server --reload`.  It's generally a good idea to check that when writing the docs as their can be gotchas where things don't render how you expect.

lib/rspec/mocks/error_generator.rb
```diff
@@ -213,6 +213,11 @@ def raise_already_invoked_error(message, calling_customization)
         notify MockExpectationAlreadyInvokedError.new(error_message)
       end
 
+      def raise_expectation_on_nil_error(method_name, source)
+        __raise "An expectation of :#{method_name} was set on nil. Called from #{source}. " \
+          "To allow expectations on nil, set allow_expectations_on_nil to true."
```
myronmarston: * There's a fair bit of duplication between this method and `ProxyForNil#warn` (the message text is identical, right?).  WDYT about extracting a `expectation_on_nil_message` helper method that both can delegate to?
* The "Called from source" bit is unnecessary for an exception -- exception carry backtrace info and our exception formatting will include that.  `Kernel.warn` (which is used by `ProxyForNil#warn`) does not include the call site so we were including it manually before.  Since the message was originally written, we've added `RSpec.warning` as a newer standardized way to emit warnings, which takes care of including the source call site in the message...so I think we'll also want `ProxyForNil#warn` updated to use that rather than `Kernel.warn`.

lib/rspec/mocks/proxy.rb
```diff
@@ -406,34 +406,66 @@ class PartialClassDoubleProxy < PartialDoubleProxy
...
+        @disallow_expectations || @warn_about_expectations
+      end
+
+      def warn_or_raise(method_name)
+        if disallow_expectations
+          raise_error(method_name)
+        else
+          warn(method_name)
+        end
+      end
```
myronmarston: The `@warn_about_expecations` and `@disallow_expectations` variables are checked multiple times, and I'm thinking we can clean this up considerably by introducing a polymorphic interface and 3 implementations.  I'm thinking of something like this:

``` ruby
module Notifiers
  module Null
    def self.notify(*); end
  end

  class Warn
    def initialize(error_generator)
      @error_generator = error_generator
    end

    def notify(method_name)
      msg = @error_generator.exception_on_nil_message(method_name)
      RSpec.warning(msg)
    end
  end

  class Raise
    def initialize(error_generator)
      @error_generator = error_generator
    end

    def notify(method_name)
      @error_generator.raise_expecation_on_nil_error(method_name)
    end
  end
end

def set_expectation_behavior
  @notifier = case RSpec::Mocks.configuration.allow_message_expectations_on_nil
    when false then Notifiers::Raise.new(@error_generator)
    when true  then Notiifers::Null
    else Notifiers::Warn.new(@error_generator)
  end
end
```

Then instead of calling `warn_or_raise(method_name) if warn_or_disallow_expectations?`, you can just call `@notifier.notify(method_name)` and trust it'll do the right thing.

Thoughts?

lib/rspec/mocks/proxy.rb
```diff
@@ -406,34 +406,66 @@ class PartialClassDoubleProxy < PartialDoubleProxy
...
+
       def warn(method_name)
         source = CallerFilter.first_non_rspec_line
         Kernel.warn("An expectation of :#{method_name} was set on nil. Called from #{source}. Use allow_message_expectations_on_nil to disable warnings.")
       end
+
+      def raise_error(method_name)
+        source = CallerFilter.first_non_rspec_line
+        ErrorGenerator.new(self).raise_expectation_on_nil_error(method_name, source)
```
myronmarston: It's a bit hard to tell in this context due to the subclassing but there's actually already an error generator instance available via `@error_generator` (see `Proxy#initialize`), so as I demonstrated above it's better to re-use that instance instead of making a new one.

spec/rspec/mocks/nil_expectation_warning_spec.rb
```diff
@@ -15,13 +15,27 @@ module Mocks
         expect(nil).not_to receive(:foo)
       end
 
-      it "does not issue a warning when expectations are set to be allowed" do
-        allow_message_expectations_on_nil
```
myronmarston: While the new global config way of configuring this is the way we'll recommend going forward, it's important that we still support this old interface (since it's a public API) until RSpec 4, so we'll need to keep existing specs around for it.

I do think we should soft-deprecate it, though.  Can you update the YARD docs for that method and add a [deprecated](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#deprecated) tag?  We won't issue a deprecation warning yet -- that'll come in 3.99 as part of the upgrade-to-RSpec-4 process -- but the YARD docs will help direct people to the new way to configure this.

spec/rspec/mocks/nil_expectation_warning_spec.rb
```diff
@@ -15,13 +15,27 @@ module Mocks
...
+      end
+
+      context 'configured to disallow expectations on nil' do
+        include_context "with isolated configuration"
+
+        it "raises an error when expectations on nil are disallowed" do
+          RSpec::Mocks.configuration.allow_message_expectations_on_nil = false
+          expect { expect(nil).to receive(:foo)     }.to raise_error(RSpec::Mocks::MockExpectationError)
+          expect { expect(nil).not_to receive(:bar) }.to raise_error(RSpec::Mocks::MockExpectationError)
+        end
```
myronmarston: :+1: Nice work on these specs :).

myronmarston:
Thanks, @lizrush, this is looking good!  I left a few more comments.  If you're short on time (or desire) to address the feedback, let us know -- one of us can merge as-is and address the comments I left above in a follow-up PR.

spec/rspec/mocks/nil_expectation_warning_spec.rb
lizrush: Thanks! Testing is one of my weaker skills :smile_cat:

myronmarston: Working on a testing library seems like a good way to improve :smile:

lizrush:
Ok, fair enough @myronmarston! I'm still fairly new to coding so I tend to go with simpler solutions :blush: 

Let me know if there's anything else needed to wrap this guy up.

myronmarston:
This looks good now.  Do you mind squashing it into a smaller number of commits before I merge?

lizrush:
:ok_woman:  Done!

JonRowe:
Thanks @lizrush, good work :)

Issue #921, opened 2015-04-09
receive_message_chain ... exactly() || twice() || <similar>() produces undefined method


brycemcd:
I'm testing a piece of code in a rails controller like:

```ruby
things.each do |thing|
  current_user.all_the_things.create(thing)
end

render json: {success: true}
```

I have a mock like:

```ruby
allow_any_instance_of(User).to receive_message_chain('all_the_things.create').
  exactly(2).times.and_return(true)
# ... more setup
expect(response).to be_ok
```

When the spec is run, an exception is thrown:
`undefined method 'exactly' for for #RSpec::Mocks::Matchers::ReceiveMessageChain:0x0000000afb3d08>`

The same happens if `exactly` is replaced with `twice` or [anything else defined in the documentation](https://relishapp.com/rspec/rspec-mocks/docs/setting-constraints/receive-counts)

I'm happy to provide a more concrete test case (and pull request), but before I get too deep into it I want to confirm that the behavior I'm expecting is rational.

According to [the message chain docs](https://relishapp.com/rspec/rspec-mocks/v/3-2/docs/working-with-legacy-code/message-chains), the `receive_message_chain` method should be a replacement for the `receive` message. My expectation is that the `receive_message_chain` message should be allowed to have the 'receive count' messages chained to them in the same way that `receive` does.

All that said, this whole approach feels a bit like a code smell, but I'm not confident of another approach. Law of Demeter would suggest not chaining `current_user.all_the_things` to `create`. However, it's good practice in rails to create relationships belonging to the user by sending messages to related objects through the original object (`current_user` in this case).

TL;DR: is the receive_message_chain behavior described above a bug or a feature?

myronmarston:
It's by design. `receive_message_chain` isn't intended to support the full level of customization that you get with `receive`.

The docs could perhaps be improved to make that more clear -- want to open a PR to improve that?

brycemcd:
yeah, happy to open a PR.

What is the intended support of `receive_message_chain` vs. `receive` ?

The more I noodle on the use case above, the more I think the 'receive count' messages make sense.

myronmarston:
[...]

PR #984, opened 2015-07-03
Doc Update - feature parity of receive_message_chain with receive

brycemcd:
Updates docs to alleviate some confusion I ran into once.

doc update for `receive_message_chain` and `receive` parity
  * summarizes conversation from rspec/rspec-mocks#921
  * adds warning

I'm not 100% confident I captured the conversation, but happy to improve this PR with a little more direction.

### DIFF
--- a/features/working_with_legacy_code/message_chains.feature
+++ b/features/working_with_legacy_code/message_chains.feature
@@ -1,6 +1,6 @@
 Feature: Message Chains
 
-  You can use `receive_message_chain` in place of `receive` in certain instances
+  You can use `receive_message_chain` in place of `receive` in certain circumstances
   to stub a chain of messages:
 
   ```ruby
@@ -23,9 +23,9 @@ Feature: Message Chains
   
   `receive_message_chain` is designed to be used with evaluating a response like `and_return`, 
   `and_yield` etc. For legacy reasons, parity with `stub_chain` is supported but its uses are
-  not considered good practice. Support for `stub_chain` pairity may be removed in future versions.
+  not considered good practice. Support for `stub_chain` parity may be removed in future versions.
   
-  Common method chains like `with` and `exactly` (i.e. `exactly(2).times`) are not supported.
+  Customisations like `exactly` (i.e. `exactly(2).times`) are not supported.
 
   Warning:
   ========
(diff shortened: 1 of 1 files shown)


features/working_with_legacy_code/message_chains.feature
```diff
@@ -19,6 +20,12 @@ Feature: Message Chains
   ```ruby
   allow(Article).to receive_message_chain("recent.published") { [Article.new] }
   ```
+  
+  `receive_message_chain` is designed to be used with evaluating a response like `and_return`, 
+  `and_yield` etc. For legacy reasons, parity with `stub_chain` is supported but its uses are
+  not considered good practice. Support for `stub_chain` pairity may be removed in future versions.
```
JonRowe: s/pairity/parity

features/working_with_legacy_code/message_chains.feature
```diff
@@ -19,6 +20,12 @@ Feature: Message Chains
...
+  
+  Common method chains like `with` and `exactly` (i.e. `exactly(2).times`) are not supported.
```
JonRowe: We refer to these as customisations internally, and I think this is confusing to call them chains here when we're talking about the chain matcher, chaining here refers to the target not how we set up the expectation.

JonRowe: Also I'm fairly certain with is supported...

features/working_with_legacy_code/message_chains.feature
```diff
@@ -1,6 +1,7 @@
 Feature: Message Chains
 
-  You can use `receive_message_chain` in place of `receive` to stub a chain of messages:
+  You can use `receive_message_chain` in place of `receive` in certain instances
```
JonRowe: "in certain instances" reads a little funny to me, in certain circumstances would probably be better but I'm not sure this change is needed at all

brycemcd:
@JonRowe updated. Should be good go.

FWIW - it's not clear to me why the tests aren't passing but this PR should not have any impact on them

JonRowe:
If you rebase from master they'll go away, they're due to the build being broken by a version of a gem at the time you forked. You can also amend the commit message with [skip ci] when dealing with doco only changes.

JonRowe:
Thanks

myronmarston:
Thanks, @brycemcd!

JonRowe:
Ah the build failure was actually due to whitespace on the EOLs, my bad should have checked better, fixed.

PR #985, opened 2015-07-05
Pin aruba to version that works on 1.8.7.

myronmarston:
Works around cucumber/aruba#279.


myronmarston:
The build failures are from the rspec-core part of the build, which'll be fixed by rspec/rspec-core#2022.  Merging.

PR #987, opened 2015-07-10
Apply some slight improvements to nil expectation warning logic.

myronmarston:
- Use backticks in the warning to indicate code snippets.
- Make specs slightly less brittle by preferring the `output`
  matcher over mocking `Kernel.warn` (an interface we don’t own),
  and by specifying only a few important bits of the warning.

Followup to #983.


JonRowe:
LGTM

PR #991, opened 2015-07-19
Mention `and_return` in "Allowing messages" feature

andrii:
Even though `and_return` is covered in detail in following features, it would make sense to mention how to return a value while allowing a single message because it's more likely the use case a reader is looking for when they start reading the Basics topic.

### DIFF
--- a/features/basics/allowing_messages.feature
+++ b/features/basics/allowing_messages.feature
@@ -19,6 +19,20 @@ Feature: Allowing messages
      When I run `rspec allow_message_spec.rb`
      Then the examples should all pass
 
+  Scenario: Allowed messages can return values using `and_return`
+    Given a file named "allow_message_and_return_value_spec.rb" with:
+      """ruby
+      RSpec.describe "and_return" do
+        it "returns the specified return value" do
+          dbl = double("Some Collaborator")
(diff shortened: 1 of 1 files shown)


JonRowe:
Hi as this isn't just a documentation change, (you've added a scenario) can you please amend your commit removing `[skip ci]` so the build runs?

We also do have a whole file dedicated to responding with `and_return` that makes this a bit superfluous, but I'm ok including it as a simple example, maybe it's worth mentioning that file in the preamble however

myronmarston:
Given that we've got an [entire feature file](https://relishapp.com/rspec/rspec-mocks/v/3-3/docs/configuring-responses/returning-a-value) on `and_return`, it seems organizationally odd (to me) to put this scenario here.

Instead, what do you think about updating the narrative to link to the [configuring responses](https://relishapp.com/rspec/rspec-mocks/v/3-3/docs/configuring-responses) section?  That'll allow people who come to this feature file to easily navigate to where it explains how to configure responses (such as `and_return`).

andrii:
I gave it more thought (as I should have in the first place) and now I agree that it's superfluous. Configuring responses also applies to expecting messages and mentioning it in that section as well would be too much. I really like how succinct the features are.

Issue #997, opened 2015-07-28
allow().to receive doesn't respect order


rwz:
```ruby
require "rspec/autorun"

class MyClass
  def initialize(client)
    @client = client
  end

  def fetch_results
    [].tap do |results|
      results << @client.get_value while results.length < 3
    end
  end
end

describe MyClass do
  let(:client){ double("Client") }
  subject{ described_class.new(client) }

  it "works" do
    expect(client).to receive(:get_value).and_return(1).ordered
    expect(client).to receive(:get_value).and_return(2).ordered
    expect(client).to receive(:get_value).and_return(3).ordered

    expect(subject.fetch_results).to eq([1,2,3])
  end

  it "should work, but fails" do
    allow(client).to receive(:get_value).and_return(1).ordered
    allow(client).to receive(:get_value).and_return(2).ordered
    allow(client).to receive(:get_value).and_return(3).ordered

    expect(subject.fetch_results).to eq([1,2,3])
  end
end
```

This is expected to pass in both cases but only actually passes with `expect` syntax.

```
Failures:

  1) MyClass should work, but fails
     Failure/Error: expect(subject.fetch_results).to eq([1,2,3])

       expected: [1, 2, 3]
            got: [3, 3, 3]

       (compared using ==)
     # run.rb:32:in `block (2 levels) in <main>'

Finished in 0.01525 seconds (files took 0.09345 seconds to load)
2 examples, 1 failure

Failed examples:

rspec 'run.rb[1:2]' # MyClass should work, but fails
```

myronmarston:
If you want to return different sequential values for multiple calls, just pass multiple values to `and_return`:

https://relishapp.com/rspec/rspec-mocks/v/3-3/docs/configuring-responses/returning-a-value#specify-different-return-values-for-multiple-calls

The problem in your example is that stubs _replace_ earlier stubs rather than specifying responses for later calls.  This is intended to support cases like:

``` ruby
RSpec.describe do
  before do
    allow(Collaborator).to receive(:foo).and_return(:generic_response)
  end

  it "does x" do
    # ...
  end

  it "does y" do
    # ...
  end

  context "with a more specific response" do
    before do
      allow(Collaborator).to receive(:foo).and_return(:specific response)
    end

    it "does z" do
      # ...
    end
  end
end
```

Here the second `allow` overrides what was set as the response in the first `before` hook.  We can't make `allow` additive like you thought it was and support that.

[...]

PR #1000, opened 2015-08-04
Warn when `ordered` is used with `allow`

JonRowe:
An attempt at resolving #997, are there other scenarios we should explicitly warn about for `ordered`?

### DIFF
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -328,6 +328,13 @@ module RSpec
       #   expect(api).to receive(:run).ordered
       #   expect(api).to receive(:finish).ordered
       def ordered(&block)
+        if type == :stub
+          RSpec.warning(
+            "`allow(...).to receive(..).ordered` is not supported and will" \
+            "have no effect, use `and_return(*ordered_values)` instead."
+          )
+        end
+
         self.inner_implementation_action = block
         additional_expected_calls.times do
           @order_group.register(self)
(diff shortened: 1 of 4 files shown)


myronmarston:
Can this be implemented on the `MessageExpectation#ordered` method (by checking if `type == :stub`) instead?  That would automatically cover any uses of `ordered` when added to a stub using the legacy syntax or via `RSpec::Mocks.allow_message`.

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -267,6 +267,14 @@ def receiver.method_missing(*a); end # a poor man's stub...
         it_behaves_like "resets partial mocks cleanly" do
           let(:target) { allow(object) }
         end
+
+        context 'ordered with receive counts' do
+          specify 'is not supported' do
```
myronmarston: Why `specify` and not `it`?

Generally, I stick to `it` when the string starts with a verb (as that makes grammatical sense) but use `specify` when my doc string starts with something else (e.g. a noun) as such phrasing makes no sense off of `it`.

JonRowe: it didn't sit well when I wrote it given the context before

JonRowe:
Possibly, there's a large amount of indirection going on in this lot it was hard to determine the best place, I might give that a go another day (its 2330 here and that's -2 from my normal timezone, lol flying)

JonRowe:
Been a while but I finally refactored this @myronmarston, turns out your solution was much simplier

lib/rspec/mocks/message_expectation.rb
```diff
@@ -336,6 +336,13 @@ def with(*args, &block)
       #   expect(api).to receive(:run).ordered
       #   expect(api).to receive(:finish).ordered
       def ordered(&block)
+        if self.type == :stub
+          RSpec.warning(
+            "`allow(...).to receive` is not supported and will" \
```
myronmarston: `allow(...).to receive` _is_ supported.  It's `allow(...).to receive(...).ordered` that's not supported.  The `ordered` bit is important or this warning is very confusing.

JonRowe: Right you are, amended.

myronmarston:
LGTM, merge when green (+ add a changlog entry).

PR #1001, opened 2015-08-11
Updates from rspec-dev (2015-08-11)

JonRowe:

### DIFF
--- a/script/functions.sh
+++ b/script/functions.sh
@@ -1,4 +1,4 @@
-# This file was generated on 2015-08-01T02:04:28-04:00 from the rspec-dev repo.
+# This file was generated on 2015-08-11T23:21:08+01:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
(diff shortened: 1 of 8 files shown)


JonRowe:
Seeing if rspec/rspec-dev#116 is fixed

script/functions.sh
```diff
@@ -11,6 +11,10 @@ export JRUBY_OPTS=${JRUBY_OPTS:-"--server -Xcompile.invokedynamic=false"}
 SPECS_HAVE_RUN_FILE=specs.out
 MAINTENANCE_BRANCH=`cat maintenance-branch`
 
+# Don't allow rubygems to pollute what's loaded. Also, things boot
+# faster without the extra load time of rubygems.
+export RUBYOPT="--disable=gem"
```
myronmarston: It looks like this needs to be made conditional, so that it only runs on MRI 1.9+.  MRI 1.8.7 and JRuby fail with this option as it is unrecognized.

Issue #931, opened 2015-04-14
rspec-mocks + minitest 5.6 dont play nice together


arthurnn:
see https://github.com/jekyll/jekyll/issues/3656

Minitest 5.6 now defines a `expect` method, which collides with rspec-mocks.
What we can do about it:
 - Dont do anything, but I am sure people will open issues here confused.
 - Lock minitest version in rspec-mocks.
 - Give a warning to the user, so they will better understand the error and the solution.

thoughts?

myronmarston:
Good to know.  Ideally, I'd like to find a way to make rspec-mocks and minitest 5.6+ play nice together.  For rspec-expectations, we added a file that takes care of integrating it with minitest awhile back:

https://github.com/rspec/rspec-expectations/blob/v3.2.1/lib/rspec/expectations/minitest_integration.rb

https://github.com/rspec/rspec-expectations/blob/v3.2.1/features/test_frameworks/minitest.feature

I think we could similarly add a `rspec/mocks/minitest_integration` file that minitest users who use rspec-mocks could use to deal with integrating the two and handling the dual role of `expect`.  We'll need to think carefully about how to make `expect` compatible with both libs, though.

myronmarston:
Here's a simple way to make them compatible:

* Move the `to` and `not_to` methods currently defined on `RSpec::Mocks::ExpectationTarget` into a module (which is included into `RSpec::Mocks::ExpectationTarget` to keep the method available there).
* In the new `rspec/mocks/minitest_integration` file, we would also include the module in `Minitest::Expectation` if `Minitest::Expectation` is defined.

BTW, you can see the relevant minitest change here:

https://github.com/seattlerb/minitest/commit/9e78cc974f3ef0d9716f1cca2675753cf5f648d0

arthurnn:
[related https://github.com/jekyll/jekyll/pull/3658]

envygeeks:
Attaching myself to this ticket so I get notified if you guys make any changes and can remove our hacks from Jekyll.  Thanks @arthurnn and @myronmarston!

arthurnn:
I think , to fix this, we could disable expecations on minitest using that env var https://github.com/seattlerb/minitest/blob/master/lib/minitest/spec.rb#L327

myronmarston:
> I think , to fix this, we could disable expecations on minitest using that env var https://github.com/seattlerb/minitest/blob/master/lib/minitest/spec.rb#L327

[...]

PR #1004, opened 2015-08-17
Add integration with expect methods defined in minitest 5.6+

robotfelix:
Fixes issue #931.

Integration defines `#to`, `#not_to` and `#to_not` methods on `Minitest::Expectation`, which creates a new `RSpec::Mocks::ExpectationTarget` instance and calls the relevant method on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,19 @@ general we discourage its use for a number of reasons:
   which doesn't help.)
 
 
+## Using rspec-mocks with minitest 5.6+
+
+The latest versions of minitest also define an `expect` method. To use the
+`expect` method defined by rspec-mocks with minitest, load the following file
+**after** minitest is loaded:
+
+```ruby
+require "rspec/mocks/minitest_integration"
+```
+
+minitest style `expect` usage should continue to work as normal. 
+
+
 ## Further Reading
 
 There are many different viewpoints about the meaning of mocks and stubs. If
--- a/lib/rspec/mocks/minitest_integration.rb
+++ b/lib/rspec/mocks/minitest_integration.rb
@@ -1,0 +1,15 @@
+require 'rspec/mocks'
+
+Minitest::Expectation.class_eval do
+  def to(*args, &block)
+    ::RSpec::Mocks::ExpectationTarget.new(target).to(*args, &block)
+  end
+
+  def not_to(*args, &block)
+    ::RSpec::Mocks::ExpectationTarget.new(target).not_to(*args, &block)
+  end
+
+  def to_not(*args, &block)
+    ::RSpec::Mocks::ExpectationTarget.new(target).to_not(*args, &block)
+  end
+end


README.md
```diff
@@ -425,6 +425,19 @@ general we discourage its use for a number of reasons:
...
+
+The latest versions of minitest also define an `expect` method. To use the
+`expect` method defined by rspec-mocks with minitest, load the following file
+**after** minitest is loaded:
+
+```ruby
+require "rspec/mocks/minitest_integration"
+```
+
+minitest style `expect` usage should continue to work as normal.
```
myronmarston: One of the build failures is due to excess whitespace on the end of this line (we enforce not having excess EOL-whitespace to prevent future unneeded git churn for whitespace cleanup).

lib/rspec/mocks/minitest_integration.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def not_to(*args, &block)
+    ::RSpec::Mocks::ExpectationTarget.new(target).not_to(*args, &block)
+  end
+
+  def to_not(*args, &block)
+    ::RSpec::Mocks::ExpectationTarget.new(target).to_not(*args, &block)
+  end
+end
```
myronmarston: If we're going to provide minitest integration, I think we should go all the way and provide a file that handles _all_ of the integration instead of just this bit.  Our [cuke](https://github.com/rspec/rspec-mocks/blob/master/features/outside_rspec/minitest.feature) shows an example of the sort of code that a user has to write to integrate the two libraries, and that setup/verify/teardown logic should go in this file as well.  Then minitest users can just do `require 'rspec/mocks/minitest_integration'` and it'll handle all of the integration :).

myronmarston: BTW, once we do that I think the [cuke](https://github.com/rspec/rspec-mocks/blob/master/features/outside_rspec/minitest.feature) should be updated to use the new `rspec/mocks/minitest_integration` file instead of manually integrating it.  However, we still need a cuke showing how rspec-mocks can integration with _any_ test framework by calling the `setup`/`verify`/`teardown` methods at the appropriate time...so I think we should add a new cuke to take the place of the minitest one that integrates manually with some other test framework.  Maybe with the test-unit gem?  (As opposed to test-unit in stdlib which subclasses minitest in Ruby 1.9+).

lib/rspec/mocks/minitest_integration.rb
```diff
@@ -0,0 +1,15 @@
...
+Minitest::Expectation.class_eval do
+  def to(*args, &block)
+    ::RSpec::Mocks::ExpectationTarget.new(target).to(*args, &block)
+  end
+
+  def not_to(*args, &block)
+    ::RSpec::Mocks::ExpectationTarget.new(target).not_to(*args, &block)
+  end
+
+  def to_not(*args, &block)
```
myronmarston: I think `to_not` can be simplified to an `alias`:

``` ruby
alias to_not not_to
```

lib/rspec/mocks/minitest_integration.rb
```diff
@@ -0,0 +1,15 @@
+require 'rspec/mocks'
+
+Minitest::Expectation.class_eval do
```
myronmarston: If we implement the suggestion I left below about putting the setup/verify/teardown integration here as well then I think we should make this file work on all Minitest 5 versions, not just on Minitest 5.6+.  I believe that `Minitest::Expectation` is only defined on 5.6+ so we're probably going to have to make this conditional on `Minitest::Expectation` existing as a constant.

README.md
myronmarston: BTW, thanks for taking the time to add docs!

lib/rspec/mocks/minitest_integration.rb
```diff
@@ -0,0 +1,15 @@
...
+  def to(*args, &block)
+    ::RSpec::Mocks::ExpectationTarget.new(target).to(*args, &block)
```
myronmarston: RSpec's `expect(...).to`/`expect(...).not_to`/`expect(...).to_not` syntax has a bit of implementation trickiness around the fact that both rspec-mocks and rspec-expectations use that syntax, and both libraries must be usable individually (e.g. for a minitest user who uses one but not the other) and they also must be usable together.  Here's how this is currently implemented:

* Both libraries define an `ExpectationTarget` class (`RSpec::Expectations::ExpectationTarget` and `RSpec::Mocks::ExpectationTarget`).
* In rspec-mocks, the classes behind `receive`, `receive_messages`, `have_received` and `receive_message_chain` implement the rspec-expectations matcher protocol so that rspec-expectations' `expect` can treat them just like any other matcher without any knowledge of rspec-mocks.
* In rspec-mocks, we define `expect` in [such a way](https://github.com/rspec/rspec-mocks/blob/574e73b56eb703ae92a0b30be981961b98cb32ee/lib/rspec/mocks/example_methods.rb#L380-L386) that it is further up the superclass chain and would only get used if rspec-expectations is not included (either before or after rspec-mocks is included).

All that's to say that if the user is using rspec-mocks and rspec-expectations with minitest, `expect(...).to` is going to have to call `RSpec::Expectations::ExpectationTarget#to` instead of `RSpec::Mocks::ExpectationTarget#to` for things to work as the user expects.  This would perhaps need to delegate to the rspec-expectations `to` method here to work properly.  However, it's pretty clearly out of scope for this file to fix minitest's `expect` to work with rspec-expectations.  So I'm not exactly sure what we should do to make things work for all the cases....I really never expected minitest to add an `expect` method to interfere with ours.  I do think we should at least do some testing so that we understand how things work for the various cases (e.g. minitest with just rspec-mocks, minitest with just rspec-expectations, and minitest with both).  It would also be nice to confirm that whatever we have here does not interfere with rspec-expectations and minitest playing nicely together.

Does that all make sense?

lib/rspec/mocks/minitest_integration.rb
myronmarston: One other thought: it can often work better to define our monkey patch methods in a module we own, and mix that in to `Minitest::Expectation` rather than re-opening `Minitest::Expectation` directly.  Among other things, it means that `Minitest::Expectation.instance_method(:to).owner` would list itself as being defined by an rspec module rather than a minitest one.  See http://www.justinweiss.com/blog/2015/01/20/3-ways-to-monkey-patch-without-making-a-mess/ for more discussion of this technique.

myronmarston:
Hey @robotfelix, thanks for tackling this!  I left a bunch of feedback.  Hopefully it's not too overwhelming.  What you've done here is good but isn't going to provide the good experience we always strive for for various alternate environments (e.g. earlier minitest 5.x versions, mixing minitest with rspec-mocks and rspec-expectations, etc).  I think we'll want the feedback I left above addressed before this is ready to merge.  If you need help with it, please speak up!

README.md
```diff
@@ -425,6 +425,19 @@ general we discourage its use for a number of reasons:
   which doesn't help.)
 
 
+## Using rspec-mocks with minitest 5.6+
+
+The latest versions of minitest also define an `expect` method. To use the
```
JuanitoFatas: `minitest` should be `Minitest` :heavy_check_mark:

README.md
```diff
@@ -425,6 +425,19 @@ general we discourage its use for a number of reasons:
...
+`expect` method defined by rspec-mocks with minitest, load the following file
```
JuanitoFatas: ditto

README.md
JuanitoFatas: ditto

README.md
robotfelix: I wrote it as `Minitest` in my first draft, but then changed if after checking `minitest`'s docs (https://github.com/seattlerb/minitest) and finding that they seems to use the lowercase `minitest` more? The exception seems to be when it's at the start of a sentence.

robotfelix:
Sorry I haven't had time to make the amendments suggested yet. I'll try to get them done next weekend.

samphippen:
@robotfelix just wanted to loop back on this. Are you still able to carry this PR?

myronmarston:
Closing in favor of #1065.

PR #1006, opened 2015-08-19
Forward block received with message on to `have_received` block.

### LINKED ISSUE #1003: have_received with block argument is nil
If I do:

    subject
    expect(foo).to have_received(:bar) {|&block| expect(block.call).to eq 'something' }

I get undefined method call for nil class-- block is nil inside the block.

But if I do:

    expect(foo).to receive(:bar) {|&block| expect(block.call).to eq 'something' }
    subject

Then it works.

myronmarston:
Fixes #1003.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,11 @@ Enhancements:
 * Add option for failing tests when expectations are set on `nil`.
   (Liz Rush, #983)
 
+Bug Fixes:
+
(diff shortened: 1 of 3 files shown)


myronmarston:
Green and ready for review, @rspec/rspec.

samphippen:
:+1:

PR #1013, opened 2015-09-22
Allow stubbing respond_to? on partial doubles

JonRowe:
Previously stubbing `respond_to?` would causing an infinite loop when hit, this prevents that.

//cc @samphippen

### DIFF
--- a/lib/rspec/mocks/method_reference.rb
+++ b/lib/rspec/mocks/method_reference.rb
@@ -94,7 +94,24 @@ module RSpec
           # write `method_missing` in such a way that it handles a dynamic message
           # with private or protected visibility. Ruby doesn't provide you with
           # the caller info.
-          return :public if vis.nil? && object.respond_to?(method_name, false)
+          return :public if vis.nil? && object_responds_to?(object, method_name, false)
+        end
+      end
+
+      # @api private
+      KERNEL_RESPOND_TO_METHOD = ::Kernel.instance_method(:respond_to?)
+
+      if Support::RubyFeatures.supports_rebinding_module_methods?
+        def self.object_responds_to?(object, method_name, visibility)
+          KERNEL_RESPOND_TO_METHOD.bind(object).call(method_name, visibility)
+        end
+      else
+        def self.object_responds_to?(object, method_name, visibility)
+          if ::Kernel === object
+            KERNEL_RESPOND_TO_METHOD.bind(object).call(method_name, visibility)
+          else
+            object.respond_to?(method_name, visibility)
+          end
         end
       end
     end
--- a/lib/rspec/mocks/method_reference.rb
+++ b/lib/rspec/mocks/method_reference.rb
@@ -94,24 +94,16 @@ module RSpec
           # write `method_missing` in such a way that it handles a dynamic message
           # with private or protected visibility. Ruby doesn't provide you with
           # the caller info.
-          return :public if vis.nil? && object_responds_to?(object, method_name, false)
-        end
-      end
+          proxy = RSpec::Mocks.space.proxy_for(object)
+          respond_to = proxy.method_double_if_exists_for_message(:respond_to?)
 
-      # @api private
-      KERNEL_RESPOND_TO_METHOD = ::Kernel.instance_method(:respond_to?)
-
-      if Support::RubyFeatures.supports_rebinding_module_methods?
-        def self.object_responds_to?(object, method_name, visibility)
-          KERNEL_RESPOND_TO_METHOD.bind(object).call(method_name, visibility)
-        end
-      else
-        def self.object_responds_to?(object, method_name, visibility)
-          if ::Kernel === object
-            KERNEL_RESPOND_TO_METHOD.bind(object).call(method_name, visibility)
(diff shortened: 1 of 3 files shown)


JonRowe:
Fixes rspec/rspec-core#2067

lib/rspec/mocks/method_reference.rb
```diff
@@ -94,7 +94,24 @@ def self.method_visibility_for(object, method_name)
...
+          return :public if vis.nil? && object_responds_to?(object, method_name, false)
+        end
+      end
+
+      # @api private
+      KERNEL_RESPOND_TO_METHOD = ::Kernel.instance_method(:respond_to?)
+
+      if Support::RubyFeatures.supports_rebinding_module_methods?
+        def self.object_responds_to?(object, method_name, visibility)
+          KERNEL_RESPOND_TO_METHOD.bind(object).call(method_name, visibility)
```
samphippen: what if the user correctly implements respond to on their object to delegate on to some other object or something? This won't work because the kernel implementation of respond to doesn't understand that implementation

JonRowe: Hm, I guess we need to stash the original `respond_to?` implementation before stubbing it and use that

samphippen: I guess the solution here is to use `RSpec::Support.method_handle_for` to store respond_to in the proxy at instantiation time (before any stub/mock gets applied). The reason to do this at instantiation time is to remove any kind of conditional logic down the pipe. If we always have the original implementation that'll be nice. Also, proxy instantiation time feels right to me because we know at that point there are no existing stubs/mocks on the object. Now to insert a serious question: do we also need to ensure this doesn't happen via any instance as well?

myronmarston: The original method is already available:

https://github.com/rspec/rspec-mocks/blob/abb875d2cc10998d407b07c93d5686675d36b61c/lib/rspec/mocks/method_double.rb#L34-L38

Why not just use that?

Also, we should add specs for the cases discussed here.

alexdowad:
This breaks `class_double`, when used on an object which dynamically defines methods in `method_missing`, the first time they are used. (For the same reasons as already discussed above.)

JonRowe:
@alexdowad yes I'm aware, but I've just finished the refactoring @myronmarston suggested which should avoid that problem

JonRowe:
@myronmarston if this is satisfactory we probably have to write some follow up PRs to move something similar into RSpec support that will still work when mocks isn't available, to leverage in expectations etc.

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -366,7 +366,7 @@ def receiver.method_missing(*a); end # a poor man's stub...
 
               dbl.one
               dbl.two
-            }.to raise_error
+            }.to raise_error RSpec::Mocks::ExpectationNotMet
```
myronmarston: I think this can be `to fail` since we defined a `fail` helper method.

lib/rspec/mocks/proxy.rb
```diff
@@ -228,6 +228,11 @@ def prepended_modules_of_singleton_class
         end
       end
 
+      # @private
+      def method_double_if_exists_for_message(message)
```
myronmarston: Putting `if` in the method name kinda rubs me the wrong way.  WDYT of `existing_method_double_for(message)` instead?

lib/rspec/mocks/method_reference.rb
```diff
@@ -94,7 +94,16 @@ def self.method_visibility_for(object, method_name)
...
+          respond_to = proxy.method_double_if_exists_for_message(:respond_to?)
+
+          if proxy && respond_to
+            object_responds_to_method_name = respond_to.original_method.call(method_name)
+          else
+            object_responds_to_method_name = object.respond_to?(method_name)
+          end
+
+          return :public if vis.nil? && object_responds_to_method_name
         end
```
myronmarston: I think the entire method should be rewritten to not use `tap`:

``` ruby
vis = instance_method_visibility_for(class << object; self; end, method_name)
return vis if vis

# the reset of your logic
```

As you've written it, it wastes time trying to get the proxy, method double and original method but we only need to bother if the return value of `instance_method_visibility_for` is `nil`.  The use of `tap` here makes it kinda confusing, too.

samphippen:
@JonRowe what's the status on this one? Let me know if you need any help getting it over the line.

JonRowe:
@samphippen I believe this is ready, it's certainly not awaiting anything from me and myrons only outstanding point is one of vocab

lib/rspec/mocks/proxy.rb
soulcutter: Even just `method_double_if_exists_for(message)` - there's already a handful of `if` methods in Proxy, so that doesn't both me here too much, but `…for_message(message)` seems redundant

JonRowe:
I'm going to merge this, @myronmarston / @soulcutter if you feel strongly about the wording please open a followup! (or a cleanup commit etc :))

Issue #1020, opened 2015-10-20
Stack blown when conditions for a `respond_to?` method expectation are not met


alexdowad:
Given the following example code:

    expect(object).to receive(:respond_to?).with('method_name')

ANY call to `#respond_to?` which does not meet the expectation results in a `SystemStackError`, rather than the expected failure message.

This is because of `ErrorGenerator#default_error_message`, which calls `#join` on an array which contains the proxied object.

`Array#join` is implemented by `rb_ary_join` at `array.c:2023`, in the MRI source. That function contains this line:

    tmp = rb_check_string_type(val);

Which goes to `string.c:1771`:

    VALUE
    rb_check_string_type(VALUE str)
    {
        str = rb_check_convert_type(str, T_STRING, "String", "to_str");
        return str;
    }

That goes through `rb_check_convert_type`, to `convert_type`, to `rb_check_funcall`, to `check_funcall_respond_to`. At `vm_eval.c:399`, we reach:

    result = vm_call0(th, recv, idRespond_to, arity, args, me, defined_class);

And the fun starts all over again! RSpec's proxy object checks whether the call to `respond_to?` meets its expectations, it doesn't, so it tries to generate an error message, calls `Array#join` in the process, then while building the joined string MRI calls `#respond_to?("to_str")`, so the proxy object checks its expectations...

Putting expectations on any method which MRI calls internally is a tricky business!

One (partial) solution, which would make the situation much better: *while processing a failed expectation*, temporarily disable all method expectations on the object whose expectation was failed! (Better yet, "un-proxy" all methods on that object and let calls go through to the original implementation!)

Once you are finished processing the failed expectation and are ready to return back to the user's spec code, re-enable those method expectations!

alexdowad:
I just came up with something else which makes this particular situation much better... though the potential is still there for RSpec to blow the stack when method expectations are put on methods which RSpec and/or MRI call internally.

Will send a PR promptly.

JonRowe:
See #1013

myronmarston:
Honestly, I think the solution is to not use `join` in `default_error_message`.  Instead just define a single string with interpolation.  `join` is convenient there but string interpolation should work better, anyway -- the `join` version of the method creates an extra array (an unnecessary object allocation) and does the extra `join` method send plus it causes the problem you've mentioned.

Want to take a stab at a PR addressing that, @alexdowad? Be sure it includes a failing test for this case as well.

alexdowad:
[...]

PR #1022, opened 2015-10-20
Avoid stack overflow when #respond_to? expectation fails

alexdowad:
As per discussion in issue #1020.

(I slipped a spelling fix in as an extra commit...)


JonRowe:
Thanks

PR #1024, opened 2015-10-20
Supress warnings from not specifying error

JonRowe:
Pulled this out to it's own PR because it's simple, and used your suggestion @myronmarston


myronmarston:
LGTM

PR #1026, opened 2015-10-23
Fix problem with 'receive' on MRI 1.8.7 (when used on singleton methods in superclass and subclass)

alexdowad:
The following code is broken on MRI 1.8.7:

```ruby
class A; def self.foo; "foo"; end end
class B < A; end

allow(A).to receive(:foo) { ... }
allow(B).to receive(:foo) { ... }
```

The problem is that `Proxy#original_method_handle_for` gets the `UnboundMethod` which is stashed in the superclass' proxy, and tries to rebind it to the subclass. This doesn't work in 1.8.7.

Can you suggest a cleaner way to implement the workaround?

### DIFF
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -371,20 +371,36 @@ module RSpec
 
         return super unless unbound_method
         unbound_method.bind(object)
+        # :nocov:
+      rescue TypeError
+        if RUBY_VERSION == '1.8.7'
+          # In MRI 1.8.7, a singleton method on a class cannot be rebound to its subclass
+          if unbound_method && unbound_method.owner.ancestors.first != unbound_method.owner
+            # This is a singleton method; we can't do anything with it
+            # But we can work around this using a different implementation
+            double = method_double_from_ancestor_for(message)
+            return object.method(double.method_stasher.stashed_method_name)
+          end
+        end
+        raise
+        # :nocov:
       end
 
     protected
 
       def original_unbound_method_handle_from_ancestor_for(message)
-        method_double = @method_doubles.fetch(message) do
+        double = method_double_from_ancestor_for(message)
+        double && double.original_method.unbind
+      end
+
+      def method_double_from_ancestor_for(message)
+        @method_doubles.fetch(message) do
           # The fact that there is no method double for this message indicates
           # that it has not been redefined by rspec-mocks. We need to continue
           # looking up the ancestor chain.
           return superclass_proxy &&
-            superclass_proxy.original_unbound_method_handle_from_ancestor_for(message)
+            superclass_proxy.method_double_from_ancestor_for(message)
         end
-
-        method_double.original_method.unbind
       end
 
       def superclass_proxy
--- a/spec/rspec/mocks/and_call_original_spec.rb
+++ b/spec/rspec/mocks/and_call_original_spec.rb
@@ -30,7 +30,7 @@ RSpec.describe "and_call_original" do
       end
     end
 
-    context "when a non-existant method has been stubbed previously" do
+    context "when a non-existent method has been stubbed previously" do
       it 'restores the original NameError behavior' do
         expect { instance.abcd }.to raise_error(NameError).with_message(/abcd/)
 
@@ -249,7 +249,7 @@ RSpec.describe "and_call_original" do
 
     let(:request) { request_klass.new(:get, "http://foo.com/bar") }
 
-    it 'still works even though #method has been overriden' do
+    it 'still works even though #method has been overridden' do
       expect(request).to receive(:perform).and_call_original
       expect(request.perform).to eq(:the_response)
     end
--- a/spec/rspec/mocks/and_call_original_spec.rb
+++ b/spec/rspec/mocks/and_call_original_spec.rb
@@ -143,6 +143,19 @@ RSpec.describe "and_call_original" do
       expect(subclass.new_instance).to be_a(subclass)
     end
 
+    context 'when a class method is stubbed in the superclass' do
+      it 'still works for class methods defined on a superclass' do
+        superclass = Class.new { def self.foo; "foo"; end }
+        subclass   = Class.new(superclass)
+
+        allow(superclass).to receive(:foo).and_return(:fake)
+        expect(subclass).to receive(:foo).and_call_original
+
+        expect(superclass.foo).to be :fake
+        expect(subclass.foo).to eq "foo"
+      end
+    end
+
     it 'works for class methods defined on a grandparent class' do
       sub_subclass = Class.new(Class.new(klass))
       expect(sub_subclass).to receive(:new_instance).and_call_original
--- a/spec/rspec/mocks/matchers/receive_spec.rb
+++ b/spec/rspec/mocks/matchers/receive_spec.rb
@@ -267,6 +267,44 @@ module RSpec
         it_behaves_like "resets partial mocks cleanly" do
           let(:target) { allow(object) }
         end
+
+        context 'on a class method, from a class with subclasses' do
+          let(:superclass)     { Class.new { def self.foo; "foo"; end }}
+          let(:subclass_redef) { Class.new(superclass) { def self.foo; ".foo."; end }}
+          let(:subclass_deleg) { Class.new(superclass) { def self.foo; super.upcase; end }}
+          let(:subclass_asis)  { Class.new(superclass) }
+
+          context 'if the method is redefined in the subclass' do
+            it 'does not stub the method in the subclass' do
+              allow(superclass).to receive(:foo) { "foo!!" }
+              expect(superclass.foo).to eq "foo!!"
+              expect(subclass_redef.foo).to eq ".foo."
+            end
+          end
+
+          context 'if the method is not redefined in the subclass' do
+            it 'stubs the method in the subclass' do
+              allow(superclass).to receive(:foo) { "foo!!" }
+              expect(superclass.foo).to eq "foo!!"
+              expect(subclass_asis.foo).to eq "foo!!"
+            end
+          end
+
+          it 'creates stub which can be called using `super` in a subclass' do
+            allow(superclass).to receive(:foo) { "foo!!" }
+            expect(subclass_deleg.foo).to eq "FOO!!"
+          end
+
+          it 'can stub the same method simultaneously in the superclass and subclasses' do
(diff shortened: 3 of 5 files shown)


alexdowad:
On MRI 2.2.0, one of my added specs caused a warning about "possible reference to past scope". I amended the spec to avoid this warning.

lib/rspec/mocks/proxy.rb
```diff
@@ -371,20 +371,34 @@ def original_method_handle_for(message)
 
         return super unless unbound_method
         unbound_method.bind(object)
+      rescue TypeError
+        if RUBY_VERSION == '1.8.7'
```
myronmarston: Instead of checking the ruby version at runtime each time this occurs, we prefer to check the ruby version once at file load time, and conditionally define different implementations of methods based on the version.  For example, you could do something like:

``` ruby
if RUBY_VERSION == '1.8.7'
  def handle_bind_type_error
    yield
  rescue TypeError
    # your logic...
  end
else
  def handle_bind_type_error
    yield
  end
end
```

Then `original_method_handle_for` can wrap the `bind` call in `handle_bind_type_error { ... }` and the appropriate thing will happen, without the cost of checking the ruby version or rescuing and re-raising.

There may be a simpler, better way to organize this, but hopefully that gives you an idea.

spec/rspec/mocks/and_call_original_spec.rb
```diff
@@ -30,7 +30,7 @@ def self.new_instance
       end
     end
 
-    context "when a non-existant method has been stubbed previously" do
+    context "when a non-existent method has been stubbed previously" do
```
myronmarston: Thanks for fixing our silly spelling mistakes!

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -267,6 +267,43 @@ def receiver.method_missing(*a); end # a poor man's stub...
         it_behaves_like "resets partial mocks cleanly" do
           let(:target) { allow(object) }
         end
+
+        context 'on a class method, from a class with subclasses' do
+          let(:superclass) { Class.new { def self.foo; "foo"; end }}
+          let(:subclass1)  { Class.new(superclass) { def self.foo; super.upcase; end }}
+          let(:subclass2)  { Class.new(superclass) }
```
myronmarston: I think the test would be easier to follow if `subclass1` and `subclass2` were better named, e.g. `upcase_subclass` and `no_override_subclass` or similar.

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -267,6 +267,43 @@ def receiver.method_missing(*a); end # a poor man's stub...
...
+
+          context 'if the method is redefined in the subclass' do
+            it 'does not stub the method in the subclass' do
+              allow(superclass).to receive(:foo) { "foo!!" }
+              expect(superclass.foo).to eq "foo!!"
+              expect(subclass1.foo).not_to eq "foo!!"
```
myronmarston: It seems odd to claim what `subclass1.foo` does not equal (and actually, I originally read it as `to eq`, and got really confused...).  Mind restating it positively?

myronmarston: Oh nevermind, I see that you stated the positive in a spec below. Ignore my earlier comment.

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -267,6 +267,43 @@ def receiver.method_missing(*a); end # a poor man's stub...
...
+              expect(subclass2.foo).to eq "foo!!"
+            end
+          end
+
+          it 'creates stub which can be called using `super` in a subclass' do
+            allow(superclass).to receive(:foo) { "foo!!" }
+            expect(subclass1.foo).to eq "FOO!!"
+          end
+
+          it 'can be used simultaneously in the superclass and subclasses' do
```
myronmarston: What does "used simultaneously" mean here?  Do you mind "stubbed simultaneously"?

lib/rspec/mocks/proxy.rb
```diff
@@ -371,20 +371,34 @@ def original_method_handle_for(message)
...
+          # In MRI 1.8.7, a singleton method on a class cannot be rebound to its subclass
+          if unbound_method && unbound_method.owner.ancestors.first != unbound_method.owner
+            # This is a singleton method; we can't do anything with it
+            # But we can work around this using a different implementation
+            double = method_double_from_ancestor_for(message)
+            return object.method(double.method_stasher.send(:stashed_method_name))
```
myronmarston: I'd rather we not violate the privacy via `send`.  If this method needs to be public.

More generally: this logic seems to be all about the method double, which makes me wonder if we should expose some kind of method there and then delegate to it from here.

myronmarston:
> Can you suggest a cleaner way to implement the workaround?

Sadly, no.  This stuff gets waaaaay too complicated, particularly when on 1.8.7 behavior is thrown into the mix.

myronmarston:
Thanks, @alexdowad, this is looking really good!

alexdowad:
Thanks for the **thorough** code review! Will address everything point-by-point now.

lib/rspec/mocks/proxy.rb
alexdowad: If that's what you want me to do, I will do it, but frankly I think the existing code is better.

Notice that the conditional is in an exception handler, out of the main flow of the method. On post-1.8.7 interpreters, the exception handler should never be hit. Even on MRI 1.8.7, in most cases it will never be hit. So in all those cases, the performance impact should be zero. There's just this one corner case involving stubbing of singleton methods on classes, when it will actually be used.

If we extract a new method, then the performance impact will not be zero. It will have a (small) performance impact on all invocations, on all platforms.

Additionally, the code is a bit more compact this way, and there is less indirection. Indirection is a double-edged sword; it can make your code better, or it can make it worse. I feel a bit less indirection is better in this case. But it's up to you.

spec/rspec/mocks/matchers/receive_spec.rb
alexdowad: Check!

lib/rspec/mocks/proxy.rb
myronmarston: That's a good point.  It's the general pattern we use but when there's a reason not to follow it (as you've explained well) then it's fine not to -- so let's stick with what you have.

Can you add `# :nocov:` before and after this section so that it isn't considered for simplecov code coverage?

spec/rspec/mocks/matchers/receive_spec.rb
alexdowad: Adjusted this anyways.

spec/rspec/mocks/matchers/receive_spec.rb
alexdowad: Check!

lib/rspec/mocks/proxy.rb
alexdowad: Working on this now.

alexdowad: The tricky thing about adding a new method to the method double is that `stashed_method_name` only exists when you are running on Ruby 1.8! So do we conditionally define another method on `MethodDouble` which also exists only on 1.8? Seems yucky.

alexdowad:
PR updated.

alexdowad:
Updated PR again with `:nocov:` comments, hope I got them right.

lib/rspec/mocks/proxy.rb
```diff
@@ -371,20 +371,36 @@ def original_method_handle_for(message)
...
+          # In MRI 1.8.7, a singleton method on a class cannot be rebound to its subclass
+          if unbound_method && unbound_method.owner.ancestors.first != unbound_method.owner
+            # This is a singleton method; we can't do anything with it
+            # But we can work around this using a different implementation
+            double = method_double_from_ancestor_for(message)
+            return object.method(double.method_stasher.stashed_method_name)
+          end
+        end
+        raise
+      # :nocov:
```
myronmarston: Rubocop is complaining about the indentation of this -- it should be over two spaces since it's within the `rescue/end`.

spec/rspec/mocks/and_call_original_spec.rb
```diff
@@ -143,6 +143,19 @@ def inst.foo; :bar; end
       expect(subclass.new_instance).to be_a(subclass)
     end
 
+    context 'when a class method is stubbed in the superclass' do
+      it 'still works for class methods defined on a superclass' do
```
myronmarston: I find this doc string confusing -- it seems to just restate the context (e.g. "class method on a superclass").  Isn't the point of the spec that a subclass can still use `and_call_original` after it's superclass has been stubbed?

myronmarston:
This looks good, @alexdowad.  I'll be happy to merge once it's green (just need the indentation fixed to satisfy rubocop).

JonRowe:
Be good to squash this as well :)

myronmarston:
> Be good to squash this as well :)

I disagree -- if you look at the commits commit-by-commit you'll see that @alexdowad was intentional about dividing this into logical chunks and writing up _excellent_ commit messages.  Squashing is nice when the later commits are just fixups but that's not the case here.

JonRowe:
I did look, but I disagree, I feel the spec file should be in the same commit as the fix, to aid future bisects.

lib/rspec/mocks/proxy.rb
alexdowad: I just ran Rubocop on `rspec-mocks` and it comes up with over 700 warnings and errors for the whole codebase. Do you have some custom settings for Rubocop which you use to enforce your desired coding style?

JonRowe: Yes, see the Rake task.

spec/rspec/mocks/and_call_original_spec.rb
alexdowad: We have 2 different specs here:

- Class method defined in superclass, inherited by subclass, stubbed in subclass
- Class method defined in superclass, inherited by subclass, *stubbed in superclass*, stubbed separately in subclass

Any suggestions on a better docstring to convey the difference between these 2 examples?

alexdowad:
I'm not sure how you are running RuboCop -- when I run it on that file, it doesn't find any style problems. Anyways, I am just fixing the indentation and squashing.

JonRowe:
:heart:

Issue #1041, opened 2015-12-15
Ruby 2.3.0-preview2 generates ruby warnings and fails many stub/double tests -


jrafanie:
I'm seeing issues with ruby 2.3.0-preview2 when rspec stub/doubles are involved where the wrong super method is being invoked.  This is probably [too much information but here it is](https://github.com/ManageIQ/manageiq/issues/5814) in case you want to how I got to this issue.

Because of that failure, I tried the rspec-mocks test suite with both 2.2.3 and 2.3.0-preview2 and noticed ruby warnings and test failures in 2.3.0 that don't occur on 2.2.3.

The warnings look like: 
`/Users/joerafaniello/Code/rspec-mocks/spec/rspec/mocks/double_spec.rb:254: warning: previous definition of method_with_default_argument was here`

I have the full output of  `bundle exec rake spec` from [both 2.2.3 and 2.3.0-preview2 here](https://gist.github.com/jrafanie/570aa8a80f78a2b51b61).  They indicate a problem with stubbing/mocking methods in 2.3.0-preview2.

I looked through the release notes and [this section looked very interesting](https://github.com/ruby/ruby/blob/v2_3_0_preview2/ChangeLog#L5720-L5834).

It references this bug: https://bugs.ruby-lang.org/issues/11278
Also, https://bugs.ruby-lang.org/issues/11279, which says "This is big change of method data structure."

It feels like this might be a ruby regression but I can't limit the scope of the warning outside of rspec yet.

jrafanie:
I just noticed the ruby-head tests are also failing with the same errors:
![screen shot 2015-12-15 at 6 18 21 pm](https://cloud.githubusercontent.com/assets/19339/11827181/514630f0-a358-11e5-9fa4-bf8e7a64e498.png)

See: https://travis-ci.org/rspec/rspec-mocks/jobs/95425009

jrafanie:
cc @JonRowe @myronmarston (I didn't see this reported elsewhere)

JonRowe:
Thanks for reporting this, we haven't tried `2.3.0` ourselves yet, it's worth noting that this is fairly typical of a Ruby preview, and the final version may work as before.

jrafanie:
I'm glad to help @JonRowe.  Thanks.  I didn't see a bug reported in https://bugs.ruby-lang.org/projects/ruby-trunk/issues?set_filter=1&tracker_id=1 that resembled the symptoms I was seeing. Please let me know if I can help in any way.

[...]

PR #1043, opened 2015-12-18
Only restore an instance method on the owning class

### LINKED ISSUE #1042: RSpec mocks isn't properly rolling back stubbed any_instance methods with ruby 2.3.0-preview2
The gist below causes randoms errors depending on the order of the tests but shows that with either rspec 2 or 3 ruby 2.3.0-preview2 randomly fails because the super method is wrong in a different test after a prior test stubbed the method with any_instance.

https://gist.github.com/jrafanie/c94e4c8a4f940e71c9e6

The gist passes with ruby 2.2.3 every time I have run it.

I think the problem is somewhere in the recorder but I have no idea what to try next.
https://github.com/rspec/rspec-mocks/blob/master/lib/rspec/mocks/any_instance/recorder.rb

jrafanie:
Fixes #1042

```ruby
class Shape
  def area(value)
    value
  end
end

class Circle < Shape
  def area
    super(2)
  end
end

class SmallCircle < Circle
end
```

When using: allow_any_instance_of(SmallCircle).to receive(:area)

Previously, we'd copy the area method from the first superclass
that implements it.  In this case, we'd copy Circle#area into
SmallCircle with a backup name.

At the end of the example, we'd remove the stubbing and rename the
copied method back, causing SmallCircle to implement area when it never
did previously.

Any calls to SmallCircle#area would end up calling the contents of
Circle#area and the call to super(2) would be sent to the wrong class:
Circle, not Shape.  In this case, Circle#area doesn't take any
arguments so you get an ArgumentError.

Previously, ruby 2.2 and below allowed this strange behavior to work.

This appears to have been "fixed" in ruby in this commit:
https://github.com/ruby/ruby/commit/c8854d2ca4be9ee6946e6d17b0e17d9ef130ee81

### DIFF
--- a/lib/rspec/mocks/any_instance/recorder.rb
+++ b/lib/rspec/mocks/any_instance/recorder.rb
@@ -18,6 +18,7 @@ module RSpec
           @stubs = Hash.new { |hash, key| hash[key] = [] }
           @observed_methods = []
           @played_methods = {}
+          @backed_up_method_owner = {}
           @klass = klass
           @expectation_set = false
         end
@@ -190,9 +191,9 @@ module RSpec
           return unless @klass.instance_method(method_name).owner == @klass
 
           alias_method_name = build_alias_method_name(method_name)
-          @klass.class_exec do
+          @klass.class_exec(@backed_up_method_owner) do |backed_up_method_owner|
             remove_method method_name
-            alias_method method_name, alias_method_name
+            alias_method method_name, alias_method_name if backed_up_method_owner[method_name.to_sym] == self
             remove_method alias_method_name
           end
         end
@@ -204,10 +205,13 @@ module RSpec
         end
 
         def backup_method!(method_name)
+          return unless public_protected_or_private_method_defined?(method_name)
+
           alias_method_name = build_alias_method_name(method_name)
+          @backed_up_method_owner[method_name.to_sym] ||= @klass.instance_method(method_name).owner
           @klass.class_exec do
             alias_method alias_method_name, method_name
-          end if public_protected_or_private_method_defined?(method_name)
+          end
         end
 
         def public_protected_or_private_method_defined?(method_name)
--- a/spec/rspec/mocks/any_instance_spec.rb
+++ b/spec/rspec/mocks/any_instance_spec.rb
@@ -1,5 +1,22 @@
 require 'delegate'
 
+module AnyInstanceSpec
+  class GrandparentClass
+    def foo(_a)
+      'bar'
+    end
+  end
+
+  class ParentClass < GrandparentClass
+    def foo
+      super(:a)
+    end
+  end
+
+  class ChildClass < ParentClass
+  end
+end
+
 module RSpec
   module Mocks
     RSpec.describe "#any_instance" do
@@ -1223,6 +1240,14 @@ module RSpec
           verify_all
 
           expect(instance.existing_method).to eq :existing_method_return_value
+        end
+
+       it "does not restore a stubbed method not originally implemented in the class" do
+          allow_any_instance_of(::AnyInstanceSpec::ChildClass).to receive(:foo).and_return :result
+          expect(::AnyInstanceSpec::ChildClass.new.foo).to eq :result
+
+          reset_all
+          expect(::AnyInstanceSpec::ChildClass.new.foo).to eq 'bar'
         end
 
         it "restores the original behaviour, even if the expectation fails" do


jrafanie:
Note, I had to locally hack rspec-support to get around the warnings from #1041.

With the hack below, I was able to test ruby 1.9.3, 2.0.0, 2.1.7, 2.2.3 and 2.3.0-preview2:

```diff
diff --git a/lib/rspec/support/spec/stderr_splitter.rb b/lib/rspec/support/spec/stderr_splitter.rb
index f797a23..bd92e54 100644
--- a/lib/rspec/support/spec/stderr_splitter.rb
+++ b/lib/rspec/support/spec/stderr_splitter.rb
@@ -43,7 +43,16 @@ module RSpec
       end

       def has_output?
-        !output.empty?
+        output.lines.any? do |line|
+          if line.include?("warning: method redefined")
+            false
+          elsif line.include?("warning: previous definition")
+            false
+          else
+            !line.empty?
+          end
+        end
+
       end

```

spec/rspec/mocks/any_instance_spec.rb
```diff
@@ -1213,6 +1230,17 @@ def dup(_)
         end
       end
 
+      context "on a grandchild class" do
+        it "stubs a method not implemented in grandchild" do
+          allow_any_instance_of(::AnyInstanceSpec::ChildClass).to receive(:foo).and_return :result
+          expect(::AnyInstanceSpec::ChildClass.new.foo).to eq :result
```
jrafanie: These tests would not fail with anonymous classes!  I had to pollute the test suite with some test constants so I wrapped them in a module to avoid a collision.  I'm open to doing this differently.

I originally had something like this but it wouldn't reproduce the ArgumentError.

```diff
diff --git a/spec/rspec/mocks/any_instance_spec.rb b/spec/rspec/mocks/any_instance_spec.rb
index f2375cf..e33b8b1 100644
--- a/spec/rspec/mocks/any_instance_spec.rb
+++ b/spec/rspec/mocks/any_instance_spec.rb
@@ -31,6 +31,12 @@ module RSpec
           def private_method; :private_method_return_value; end
         end
       end
+
+      # TODO: Remove me... anonymous classes can't reproduce #1042
+      let(:grandparent_class) { Class.new                    { def foo(_a); 'bar'; end }}
+      let(:parent_class)      { Class.new(grandparent_class) { def foo; super(:a); end }}
+      let(:child_class)       { Class.new(parent_class)      {}}
+
       let(:existing_method_return_value) { :existing_method_return_value }
```

jrafanie:
Closing/re-opening in hopes the nokogiri build failures in travis won't happen again.

jrafanie:
Re-pushed to fix style `map over collect` preference

jrafanie:
@myronmarston @JonRowe @I fixed the rubocop error but now it's failing due to nokogiri build failures... It looks like this one:  https://github.com/sparklemotion/nokogiri/pull/1393  Let me know if I can help any further to get this in.

myronmarston:
Thanks for taking a stab at this, @jrafanie.  I'm a bit concerned about the performance implications of your solution here -- mapping over all methods (e.g. `map(&:to_sym)`) is a linear operation that we'd like to avoid (particularly because the sorts of legacy systems where users use `any_instance` tend to be ones where there objects have very large interfaces).  Likewise, the `include?` check is a linear operation as well.

Did an `owner` check not work?  That route seems simpler and likely to perform better.

Don't worry about the nokogiri install failure for now.  @JonRowe was working on that recently and I hope he has some ideas here :).

jrafanie:
@myronmarston I agree, I'm not a fan of the `include?` especially.

I'll try to see if I can get the "don't backup this method if I don't own it" logic to work.  When I tried it originally, it seemed like the rest of the recorder blew up in the tests.

lib/rspec/mocks/any_instance/recorder.rb
```diff
@@ -18,6 +18,9 @@ def initialize(klass)
           @stubs = Hash.new { |hash, key| hash[key] = [] }
           @observed_methods = []
           @played_methods = {}
+          @own_methods = (klass.public_instance_methods(false) +
+                         klass.protected_instance_methods(false) +
+                         klass.private_instance_methods(false)).map(&:to_sym)
```
jrafanie: Since a force push will wipe this out... putting @myronmarston's comments here so the "outdated" diff will still match up with his comments. 

> Thanks for taking a stab at this, @jrafanie. I'm a bit concerned about the performance implications of your solution here -- mapping over all methods (e.g. map(&:to_sym)) is a linear operation that we'd like to avoid (particularly because the sorts of legacy systems where users use any_instance tend to be ones where there objects have very large interfaces). Likewise, the include? check is a linear operation as well.
Did an owner check not work? That route seems simpler and likely to perform better.

lib/rspec/mocks/any_instance/recorder.rb
```diff
@@ -190,9 +193,9 @@ def restore_original_method!(method_name)
           return unless @klass.instance_method(method_name).owner == @klass
 
           alias_method_name = build_alias_method_name(method_name)
-          @klass.class_exec do
+          @klass.class_exec(@own_methods) do |own_methods|
             remove_method method_name
-            alias_method method_name, alias_method_name
+            alias_method method_name, alias_method_name if own_methods.include?(method_name.to_sym)
```
jrafanie: And also, this one will be lost on a force push...

> Thanks for taking a stab at this, @jrafanie. I'm a bit concerned about the performance implications of your solution here -- mapping over all methods (e.g. map(&:to_sym)) is a linear operation that we'd like to avoid (particularly because the sorts of legacy systems where users use any_instance tend to be ones where there objects have very large interfaces). Likewise, the include? check is a linear operation as well.
Did an owner check not work? That route seems simpler and likely to perform better.

jrafanie:
Ok @myronmarston, thanks for reviewing this change... I have pushed a change to capture the method ownership before we backup the method (once it's copied to the `@klass`, it's too late) and use this original ownership on the restoration of the method for only the "owning class".  It feels like this backup/restore functionality should be encapsulated in an object, not a hash, but I felt that was out of the scope of this fix.

jrafanie:
> Did an owner check not work? That route seems simpler and likely to perform better.

@myronmarston Yes, the owner check is only possible if you have the original method object before we alias (copy) it to the `@klass`.  Due to your comment, I went back to that mechanism and found a way to get it to work.

lib/rspec/mocks/any_instance/recorder.rb
```diff
@@ -190,9 +191,9 @@ def restore_original_method!(method_name)
           return unless @klass.instance_method(method_name).owner == @klass
 
           alias_method_name = build_alias_method_name(method_name)
-          @klass.class_exec do
+          @klass.class_exec(@backed_up_method_owner) do |backed_up_method_owner|
             remove_method method_name
-            alias_method method_name, alias_method_name
+            alias_method method_name, alias_method_name if backed_up_method_owner[method_name.to_sym] && backed_up_method_owner[method_name.to_sym] == self
```
myronmarston: The first part of the `if` conditional seems unnecessary.  It only plays a role if it evaluates to `false` and that can only happen if it returns `false` or `nil`.  The second part checks the return value against `self`, which I don't believe can be `false` or `nil` -- so I think the first part can be removed entirely.  Does that sound right?

spec/rspec/mocks/any_instance_spec.rb
```diff
@@ -1213,6 +1230,17 @@ def dup(_)
...
+        end
+
+        it "stubbed method doesn't get restored to grandchild" do
```
myronmarston: It's nice to phrase the test descriptions to read gramatically well off of "it", and "it stubbed method..." does not read well.  How about `it "does not restore stubbed method to grandchild"`?

myronmarston: Also, this test has a slight problem.  It only tests that the method is not restored if it runs _after_ the test above it, but since we use random ordering the order is not guaranteed (and we don't want order-dependent tests, anyway!).

Instead it should be written to do the any instance stubbing, restore things, and then make the expectation.  You can use [`reset_all`](https://github.com/rspec/rspec-mocks/blob/b94e242ebbc51081fc2245e37c574f5d6d02b7f8/spec/spec_helper.rb#L39-L41) to restore things.

myronmarston:
Thanks, @jrafanie.  This solution looks a lot better.  Now we just have to get the PR green (which will hopefully happen when one of us fixes the nokogiri install issue).

lib/rspec/mocks/any_instance/recorder.rb
jrafanie: Yes, thanks.

spec/rspec/mocks/any_instance_spec.rb
jrafanie: Thanks, I wasn't sure how to guarantee it.  I'll try out your suggestion.

spec/rspec/mocks/any_instance_spec.rb
```diff
@@ -1225,6 +1242,14 @@ def dup(_)
           expect(instance.existing_method).to eq :existing_method_return_value
         end
 
+       it "does not restore a stubbed method not originally implemented in the class" do
+          allow_any_instance_of(::AnyInstanceSpec::ChildClass).to receive(:foo).and_return :result
+          expect(::AnyInstanceSpec::ChildClass.new.foo).to eq :result
+
+          reset_all
```
jrafanie: Thanks @myronmarston, `reset_all` is what I needed... I was able to drop the extra example and guarantee the test failure on ruby 2.3 without the `recorder.rb` changes.  :+1: 

I moved the example to an existing context `...after the test has finished` that seems to be describe what we're doing here.  The test output looks much better now, thanks!

```
  when used after the test has finished
    restores the original behaviour, even if the expectation fails
    does not restore a stubbed method not originally implemented in the class
    restores the original behavior of a stubbed method
```

myronmarston: :+1:

deivid-rodriguez:
Hi! Since I worked on this same build failure I thought I may post my findings. This should get fixed once my PR sparklemotion/nokogiri#1393 get merged or once you remove `--disable=gem` from your workflow.

JonRowe:
If you rebase this it should be good to go...

jrafanie:
Thank you @deivid-rodriguez and @JonRowe.

@JonRowe This branch was already rebased.  I wasn't sure if you wanted me to run the build again, in case you a dependency was updated... so I amended my commit and force pushed to restart the travis build.

Let me know if there's something else you would like.

Thanks!

jrafanie:
@JonRowe i rebased this PR and it's still failing for the `gem install nokogiri` issue @deivid-rodriguez references. Please advise. 2.3.0 was released a few days ago so I'd like to help in any way to get this in and released. Thanks for the help!

jrafanie:
@myronmarston @JonRowe this PR continues to fail with the undefined method gem error in nokogiri's extconf.rb.  If you want I can rebase this PR after #1049 is merged since it appears it will [fix that error for now using the default require rubygems behavior of ruby](https://github.com/rspec/rspec-mocks/pull/1049/files#diff-df46e7fa9b07f655ea05016791e8acb1L17).  Thanks!

myronmarston:
Hey @jrafanie -- sorry for the delay.  I looked into this last week but as you see, we were still having nokogiri install problems.  I believe those have been sorted now, so I think if you rebase on top of master (now that #1049 has been merged) it should work.  Thanks for following up!

jrafanie:
Thanks @myronmarston, no problem.  We're :green_heart:  now.

myronmarston:
Gladly merged!  Thanks again, @jrafanie!

jrafanie:
Thanks @myronmarston @JonRowe @deivid-rodriguez @tenderlove for the help :heart:

PR #1052, opened 2016-01-06
Updates from rspec-dev (2016-01-06)

myronmarston:


JonRowe:
I think this is missing a 2.3.0 fix... Hopefully this will now go green...

JonRowe:
Merged despite failure as I think this will allow all the other builds to go green now

PR #1053, opened 2016-01-11
fix typo

paddor:


myronmarston:
Thanks! (merged in spite of the build failure because its a transient jruby failure we see occaionally and haven't yet solved)

PR #1054, opened 2016-01-14
Chain and_yields for any_instance

### LINKED ISSUE #1025: allow_any_instance_of().to_receive cannot yield multiple times
When using allow_any_instance_of() instead of allow(), chaining multiple and_yield()s throws an error.  So either of these work fine:

    allow(Klass).to_receive(:method).and_yield(1).and_yield(2)
    allow_any_instance_of(Klass).to_receive(:method).and_yield(1)

But attempting the following:

    allow_any_instance_of(Klass).to_receive(:method).and_yield(1).and_yield(2)

Results in a `NoMethodError: Undefined method and_yield`.  Using rspec-mocks 3.3.2.

kiliancs:
This fixes #1025.
```
allow_any_instance_of(Klass).to receive(:method).and_yield(1).and_yield(2)
```
now behaves as expected, instead of raising `NoMethodError: Undefined method and_yield`.

### DIFF
--- a/lib/rspec/mocks/any_instance/stub_chain.rb
+++ b/lib/rspec/mocks/any_instance/stub_chain.rb
@@ -30,7 +30,7 @@ module RSpec
             :with => [nil],
             :and_return => [:with, nil],
             :and_raise => [:with, nil],
-            :and_yield => [:with, nil],
+            :and_yield => [:with, :and_yield, nil],
             :and_call_original => [:with, nil],
             :and_wrap_original => [:with, nil]
           }
--- a/spec/rspec/mocks/any_instance_spec.rb
+++ b/spec/rspec/mocks/any_instance_spec.rb
@@ -33,6 +33,13 @@ module RSpec
       end
       let(:existing_method_return_value) { :existing_method_return_value }
 
+      context "chain" do
+        it "yields the value specified" do
+          allow_any_instance_of(klass).to receive(:foo).and_yield(1).and_yield(2)
+          expect { |b| klass.new.foo(&b) }.to yield_successive_args(1, 2)
+        end
+      end
+
       context "invocation order" do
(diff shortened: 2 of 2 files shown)


lib/rspec/mocks/any_instance/stub_chain.rb
```diff
@@ -30,7 +30,7 @@ def invocation_order
             :with => [nil],
             :and_return => [:with, nil],
             :and_raise => [:with, nil],
-            :and_yield => [:with, nil],
+            :and_yield => [:with, :and_yield, nil],
```
JonRowe: I'm not sure this is the way to fix this, other combinations would cause the same error as before, we need to return `self` from all the method calls that support further chaining I think.

kiliancs: Makes sense. Which are the method calls that support further chaining? All except `with`?

JonRowe: `with` is first and so always supports further chaining, `and_yield` also supports (I think) `and_return` and `and_raise` after it, I'm sure `and_call_original` and `and_wrap_original` should have some combinations too

kiliancs: Would it be safe to say that `with` always comes first, then any combination of the other methods?

JonRowe: Not quite, as `and_return`,  `and_raise`, and `and_call_original` or `and_wrap_original` should all be mutually exclusive I think.

JonRowe:
I've decided to manually merge this (I squashed some commits), if theres any follow up problems with other combinations another PR can always be opened :)

JonRowe:
Thanks for your help! :)

Issue #1055, opened 2016-01-14
Too strong restriction for matchers


nepalez:
Seems like the requirement `only the 'receive', 'have_received' and 'receive_messages' matchers are supported with 'allow(...).to'` is too restrictive.

For example, I provide the matcher, that is sort of [semi-transparent extension for the `receive`](https://gist.github.com/nepalez/ac049b788c369e8d98a3#file-receive_request-rb-L59-L65)

I believe it could work, but it doesn't because of the restriction above. Do we really need it?

Surely, I could just reload the `receive` method (and actually I did) and then monkey-patch the whole `RSpec::Mocks::Matchers::Receive` class. But also I do need to reload both the `allow` and `expect` to [switch `client` to `client.connection`](https://gist.github.com/nepalez/ac049b788c369e8d98a3#file-receive_request-rb-L64). It works, but seems far too messy, not to mention dark sides of monkey patching by itself.

I think, it would be far better way to allow developer to shoot himself :gun:  in the foot, not restricting him so strongly.

nepalez:
Actually I understood that a proposed solution is not sufficient. What I do need is a way to extend receive matcher to make it accessible to both `expect` and `allow` (my solution doesn't do this)

myronmarston:
Exactly.  Matchers in general don't work with `allow`.  `allow` doesn't even use the matcher protocol, because it has to act differently than `expect` does.  It actually calls `setup_allowance` on the matcher, but that's not a published public API.

I think you could get `allow(client).to receive_request(:get, url)` to work if you define `receive_request` as a method that delegates to `receive` internally -- which would just return the same kind of object that `receive` returns which `allow` would understand.

nepalez:
Hi, @myronmarston, and thank you for the suggestion!

Yes, this is [what I've done eventually](https://gist.github.com/nepalez/7a33e39d5fc3162c0ba2#file-receive_request_1-rb-L46-L51).

Though this works, still there is one more issue. Seems like the restriction requires exact classes (`RSpec::Mocks::Matchers::Receive` and so on) for the allowance matchers. If it **allowed subclassing too**, I could remove all this mess from the `#receive_request` and just define my own receiver, and do something like this:

```ruby
module Receiver
  def initialize(method, path = nil)
    # ...
    super # ...
  end

  def where(&block)
    # ...
  end
end

class ReceiveRequest < RSpec::Mocks::Matchers::Receive
  include Receiver
end

[...]

PR #1056, opened 2016-01-16
Allow inheriting mocks matchers

nepalez:
With reference to #1055

Allows rspec matchers inherited from `RSpec::Mocks::Matchers::*` to be used
both in mocks and expectations:

```ruby
  class ReceiveFoo < RSpec::Mocks::Matchers::Receive
    def initialize
      super :foo, nil
    end
  end

  describe "receiving foo" do
    it "works" do
      object = double

      allow(object).to receive_foo
      expect(object).to receive_foo

      object.foo
    end
  end
```

### DIFF
--- a/lib/rspec/mocks/targets.rb
+++ b/lib/rspec/mocks/targets.rb
@@ -38,7 +38,7 @@ module RSpec
     private
 
       def matcher_allowed?(matcher)
-        matcher.class.name.start_with?("RSpec::Mocks::Matchers".freeze)
+        Matchers::Matcher === matcher
       end
 
       def define_matcher(matcher, name, &block)
--- a/spec/rspec/mocks/matchers/receive_spec.rb
+++ b/spec/rspec/mocks/matchers/receive_spec.rb
@@ -90,6 +90,12 @@ module RSpec
           }.to raise_error(UnsupportedMatcherError)
         end
 
+        it 'does support inherited matchers', :unless => options.include?(:allow_other_matchers) do
+          receive_foo = Class.new(RSpec::Mocks::Matchers::Receive).new(:foo, nil)
+          wrapped.to receive_foo
+          receiver.foo
+        end
+
         it 'does not get confused by messages being passed as strings and symbols' do
           wrapped.to receive(:foo).with(1) { :a }
           wrapped.to receive("foo").with(2) { :b }
--- a/lib/rspec/mocks.rb
+++ b/lib/rspec/mocks.rb
@@ -117,6 +117,7 @@ module RSpec
 
     # Namespace for mock-related matchers.
     module Matchers
+      autoload :Matcher,             "rspec/mocks/matchers/matcher"
       autoload :HaveReceived,        "rspec/mocks/matchers/have_received"
       autoload :Receive,             "rspec/mocks/matchers/receive"
       autoload :ReceiveMessageChain, "rspec/mocks/matchers/receive_message_chain"
--- a/lib/rspec/mocks/matchers/have_received.rb
+++ b/lib/rspec/mocks/matchers/have_received.rb
@@ -3,6 +3,8 @@ module RSpec
     module Matchers
       # @private
       class HaveReceived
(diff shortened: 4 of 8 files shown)


lib/rspec/mocks/targets.rb
```diff
@@ -38,7 +38,9 @@ def self.disallow_negation(method_name)
     private
 
       def matcher_allowed?(matcher)
-        matcher.class.name.start_with?("RSpec::Mocks::Matchers".freeze)
+        ancestors = matcher.class.ancestors
+        namespace = "RSpec::Mocks::Matchers"
+        ancestors.any? { |klass| klass.name.to_s.start_with? namespace }
```
myronmarston: Suggestion: instead of doing a linear search over all the ancestors and _implicitly_ using the matcher class name, what do you think of creating an explicit "tag" module that you can check instead?  I'm thinking of something like this:

``` ruby
module RSpec
  module Mocks
    module Matchers
      # exists just so we can "tag" all rspec-mocks matchers to make detection of them easier.
      module Matcher; end

      # now include the module in each of our matcher classes...
      class HaveReceived
        include Matcher
      end

      class Receive
        include Matcher
      end

      class ReceiveMessages
        include Matcher
      end

      class ReceiveMessageChain
        include Matcher
      end

      # ...and then we can detect rspec-mocks matchers with a simple `===` check:
      def matcher_allowed?(matcher)
        RSpec::Mocks::Matchers::Matcher === matcher
      end
    end
  end
end
```

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -90,6 +90,16 @@ module Mocks
...
 
+        it 'does support inherited matchers', :unless => options.include?(:allow_other_matchers) do
+          receive_foo = Class.new(RSpec::Mocks::Matchers::Receive).new(:foo, nil)
+          object = double
+
+          allow(object).to receive_foo
+          expect(object).to receive_foo
+
+          object.foo
+        end
```
myronmarston: This is a shared spec that is meant to run in 4 different contexts:

* `expect(object).to receive_foo`
* `allow(object).to receive_foo`
* `expect_any_instance_of(object).to receive_foo`
* `allow_any_instance_of(object).to receive_foo`

But as you've written it, it tests just the first 2 of these, and does so for all 4 contexts this shared spec is run in.  Instead, I think it should be:

``` ruby

        it 'supports inherited matchers' do
          receive_foo = Class.new(RSpec::Mocks::Matchers::Receive).new(:foo, nil)
          wrapped.to receive_foo
          receiver.foo
        end
```

`wrapped` is polymorphic and will return `allow(object)`, `expect(object)` etc.  `receiver` will return `object` or an instance of the class passed to `allow_any_instance_of(klass)`/`expect_any_instance_of(klass)` -- so by using these, you can write one spec and have it operate for all the contexts appropriately.

nepalez: Ahh, I got it, thanks!

lib/rspec/mocks/targets.rb
nepalez: Amazing technics :)

It seems offhand, that if we go this way, it can lead us to extraction of the "essential/mockable" part of the matcher to the module. For now this is just a speculation, I need time to play with the idea.

JonRowe: > It seems offhand, that if we go this way, it can lead us to extraction of the "essential/mockable" part of the matcher to the module. For now this is just a speculation, I need time to play with the idea.

No. Delegation is more useful than inheritance in this fashion. We don't want to encourage more inheritance of our private APIs only support reasonable use cases like this. This module should only be used to improve the detection of inherited matchers from our own.

soulcutter: https://en.wikipedia.org/wiki/Marker_interface_pattern :+1:

JonRowe:
Thanks! :)

PR #1060, opened 2016-02-04
Restore 'owned' stubbed methods only in ruby 2.3+.

### LINKED ISSUE #1057: Aliased method disapper on rspec-mocks 3.4.1
```ruby
class A
  def g1
    1
  end
end

class B < A
  def f
    g2
  end

  alias g2 g1
end

describe '#f', order: true do
  let(:b) { B.new }

  it 'call B#g2' do
    expect_any_instance_of(B).to receive(:g2).with(no_args)
    b.f
  end

  it { expect(b.f).to eq 1 }
end
```

This works fine on rspec-mocks 3.4.0.
But on rspec-mocks 3.4.1, below error occurs.

```
Failures:

  1) #f 
     Failure/Error: g2

     NameError:
       undefined local variable or method `g2' for #<B:0x007fd11f2f15c8>
     # ./test_spec.rb:9:in `f'
     # ./test_spec.rb:23:in `block (2 levels) in <top (required)>'
```

The version of Ruby is 2.1.5 and 2.2.3.

jrafanie:
Fixes #1057.

In b94fe10e34ececd8cda34c480d996f036a58c557, we changed behavior to
to restore 'owned' stubbed methods for all rubies.

This doesn't work for all rubies because Method#owner for aliased methods
is inconsistent [1].

Since, ruby 2.3+ broke our prior backup/restore mechanism, it's safer to change
the original behavior for ruby 2.3+ only.

A `@klass` can have methods implemented (see Method#owner) in `@klass`
or inherited from a superclass. In ruby 2.2 and earlier, we can copy
a method regardless of the 'owner' and restore it to `@klass` after
because a call to 'super' from `@klass's` copied method would end up
calling the original class's superclass's method.

With the commit below [2], available starting in 2.3.0, ruby changed
this behavior and a call to 'super' from the method copied to `@klass`
will call `@klass's` superclass method, which is the original
implementer of this method!  This leads to very strange errors
if `@klass's` copied method calls 'super', since it would end up
calling itself, the original method implemented in `@klass's`
superclass.

For ruby 2.3 and above, we need to only restore methods that
`@klass` originally owned.

[1] https://gist.github.com/jrafanie/0e5b8a9f4e8b4f84913f
[2] https://github.com/ruby/ruby/commit/c8854d2ca4be9ee6946e6d17b0e17d9ef130ee81

### DIFF
--- a/lib/rspec/mocks/any_instance/recorder.rb
+++ b/lib/rspec/mocks/any_instance/recorder.rb
@@ -193,7 +193,28 @@ module RSpec
           alias_method_name = build_alias_method_name(method_name)
           @klass.class_exec(@backed_up_method_owner) do |backed_up_method_owner|
             remove_method method_name
-            alias_method method_name, alias_method_name if backed_up_method_owner[method_name.to_sym] == self
+
+            # A @klass can have methods implemented (see Method#owner) in @klass
+            # or inherited from a superclass. In ruby 2.2 and earlier, we can copy
+            # a method regardless of the 'owner' and restore it to @klass after
+            # because a call to 'super' from @klass's copied method would end up
+            # calling the original class's superclass's method.
+            #
+            # With the commit below, available starting in 2.3.0, ruby changed
+            # this behavior and a call to 'super' from the method copied to @klass
+            # will call @klass's superclass method, which is the original
+            # implementer of this method!  This leads to very strange errors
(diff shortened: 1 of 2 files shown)


JonRowe:
:+1:

jrafanie:
Thank you @JonRowe !!!

PR #1063, opened 2016-02-29
fixes #1062 -- make the error message when and_wrap_original is called on a pure double not refer to and_call_original

### LINKED ISSUE #1062: misleading error message
https://github.com/rspec/rspec-mocks/blob/v3.5.0.beta1/lib/rspec/mocks/message_expectation.rb#L117 raises an error when and_wrap_original is called on a pure double.  However, the error message refers to `and_call_original`, not `and_wrap_original`, which leads to confusion when you are not calling `and_call_original` at all.

There was a time that `and_wrap_original` was not documented and not part of the officially supported API, and at that time, this message would have been a very reasonable choice.  However, `and_wrap_original` is now documented and public, so the error message should refer to it.

betesh:


JonRowe:
Thanks

PR #1070, opened 2016-04-05
Update README.md

Jwan622:
Making the documentation regarding mock doubles a bit clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ making the combined instantiation/stub declaration very terse:
 double(:foo => 'bar')
 ```
 
+So, the above examples can be written simply as:
+
+```ruby
+allow(book).to receive(:title).and_return(book)
+```
+
 This is particularly nice when providing a list of test doubles to a method
 that iterates through them:
 


README.md
```diff
@@ -94,6 +94,12 @@ making the combined instantiation/stub declaration very terse:
 double(:foo => 'bar')
 ```
 
+So, the above examples can be written simply as:
+
+```ruby
+allow(book).to receive(:title).and_return(book)
+```
+
```
JonRowe: This doesn't make it clearer at all to me, the examples above are already rewriting this form, in fact the first examples cover this exact syntax (without looping back to `book`). Can you explain what you're clarifying here?

Jwan622: Sure. So I'm a relatively new programmer with a writing and logic background. I think this is clearer because it actually connects the dots for people reading this documentation and it's the more fleshed out version of the conclusion that you're alluding to. It answers the question "Can you simply pass in book even though it seems to have an extra argument which functions as a name? Is this what they mean?" I think the extra information is harmless, non-verbose, and finishes the job.

JonRowe: We're not alluding to this as a conclusion, as we state it up front with `allow(book).to receive(:title).and_return("The RSpec Book")`

mikegee:
This is kinda stale now and I don't really understand the intent either. @Jwan622 would you like to have another shot at explaining your idea here with the knowledge and experience you've gained since this was opened? Or, perhaps it is time to close this?

xaviershay:
(applying our "close by default, reopen to continue discussion" policy)

PR #1072, opened 2016-04-11
Correct allow_expectations_on_nil error message

martinemde:
The error message suggested that you set the wrong config.

`config.allow_expectations_on_nil` instead of
`config.allow_message_expectations_on_nil`


JonRowe:
Thanks! :heart:

PR #1082, opened 2016-05-03
Add spec for delegating to the original instance method w/in the block implementation

padi:
In [this example](https://www.relishapp.com/rspec/rspec-mocks/v/3-3/docs/working-with-legacy-code/any-instance#block-implementation-is-passed-the-receiver-as-first-arg) `String#slice` was replaced within the block with `String#[]`. In some cases, you don't have to luxury of having an equivalent method to rely on. Sometimes, you just want to use the same method within the block and include some assertions before it (or do something else if you wanted to).

This example clears that up. Hopefully it will save someone a few minutes/hours. :) Didn't know [UnboundMethod](docs.ruby-lang.org/en/2.2.0/UnboundMethod.html) in ruby before.

### DIFF
--- a/features/old_syntax/any_instance.feature
+++ b/features/old_syntax/any_instance.feature
@@ -82,6 +82,26 @@ Feature: `any_instance`
     When I run `rspec spec/example_spec.rb`
     Then the examples should all pass
 
+  Scenario: Block implementation supports delegating to the original implementation
+    Given a file named "spec/any_instance_delegate_to_original_spec.rb" with:
+      """ruby
+      RSpec.describe "Stubbing any instance of a class" do
+        it 'supports delegating to the original instance method within the block implementation' do
+          original_instance_method = String.instance_method(:slice)
+          String.any_instance.stub(:slice) do |instance, start, length|
+            expect(start).to be_a Fixnum
+            expect(length).to be_a Fixnum
+
+            original_instance_method.bind(instance).call(start, length)
+          end
+
+          expect('string'.slice(2, 3)).to eq('rin')
+        end
+      end
+      """
+    When I run `rspec spec/any_instance_delegate_to_original_spec.rb`
+    Then the examples should all pass
+
   Scenario: Expect a message on any instance of a class
     Given a file named "spec/example_spec.rb" with:
       """ruby


JonRowe:
I'm not sure how I feel about this, for one this isn't actually documenting anything within RSpec, just Ruby itself, and for another it's using the legacy syntax, on a legacy feature...

myronmarston:
> I'm not sure how I feel about this, for one this isn't actually documenting anything within RSpec, just Ruby itself, and for another it's using the legacy syntax, on a legacy feature...

That sums up my feeling, as well.

myronmarston:
Also, this isn't how we'd recommend doing this, anyway.  We provide `and_wrap_original` for cases like this:

``` ruby
allow_any_instance_of(String).to receive(:slice).and_wrap_original do |original_method, start, length|
  expect(start).to be_a Fixnum
  expect(length).to be_a Fixnum

  original_method.call(start, length)
end

expect('string'.slice(2, 3)).to eq('rin')
```

padi:
@myronmarston @JonRowe totally missed `and_wrap_original` while reading the docs. My bad. I immediately took the examples [here](https://www.relishapp.com/rspec/rspec-mocks/v/3-4/docs/working-with-legacy-code/any-instance#block-implementation-is-passed-the-receiver-as-first-arg) as an (somewhat) exhaustive list of things you could do with `*_any_instance_of` methods. Would it make sense then to add the suggestion you gave into `features/old_syntax/any_instance.feature`? Or `*_any_instance_of` methods are bound to be deprecated in the future?

JonRowe:
Sure, if you' like to work that up

myronmarston:
FWIW, the exhaustive list of things you can do to configure responses is here:

https://www.relishapp.com/rspec/rspec-mocks/v/3-4/docs/configuring-responses

That applies to `allow`, `expect`, `allow_any_instance_of`, `expect_any_instance_of`, etc.  I don't think it makes sense to list all of them in each of those places (that's why we have a whole "configuring responses" section) but maybe an explicit callout with a link to that section would do the trick?

padi:
An explicit link would definitely help. 👍  Added in #1085

PR #1085, opened 2016-05-04
Add explicit link to Configuring Responses section for `*_any_instance_of`

padi:
To suggest that it can also be used without having to add an extraneous example in the feature specs.

As suggested in #1082

### DIFF
--- a/features/working_with_legacy_code/any_instance.feature
+++ b/features/working_with_legacy_code/any_instance.feature
@@ -11,6 +11,8 @@ Feature: Any Instance
 
   These methods add the appropriate stub or expectation to all instances of `Widget`.
 
+  You can also [configure the responses](../configuring-responses) in the same manner.
+
   This feature is sometimes useful when working with legacy code, though in general we
   discourage its use for a number of reasons:
 


padi:
Not sure how this can make the travis CI fail but the tests pass when `bin/rake` is run locally.

JonRowe:
Travis runs `script/run_build` which does more than `bin/rake` that said I'm not sure whats failing here

padi:
When I run `script/run_build` locally, the rspec specs pass but stops when attempting to start cucumber:

~~~sh
$ script/run_build
# rspec runs successfully ...
============ Starting cukes ===============
/Users/owl/Projects/rspec-dev/repos/rspec-mocks/bin/cucumber
/Users/owl/.rbenv/versions/2.2.2/bin/gem:1:in `require': cannot load such file -- bundler/setup (LoadError)
$
~~~

It seems that it failed. Tried adding a `bundle exec script/run_build` before it though, and everything runs successfully.

The message is different for travis CI though. The same error* occurs in the previous PR #1084, so it might have something to do with the setup in travis CI.

* the error being: `The command "script/run_build" exited with 1.`

JonRowe:
Build error was unrelated, fixed elsewhere so thanks!

PR #1095, opened 2016-06-15
Provide a nice string representation of a message expectation.

myronmarston:
For the new book, we're using `rspec/mocks/standalone` in an `irb` session to demonstrate some concepts.  Unfortunately, the existing `to_s` output of a message expectation is super long and really scary looking.  This makes it much nicer.


myronmarston:
@rspec/rspec I'd really like to get this in 3.5 if possible :).

soulcutter:
This looks great - I'm happy with that implementation. CI is hard to please these days - when it's green, I think this is good to go

Issue #1098, opened 2016-06-22
Using mocks on methods with refinements fails


dreyks:
In a situation when a method comes from an included module and then this same method is refined, invoking `expect().to receive` fails as if the method is not present at all.

After spending some time in debugger I realised that the expectance itself passes, but the [restore_original_visibility](https://github.com/rspec/rspec-mocks/blob/master/lib/rspec/mocks/method_double.rb#L104) fails, because no method was found to change visibility for

Reproduction self-executing snippet. Note that if I move `inner` method directly into `Parent` instead of `include Inc` the example passes without failures

``` ruby
#!/usr/bin/env ruby

module Inc
  def inner
    'inc'
  end
end

class Parent
  include Inc
end

module Ref
  refine Parent do
    def inner
      super + ' refined'
    end
  end
end

class Child < Parent
  using Ref

  def outer
    inner
  end
end

require 'rspec/autorun'

describe Child do
  describe '#outer' do
    it 'calls inner' do
      child = Child.new
      expect(child).to receive(:inner)
      child.outer
    end
  end
end
```

samphippen:
@dreyks your example works for me on RSpec master on both 2.4 and 2.2. However, I will note that refinements do not work with `any_instance`. As per @JonRowe's PR.

mikegee:
If the repro doesn’t fail anymore, is it time to close this?

dreyks:
hmm... I've just rerun the example on both ruby 2.4.1 and 2.2.6 with rspec@master (3.8.pre) and it fails with 
```
NameError:
       undefined method `inner' for class `#<Class:#<Child:0x007ffd1785b670>>'
```

PR #1101, opened 2016-06-29
Refinements break mocks

JonRowe:
Demonstration of rspec/rspec-mocks#1098


JonRowe:
Migrated to monorepo.

PR #1104, opened 2016-07-11
Temporarily suppress verifying partial doubles.

### LINKED ISSUE #1102: Add the ability to temporarily turn off `verify_partial_doubles`
In https://github.com/rspec/rspec-rails/issues/1076 and https://github.com/rspec/rspec-rails/issues/1219 we encountered some issues with partial mocks and view specs. We should add the ability to temporarily disable `verify_partial_doubles`.

The implementation signature could potentially look like `without_verifying_partial_doubles(&blk)` such that

``` ruby
around do |ex|
  without_verifying_partial_doubles { ex.run }
end
```

and

``` ruby
it "does something weird with mocks" do
  without_verifying_partial_doubles do
    ...
  end
end
```

The implementation strategy for this would simply set/reset the values in RSpec Mocks' configuration which are found [here](https://github.com/rspec/rspec-mocks/blob/master/lib/rspec/mocks/configuration.rb#L148-L154)

I'm happy to mentor on this one if someone needs help

JonRowe:
Implements: 

``` Ruby
without_verifying_partial_doubles do
  ...
end
```

Closes #1102 and should be used to suppress partial doubles in view specs, /cc @samphippen

### DIFF
--- a/lib/rspec/mocks/configuration.rb
+++ b/lib/rspec/mocks/configuration.rb
@@ -8,6 +8,7 @@ module RSpec
         @verify_doubled_constant_names = false
         @transfer_nested_constants = false
         @verify_partial_doubles = false
+        @temporarily_suppressing_verification = false
       end
 
       # Sets whether RSpec will warn, ignore, or fail a test when
@@ -152,6 +153,11 @@ module RSpec
       def verify_partial_doubles?
         @verify_partial_doubles
       end
+
+      # @private
+      # Used to track wether we are temporarily suppressing verifying doubles
+      # with `without_verifying_partial_doubles { ... }`
+      attr_accessor :temporarily_suppressing_verification
 
       if ::RSpec.respond_to?(:configuration)
         def color?
--- a/lib/rspec/mocks/configuration.rb
+++ b/lib/rspec/mocks/configuration.rb
@@ -156,7 +156,7 @@ module RSpec
 
       # @private
       # Used to track wether we are temporarily suppressing verifying partial
-      # doubles with `without_verifying_partial_doubles { ... }`
+      # doubles with `without_partial_double_verification { ... }`
       attr_accessor :temporarily_suppress_partial_double_verification
 
       if ::RSpec.respond_to?(:configuration)
--- a/lib/rspec/mocks/configuration.rb
+++ b/lib/rspec/mocks/configuration.rb
@@ -8,7 +8,7 @@ module RSpec
         @verify_doubled_constant_names = false
         @transfer_nested_constants = false
         @verify_partial_doubles = false
-        @temporarily_suppressing_partial_double_verification = false
+        @temporarily_suppress_partial_double_verification = false
       end
 
       # Sets whether RSpec will warn, ignore, or fail a test when
@@ -157,7 +157,7 @@ module RSpec
       # @private
       # Used to track wether we are temporarily suppressing verifying partial
       # doubles with `without_verifying_partial_doubles { ... }`
-      attr_accessor :temporarily_suppressing_partial_double_verification
+      attr_accessor :temporarily_suppress_partial_double_verification
 
       if ::RSpec.respond_to?(:configuration)
         def color?
--- a/lib/rspec/mocks/configuration.rb
+++ b/lib/rspec/mocks/configuration.rb
@@ -8,7 +8,7 @@ module RSpec
         @verify_doubled_constant_names = false
         @transfer_nested_constants = false
         @verify_partial_doubles = false
-        @temporarily_suppressing_verification = false
+        @temporarily_suppressing_partial_double_verification = false
       end
 
       # Sets whether RSpec will warn, ignore, or fail a test when
@@ -155,9 +155,9 @@ module RSpec
       end
 
       # @private
-      # Used to track wether we are temporarily suppressing verifying doubles
-      # with `without_verifying_partial_doubles { ... }`
-      attr_accessor :temporarily_suppressing_verification
+      # Used to track wether we are temporarily suppressing verifying partial
+      # doubles with `without_verifying_partial_doubles { ... }`
+      attr_accessor :temporarily_suppressing_partial_double_verification
 
       if ::RSpec.respond_to?(:configuration)
         def color?
--- a/lib/rspec/mocks/example_methods.rb
+++ b/lib/rspec/mocks/example_methods.rb
@@ -286,7 +286,7 @@ module RSpec
       # block, this is useful in situations where methods are defined at run
       # time and you wish to define stubs for them but not turn off partial
       # doubles for the entire run suite. (e.g. view specs in rspec-rails).
-      def without_verifying_partial_doubles
+      def without_partial_double_verification
         Mocks.configuration.temporarily_suppress_partial_double_verification = true
         yield
       ensure
--- a/lib/rspec/mocks/example_methods.rb
+++ b/lib/rspec/mocks/example_methods.rb
@@ -287,10 +287,10 @@ module RSpec
       # time and you wish to define stubs for them but not turn off partial
       # doubles for the entire run suite. (e.g. view specs in rspec-rails).
       def without_verifying_partial_doubles
-        Mocks.configuration.temporarily_suppressing_partial_double_verification = true
+        Mocks.configuration.temporarily_suppress_partial_double_verification = true
         yield
       ensure
-        Mocks.configuration.temporarily_suppressing_partial_double_verification = false
+        Mocks.configuration.temporarily_suppress_partial_double_verification = false
       end
 
       # @method expect
--- a/lib/rspec/mocks/example_methods.rb
+++ b/lib/rspec/mocks/example_methods.rb
@@ -287,10 +287,10 @@ module RSpec
       # time and you wish to define stubs for them but not turn off partial
       # doubles for the entire run suite. (e.g. view specs in rspec-rails).
       def without_verifying_partial_doubles
-        Mocks.configuration.temporarily_suppressing_verification = true
+        Mocks.configuration.temporarily_suppressing_partial_double_verification = true
         yield
       ensure
-        Mocks.configuration.temporarily_suppressing_verification = false
+        Mocks.configuration.temporarily_suppressing_partial_double_verification = false
       end
 
       # @method expect
--- a/lib/rspec/mocks/example_methods.rb
+++ b/lib/rspec/mocks/example_methods.rb
@@ -282,6 +282,17 @@ module RSpec
         Matchers::HaveReceived.new(method_name, &block)
       end
 
+      # Turns off the verifying of partial doubles for the duration of the
+      # block, this is useful in situations where methods are defined at run
+      # time and you wish to define stubs for them but not turn off partial
+      # doubles for the entire run suite. (e.g. view specs in rspec-rails).
+      def without_verifying_partial_doubles
+        Mocks.configuration.temporarily_suppressing_verification = true
+        yield
+      ensure
+        Mocks.configuration.temporarily_suppressing_verification = false
+      end
+
       # @method expect
       # Used to wrap an object in preparation for setting a mock expectation
       # on it.
--- a/lib/rspec/mocks/example_methods.rb
+++ b/lib/rspec/mocks/example_methods.rb
@@ -287,10 +287,11 @@ module RSpec
       # time and you wish to define stubs for them but not turn off partial
       # doubles for the entire run suite. (e.g. view specs in rspec-rails).
       def without_partial_double_verification
+        original_state = Mocks.configuration.temporarily_suppress_partial_double_verification
         Mocks.configuration.temporarily_suppress_partial_double_verification = true
         yield
       ensure
-        Mocks.configuration.temporarily_suppress_partial_double_verification = false
+        Mocks.configuration.temporarily_suppress_partial_double_verification = original_state
       end
 
       # @method expect
--- a/lib/rspec/mocks/verifying_proxy.rb
+++ b/lib/rspec/mocks/verifying_proxy.rb
@@ -36,7 +36,6 @@ module RSpec
       end
 
       def ensure_implemented(method_name)
-        return if Mocks.configuration.temporarily_suppress_partial_double_verification
         return unless method_reference[method_name].unimplemented?
 
         @error_generator.raise_unimplemented_error(
@@ -47,7 +46,6 @@ module RSpec
       end
 
       def ensure_publicly_implemented(method_name, _object)
-        return if Mocks.configuration.temporarily_suppress_partial_double_verification
         ensure_implemented(method_name)
         visibility = method_reference[method_name].visibility
 
@@ -120,6 +118,11 @@ module RSpec
         end
 
         optional_callback_invocation_strategy.call(@doubled_module)
+      end
+
+      def ensure_implemented(_method_name)
+        return if Mocks.configuration.temporarily_suppress_partial_double_verification
+        super
       end
 
       def method_reference
--- a/lib/rspec/mocks/verifying_proxy.rb
+++ b/lib/rspec/mocks/verifying_proxy.rb
@@ -36,7 +36,7 @@ module RSpec
       end
 
       def ensure_implemented(method_name)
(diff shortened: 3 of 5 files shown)


lib/rspec/mocks/example_methods.rb
```diff
@@ -282,6 +282,16 @@ def have_received(method_name, &block)
...
       end
 
+      # Turns off the verifying of partial doubles for the duration of the
+      # block, this is useful in situations where methods are defined at run
+      # time and you wish to define stubs for them but not turn off partial
+      # doubles for the entire run suite. (e.g. view specs in rspec-rails).
+      def without_verifying_partial_doubles
+        RSpec::Mocks.space.temporarily_suppressing_verification = true
+        yield
+        RSpec::Mocks.space.temporarily_suppressing_verification = false
```
samphippen: this should be in an ensure to make sure it gets reset if the block throws an exception

JonRowe: Good point, well made. Fixed.

lib/rspec/mocks/verifying_proxy.rb
```diff
@@ -34,23 +57,25 @@ def add_message_expectation(method_name, opts={}, &block)
         ensure_implemented(method_name)
         super
       end
+    end
 
-      def ensure_implemented(method_name)
-        return unless method_reference[method_name].unimplemented?
+    # @private
+    module MaybeVerifyProxyMethods
```
myronmarston: Is it really necessary to introduce all these `Maybe...` and `Could...` modules and classes?  They make the ancestry of these classes really confusing and it's not clear to me why they are needed.  On top of that, given that you still check `source_space.temporarily_suppressing_verification?`it seems like these classes only have different behavior if `temporarily_suppressing_verification` is set when a stub or message expectation is created--but in that case, why do we need different classes at all?  For example, I think users would rightly expect this to work:

``` ruby
dbl = double("MyClass")
without_verifying_partial_doubles do
  allow(dbl).to receive(:some_method_that_would_fail_verification)
end
```

...but as implemented, it looks to me like that won't work.

I think it would be better and simpler to just _always_ check `temporarily_suppressing_verification` before calling `ensure_implemented` (or maybe check it in `ensure_implemented` itself so we don't have to duplicate the check in the callers).  Then it seems like there'd be no need for these extra confusing modules and classes.

lib/rspec/mocks/space.rb
```diff
@@ -56,6 +56,8 @@ def raise_lifecycle_message
     # @private
     class Space
       attr_reader :proxies, :any_instance_recorders, :proxy_mutex, :any_instance_mutex
+      attr_accessor :temporarily_suppressing_verification
+      alias :temporarily_suppressing_verification? :temporarily_suppressing_verification
```
myronmarston: I'm not sure `Space` is the right place for this state.  Consider: spaces can be nested.  As currently implemented, this would fail:

``` ruby
without_verifying_partial_doubles do
  RSpec::Mocks.with_temporary_scope do
    dbl = instance_double("MyClass")
    allow(dbl).to receive(:some_method_that_would_fail_verification)
  end
end
```

`RSpec::Mocks.with_temporary_scope` creates a nested temporary space, which will have `temporarily_suppressing_verification` set to false.  Unless you specifically want `temporarily_suppressing_verification` not to inherit to nested scopes (in which case, I'd ask for a use case and for this to be well documented) I think it's better to attach this state to `RSpec::Mocks.configuration` where it would be truly global and not have this kind of gotcha.

lib/rspec/mocks/verifying_proxy.rb
JonRowe: They were introduced to prevent having to *always* check, as I think the normal use case is not going to be to skip verification, given we check the permanent setting over in the space it seems prudent to do the same for this edge case.

The class hierarchy was introduced to avoid repetition, I could refactor that a bit more into module instead if you'd like? Or I could just do a bit of repetition :)

lib/rspec/mocks/space.rb
JonRowe: I'm open to moving the state into configuration, but I was worried about the use of global state, and exposing this as a public api (I think it should be private)

lib/rspec/mocks/verifying_proxy.rb
myronmarston: > They were introduced to prevent having to always check, as I think the normal use case is not going to be to skip verification,

I think the cost of the check is negligible--but the cost of the added maintenance from trying to avoid the check when not needed is significant.

If we're going to take on the maintenance burden of the extra double class types and modules, I would want benchmarks showing what kind of performance benefit the added maintenance cost is buying us so we can make an informed decision.

On top of that, there's the snippet I posted above:

``` ruby
dbl = instance_double("MyClass")
without_verifying_partial_doubles do
  allow(dbl).to receive(:some_method_that_would_fail_verification)
end
```

If I was a user, I'd expect that to work, but if we implement things to avoid the checks for verifying doubles created before the `without_verifying_partial_doubles`, there is no way to make it work.

Do you not view that as a bug like I do?

JonRowe:
Simplified at the expense of always checking if we are suppressing verification.

myronmarston:
Thanks for the updates, @JonRowe.  It's nice to see how much simpler this got.

BTW, I just realized this is for partial doubles only.  Somehow I missed that before.  Are we sure there's not a use case for being able to disable all verification?  If/when we ever discover a use case I'd hate to have to introduce a 2nd API for that when we could have just made this handle both.  Although, thinking out loud here -- if the use case is just for view specs due to how rails works it might be kinda weird if verifying doubles just don't work in that kind of example group so making this apply to partial doubles only seems better.

Now, if we do make this apply only to partial doubles as it does now, it makes me wonder if we need the new `temporarily_suppressing_verification` flag on config.  After all, partial doubles already have a flag on configuration for enabling/disabling them via `verify_partial_doubles` -- so why do we need two flags that control the same behavior?  Couldn't `without_verifying_partial_doubles` be implemented simply like this?

``` ruby
def without_verifying_partial_doubles
  original_value = Mocks.configuration.verify_partial_doubles?
  Mocks.configuration.verify_partial_doubles = false
  begin
    yield
  ensure
    Mocks.configuration.verify_partial_doubles = original_value
  end
end
```

(On a side note, if we keep `temporarily_suppressing_verification` I think it should be renamed---the name does not indicate that it applies only to partial doubles).

igor47:
I have a use case for temporarily supressing all verification -- the AWS_SDK gem. the clients in these gems (both v1 and v2) do not actually define any methods, except dynamically at run time. they actually talk to the AWS API to retrieve a list of supported API calls. this means that even a full double (for instance, `instance_double(AWS::EC2::Client)`) will cause problems.

really, i'm looking for a way to get verifying doubles on all my other code but disable it on specific doubles which i want to be non-verifying because the objects they're doubling are complicated. this is a handy shortcut, but it would be awesome to be able to do something like `instance_double(My::Class, :verifying => false)`

JonRowe:
> I have a use case for temporarily supressing all verification -- the AWS_SDK gem. the clients in these gems (both v1 and v2) do not actually define any methods, except dynamically at run time. they actually talk to the AWS API to retrieve a list of supported API calls. this means that even a full double (for instance, instance_double(AWS::EC2::Client)) will cause problems.

So use a normal double, `double(AWS::EC2::Client)` there's no benefit to using `instance_double` if you can't verify it.

myronmarston:
> but it would be awesome to be able to do something like `instance_double(My::Class, :verifying => false)`

That's what `double('My::Class')` does.

igor47:
good point!

JonRowe:
@myronmarston finally got this green on 1.8.7, I can't use the existing config method because it breaks things all over the place (I believe thats why I went down the more complex route in the first place!) but I've renamed the config variable to better reflect this, care to re-review?

lib/rspec/mocks/verifying_proxy.rb
```diff
@@ -36,6 +36,7 @@ def add_message_expectation(method_name, opts={}, &block)
       end
 
       def ensure_implemented(method_name)
+        return if Mocks.configuration.temporarily_suppressing_partial_double_verification
```
myronmarston: This is defined in the `VerifyingProxyMethods` module which, if I'm not mistaken, applies to normal verifying doubles (e.g. `instance_double` and `class_double`) in addition to partial doubles.  As such, I think there's a bug here: while the config option is meant to only affect partial doubles, it looks like your implementation would suppress it for normal doubles, too.

I think the solution is to move these checks into methods defined in `VerifyingPartialDoubleProxy` -- although you might have to define new ones where the checks can go (e.g. with a call to `super()`)

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -377,6 +377,17 @@ def defined_private_method
...
+      it 'can be temporarily supressed' do
+        without_verifying_partial_doubles do
+          expect(object).to receive(:fictitious_method) { 'works' }
+        end
+        expect(object.fictitious_method).to eq 'works'
+
+        expect {
+          expect(object).to receive(:another_fictitious_method) { 'works' }
+        }.to raise_error RSpec::Mocks::MockExpectationError
+      end
```
myronmarston: As I mentioned above, I think there's a bug in your implementation that makes `without_verifying_partial_doubles` apply to normal doubles in addition to partial doubles.  Regardless, it would be nice to have some specs document how normal doubles work in a `without_verifying_partial_doubles` (e.g. they should work exactly the same).  Maybe I'm wrong and a spec will show that I'm wrong, or maybe I'm right and a spec will confirm that.

lib/rspec/mocks/verifying_proxy.rb
```diff
@@ -193,6 +195,8 @@ def unimplemented?
       def self.for(object, method_name, proxy)
         if ClassNewMethodReference.applies_to?(method_name) { object }
           VerifyingExistingClassNewMethodDouble
+        elsif Mocks.configuration.temporarily_suppressing_partial_double_verification
+          MethodDouble
```
myronmarston: It's not very clear to me why this is necessary.  Playing around with it locally, it looks like it's completely unnecessary for 2.x (at least, all the specs passed w/o this on 2.x!) but I can see that on 1.8.7, an argument failure occurs w/o this.

It'd be nice to do away with this, or, if we keep it only for 1.8.7, clearly document it so we know we can remove it when we remove 1.8.7.

That said, the fact that the arg verification caused the failure on 1.8.7 makes me realize that you are not specifically suppressing argument validation--which makes me worry that there could be some alternate way that it could be hit on other rubies.  So I think a better solution would be to override `validate_arguments!` on `VerifyingPartialDoubleProxy` to have it check the flag and then `super()` if needed.

Thoughts?

lib/rspec/mocks/configuration.rb
```diff
@@ -153,6 +154,11 @@ def verify_partial_doubles?
         @verify_partial_doubles
       end
 
+      # @private
+      # Used to track wether we are temporarily suppressing verifying partial
+      # doubles with `without_verifying_partial_doubles { ... }`
+      attr_accessor :temporarily_suppressing_partial_double_verification
```
myronmarston: I think it's a bit odd to have the `-ing` form of the verb in a boolean flag like this.  How about we shorten this to `temporarily_suppress_partial_double_verification`?

lib/rspec/mocks/verifying_proxy.rb
JonRowe: Moved and conditionally wrapped.

lib/rspec/mocks/verifying_proxy.rb
JonRowe: You're right, as is standard :)

spec/rspec/mocks/partial_double_spec.rb
JonRowe: Yep, added.

lib/rspec/mocks/configuration.rb
JonRowe: Done.

JonRowe:
Changes made @myronmarston if you'd care to re-review

JonRowe:
Ping @rspec/rspec

lib/rspec/mocks/example_methods.rb
```diff
@@ -282,6 +282,17 @@ def have_received(method_name, &block)
...
+      # Turns off the verifying of partial doubles for the duration of the
+      # block, this is useful in situations where methods are defined at run
+      # time and you wish to define stubs for them but not turn off partial
+      # doubles for the entire run suite. (e.g. view specs in rspec-rails).
+      def without_verifying_partial_doubles
+        Mocks.configuration.temporarily_suppress_partial_double_verification = true
+        yield
+      ensure
+        Mocks.configuration.temporarily_suppress_partial_double_verification = false
+      end
```
myronmarston: IMO, this should set the config flag back to whatever its value was before the method was called, rather than assuming it was set to `false` (which may or may not be true).  Consider a case like:

``` ruby
module MyHelpers
  def do_stuff
    without_verifyiong_partil_doubles do
      # ...
    end
  end
end

RSpec.describe MyGroup do
  include MyHelpers

  it "works" do
    without_verifying_partial_doubles do
      do_something
      do_stuff
      do_something_else
    end
  end
end
```

As you've implemented this, partial double verification will _not_ be disabled while `do_something_else` runs, even though its in a `without_verifying_partial_doubles` block.  I'd argue this is a bug.

The solution is to change this method to stash the existing value of the config flag when the method starts, then set it to true, then set it back to the original value in the `ensure` block.

lib/rspec/mocks/example_methods.rb
```diff
@@ -282,6 +282,17 @@ def have_received(method_name, &block)
         Matchers::HaveReceived.new(method_name, &block)
       end
 
+      # Turns off the verifying of partial doubles for the duration of the
+      # block, this is useful in situations where methods are defined at run
+      # time and you wish to define stubs for them but not turn off partial
+      # doubles for the entire run suite. (e.g. view specs in rspec-rails).
+      def without_verifying_partial_doubles
```
myronmarston: Do you think `without_partial_double_verification` reads better for the method name? (I do...)

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -377,6 +377,27 @@ def defined_private_method
...
+
+      specify 'temporarily supressing partial doubles does not affect normal verifying doubles' do
+        klass = Class.new
+        object = nil
```
myronmarston: What is the purpose of `klass = Class.new`, `object = nil` and the `object =` bit below?  I don't see how these variables get used.

lib/rspec/mocks/example_methods.rb
JonRowe: No, but I don't feel strongly about it, changed.

lib/rspec/mocks/example_methods.rb
JonRowe: Fixed and added a spec demonstrating that nesting works.

lib/rspec/mocks/example_methods.rb
```diff
@@ -282,6 +282,18 @@ def have_received(method_name, &block)
...
+      # block, this is useful in situations where methods are defined at run
+      # time and you wish to define stubs for them but not turn off partial
+      # doubles for the entire run suite. (e.g. view specs in rspec-rails).
+      def without_partial_double_verification
+        original_state = Mocks.configuration.temporarily_suppress_partial_double_verification
+        Mocks.configuration.temporarily_suppress_partial_double_verification = true
+        yield
+      ensure
+        Mocks.configuration.temporarily_suppress_partial_double_verification = original_state
+      end
```
myronmarston: Is it worth caching `Mocks.configuration` in a local variable so there are fewer message sends here, given you use the config 3 times?

JonRowe: I benchmarked this, it's approximately 1ms quicker over 10000 uses to store the object in a local variable rather than call the method, so no?

JonRowe: I've attached a benchmark, if you agree merge away otherwise I can change it but its kinda ugly :(

myronmarston: No need.  Merge away.

PR #1106, opened 2016-07-13
Updates from rspec-dev (2016-07-14)

JonRowe:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,10 @@ end
 
 gem 'simplecov', '~> 0.8'
 
+if RUBY_VERSION < '2.0.0' || RUBY_ENGINE == 'java'
+  gem 'json', '< 2.0.0'
+end
+
 platforms :jruby do
   gem "jruby-openssl"
 end
--- a/spec/rspec/mocks_spec.rb
+++ b/spec/rspec/mocks_spec.rb
@@ -132,7 +132,7 @@ RSpec.describe RSpec::Mocks do
         expect(defined?(ValueY)).to be_falsey
       end
 
-      it 'does not allow the stubbed constants to be used after the scope in before(:all)' do
+      it 'does not allow the stubbed constants to be used after the scope in before(:all)', :pending => RSpec::Support::Ruby.jruby_9000? do
         expect(@error).to be_a(NameError)
         expect(@error.message).to include("ValueX")
       end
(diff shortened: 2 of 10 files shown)


spec/rspec/mocks_spec.rb
```diff
@@ -132,7 +132,7 @@
         expect(defined?(ValueY)).to be_falsey
       end
 
-      it 'does not allow the stubbed constants to be used after the scope in before(:all)' do
+      it 'does not allow the stubbed constants to be used after the scope in before(:all)', :pending => RSpec::Support::Ruby.jruby_9000? do
```
JonRowe: This is broken on JRuby 9000, but because of an internal `sprintf` which I'm not sure we can do anything about?

Gemfile
```diff
@@ -23,6 +23,10 @@ end
 
 gem 'simplecov', '~> 0.8'
 
+if RUBY_VERSION < '2.0.0' || RUBY_ENGINE == 'java'
+  gem 'json', '< 2.0.0'
+end
```
myronmarston: Why do we need to specify the `json` gem in the `Gemfile` for each gem?  Does rspec-mocks use it?

spec/rspec/mocks_spec.rb
myronmarston: I don't understand.  Can you explain a bit more?

JonRowe: It produces a warning on JRuby for rspec-expectations:

`lib/rspec/matchers/built_in/raise_error.rb:131 warning: too many arguments for format string`

but all it is a `to_s` call, googling this warning resulted in a slew of results pointing to an issue with `sprintf` on JRuby.

Gemfile
JonRowe: It gets installed as a dependency of our build as a whole, and then errors on older rubies. I think this is related to our cross repo testing.

JonRowe:
This is green baring the pending spec, ok to merge @myronmarston or does somebody want to look into that more?

Gemfile
myronmarston: I'm still confused by this...when we install gems for cross repo testing, we run `bundle install` from that repo's directory, and if it needs `json < 2.0.0` it should specify it.

myronmarston:
Merge away, although I still don't understand the need for the json thing :(

Gemfile
JonRowe: Ok I've looked into this some more, it's a dependency of simplecov. Which accepts `json (>= 1.8, < 3)`, but `json >= 2.0.0` requires Ruby 2.0.0, so it's either pin it for all, or pin it on older rubies.

myronmarston: That makes sense.  Do you mind leaving a comment about it being used by simplecov?

JonRowe: Done in a seperate commit and across repos too.

Issue #1109, opened 2016-08-04
TestDouble #freeze should return #self


alxberardi:
When calling the freeze method on a double, it is a good thing that actual freezing does not happen and that a warning is shown. 

That said, the expected return value for the freeze method on an object is the object itself.

In my particular case, I'm experiencing issues writing unit tests for classes that include [Memoizable](https://github.com/dkubb/memoizable).

With memoized methods, the gem caches the return value of the freeze method called on the return value of the memoized method, which, in the case of doubles, would be ``nil`` instead of the double itself.

The suggestion would be to return ``self`` in ``TestDouble#freeze`` (lib/rspec/mocks/test_double.rb:67)
```ruby
def freeze
  RSpec.warn_with("WARNING: you attempted to freeze a test double. This is explicitly a no-op as freezing doubles can lead to undesired behaviour when resetting tests.")
  self
end
```

Thanks!

JonRowe:
Happy to accept a PR (with a spec) for this :)

alxberardi:
Ok, I'll see if I can find a moment at some stage, thanks!

alxberardi:
Ok, I'll see if I can find a moment at some stage, thanks!

alxberardi:
Please see https://github.com/rspec/rspec-mocks/pull/1110 , thanks!

JonRowe:
Closed by #1110

PR #1110, opened 2016-08-05
Fixes Issue #1109: `TestDouble#freeze` should return `self`

alxberardi:


JonRowe:
Thanks!

JonRowe:
(pushed to 3-5-maintenance)

PR #1115, opened 2016-08-25
Remove legacy add_negative_message_expectation method

mrageh:
In the past this method was used to negate expectations, but since [1] it's not
been used. Therefore this commit removes it.

1.) https://github.com/rspec/rspec-mocks/commit/9357e4de76838157c8ea7d4a4ad6d0ae11882b22


mrageh:
@myronmarston I think this code is a relic of the past

JonRowe:
LGTM

PR #1131, opened 2016-12-12
Allow duck_type to expect methods to return specific values

maxhollmann:
Allows things like this:

    is_expected.to contain_exactly(
      a_hash_including(entity: duck_type(name: "Tim"), count: 1),
      a_hash_including(entity: duck_type(name: "Tom"), count: 4),
    )


myronmarston:
rspec-expectations already provides a matcher like this in the form of `have_attributes` (or `an_object_having_attributes`).  You can use that for this situation.

I don't think it's appropriate to expand what `duck_type` does because it's purpose is really just to match an object that implements a required set of methods, and not to actually call any of those methods.

Issue #1129, opened 2016-11-23
Add stub_chain documentation to README.md


glaucocustodio:
I find `stub_chain` very useful for some cases. I know there are concerns regarding Law of Demeter but I think we should instruct its usage at README.md.

Would a PR be welcome? Thank you.

JonRowe:
Given that nearly everything else is, sure, however it would have to be `receive_message_chain` which is the modern incarnation.

samphippen:
@glaucocustodio is this still something you'd like to file a PR for?

glaucocustodio:
I'm sorry for the delay. Please, see https://github.com/rspec/rspec-mocks/pull/1137.

PR #1137, opened 2017-01-03
Update README.md

glaucocustodio:
Add `receive_message_chain` example to README.md.

It closes https://github.com/rspec/rspec-mocks/issues/1129.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,22 @@ that iterates through them:
 order.calculate_total_price(double(:price => 1.99), double(:price => 2.99))
 ```
 
+### Stubbing a chain of methods
+
+You can use `receive_message_chain` in place of `receive` in certain circumstances to stub a chain of messages:
+
+```ruby
+allow(double).to receive_message_chain("foo.bar") { :baz }
+allow(double).to receive_message_chain(:foo, :bar => :baz)
+allow(double).to receive_message_chain(:foo, :bar) { :baz }
+```
+
+Given any of these three forms:
+
+```ruby
+double.foo.bar # => :baz
+```
+
 ## Consecutive return values
 
 When a stub might be invoked more than once, you can provide additional
--- a/README.md
+++ b/README.md
@@ -103,19 +103,17 @@ order.calculate_total_price(double(:price => 1.99), double(:price => 2.99))
 
 ### Stubbing a chain of methods
 
-You can use `receive_message_chain` in place of `receive` in certain circumstances to stub a chain of messages:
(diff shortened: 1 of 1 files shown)


samphippen:
I like this. Any thoughts from @rspec/rspec members?

README.md
```diff
@@ -101,6 +101,22 @@ that iterates through them:
 order.calculate_total_price(double(:price => 1.99), double(:price => 2.99))
 ```
 
+### Stubbing a chain of methods
+
+You can use `receive_message_chain` in place of `receive` in certain circumstances to stub a chain of messages:
```
myronmarston: > in certain circumstances

This makes it sound like these code snippets only work _sometimes_; you have to know if you're in one of the "certain circumstances" where these work.

These code snippets always work, though; (assuming you haven't disabled the newer stubbing syntax), so I'd remove the phrase "in certain circumstances" entirely.

README.md
```diff
@@ -101,6 +101,22 @@ that iterates through them:
...
+
+```ruby
+allow(double).to receive_message_chain("foo.bar") { :baz }
+allow(double).to receive_message_chain(:foo, :bar => :baz)
+allow(double).to receive_message_chain(:foo, :bar) { :baz }
+```
+
+Given any of these three forms:
```
myronmarston: Usually when I see a sentence like this, I expect the 3 forms to follow (it seems like it is introducing the 3 forms)....but it's not.

Might make more sense to just merge this snippet with the above:

``` ruby
# ...
allow(double).to receive_message_chain(:foo, :bar) { :baz }

# Given any of the above forms:
double.foo.bar # => :baz
```

glaucocustodio:
Please, check now @myronmarston.

README.md
```diff
@@ -101,6 +101,20 @@ that iterates through them:
 order.calculate_total_price(double(:price => 1.99), double(:price => 2.99))
 ```
 
+### Stubbing a chain of methods
+
+You can use `receive_message_chain` in place of `receive` to stub a chain of messages:
+
+```ruby
+allow(double).to receive_message_chain(:foo, :bar) { :baz }
```
myronmarston: I think it's fine that you showed the 3 ways that work before.  Any reason you removed it?  I didn't mean to suggest you remove them; in my suggested snippet, the `# ...` was a standin because I was too lazy :).

glaucocustodio: I'm sorry, I got it wrong. Will fix.

myronmarston:
LGTM.  Thanks!

PR #1138, opened 2017-01-07
Remove rubocop from 1.9.2

JonRowe:
As per rspec/rspec-expectations#963


xaviershay:
LGTM

JonRowe:
Merging as the failure is in expectations (which is in turn coupled to this)

Issue #112, opened 2012-03-01
mocking Observers is not cleaned up between tests


travisp:
This looks similar to issue 52: https://github.com/rspec/rspec-mocks/issues/52

Basically, if I am touching an observer (a singleton) in one test, it seems to affect another test.  Example:

```
it "acts on the after_create notification" do
  thing = FactoryGirl.build(:thing)
  FeedObserver.any_instance.should_receive(:after_create) #or FeedObserver.instance.should_receive(:after_create)
  thing.save!
end
it "creates a LogEvent" do
  obs = FeedObserver.instance
  expect {
    obs.after_create(@thing)
  }.to change(TimelineEvent, :count).by(1)  
end
```

In the second test FeedObserver.instance returns a mocked object

PR #1144, opened 2017-02-08
[InstanceMethodStasher] Say method is owned by class when the owner i…

### LINKED ISSUE #1143: Expectations on FileUtils aren't reverted on Ruby 1.8.7
The following test suite passes on Ruby 2.4 but fails on 1.8.7

```ruby
require "rspec"
require "fileutils"

RSpec.describe "dss" do
  it "foo" do
    expect(FileUtils).to receive(:mkpath).once
    FileUtils.mkpath "a"
  end

  it "bar" do
    expect(FileUtils.mkpath Dir.pwd).to eq([Dir.pwd])
  end
end
```

with the following output

```
.F

Failures:

  1) dss bar
     Failure/Error: expect(FileUtils.mkpath Dir.pwd).to eq([Dir.pwd])

     NoMethodError:
       undefined method `mkpath' for FileUtils:Module
     # ./foo_spec.rb:11

Finished in 0.00868 seconds (files took 0.21881 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./foo_spec.rb:10 # dss bar
```

segiddins:
…s the object

Closes #1143

### DIFF
--- a/lib/rspec/mocks/instance_method_stasher.rb
+++ b/lib/rspec/mocks/instance_method_stasher.rb
@@ -128,6 +128,17 @@ module RSpec
         # Hence, we verify that the owner actually has the method defined.
         # If the given owner does not have the method defined, we assume
         # that the method is actually owned by @klass.
+        #
+        # On 1.8, aliased methods can also report the wrong owner. Example:
+        # module M
+        #   def a; end
+        #   module_function :a
+        #   alias b a
+        #   module_function :b
+        # end
+        # The owner of M.b is the raw Module object, instead of the expected
+        # singleton class of the module
+        return true if RUBY_VERSION < '1.9' && owner == @object
         owner == @klass || !(method_defined_on_klass?(owner))
       end
     end
--- a/spec/rspec/mocks/stub_spec.rb
+++ b/spec/rspec/mocks/stub_spec.rb
@@ -285,6 +285,25 @@ module RSpec
           expect(@class.existing_class_method).to eq(:original_value)
         end
 
+        it "restores existing aliased module_function methods" do
+          m = Module.new do
+            def mkdir_p
+              :mkdir_p
+            end
+            module_function :mkdir_p
+
+            alias mkpath mkdir_p
+
+            module_function :mkpath
+          end
+
+          allow(m).to receive(:mkpath) { :stub_value }
+          allow(m).to receive(:mkdir_p) { :stub_value }
+          reset m
+          expect(m.mkpath).to eq(:mkdir_p)
+          expect(m.mkdir_p).to eq(:mkdir_p)
+        end
+
         it "restores existing private class methods" do
           # See bug reports 8302 and 7805
           allow(@class).to receive(:existing_private_class_method) { :stub_value }


lib/rspec/mocks/instance_method_stasher.rb
```diff
@@ -128,7 +128,7 @@ def method_owned_by_klass?
         # Hence, we verify that the owner actually has the method defined.
         # If the given owner does not have the method defined, we assume
         # that the method is actually owned by @klass.
-        owner == @klass || !(method_defined_on_klass?(owner))
+        owner == @klass || owner == @object || !(method_defined_on_klass?(owner))
```
JonRowe: So this looks like an extension of the bug reported on line #112, would you mind just confirming what owner looks like in your example, and then elaborate with a comment for future reference.

JonRowe:
Nice find, thank you, would you be able to provide just a bit more assistance and then I'll get this merged.

lib/rspec/mocks/instance_method_stasher.rb
segiddins: added a comment :)

myronmarston:
Given this works around a 1.8.7 bug and we will drop support for 1.8.7 in RSpec 4, it would be nice if this part of the conditional was applied only on 1.8.7 so we don't keep it indefinitely.

JonRowe:
@myronmarston I was thinking the extra comment would be sufficient for us to remove it in 4.

segiddins:
Or would you prefer something like:

```diff
diff --git a/lib/rspec/mocks/instance_method_stasher.rb b/lib/rspec/mocks/instance_method_stasher.rb
index 8a113db6..679bd0d1 100644
--- a/lib/rspec/mocks/instance_method_stasher.rb
+++ b/lib/rspec/mocks/instance_method_stasher.rb
@@ -114,6 +114,8 @@ module RSpec
         # `#<MyClass:0x007fbb94e3cd10>`, rather than the expected `MyClass`.
         owner = owner.class unless Module === owner
 
+        return true if owner == @klass
+
         # On some 1.9s (e.g. rubinius) aliased methods
         # can report the wrong owner. Example:
         # class MyClass
@@ -128,7 +130,9 @@ module RSpec
         # Hence, we verify that the owner actually has the method defined.
         # If the given owner does not have the method defined, we assume
         # that the method is actually owned by @klass.
-        #
+
+        return true unless method_defined_on_klass?(owner)
+
         # On 1.8, aliased methods can also report the wrong owner. Example:
         # module M
         #   def a; end
@@ -138,7 +142,9 @@ module RSpec
         # end
         # The owner of M.b is the raw Module object, instead of the expected
         # singleton class of the module
-        owner == @klass || owner == @object || !(method_defined_on_klass?(owner))
+        return true if RUBY_VERSION < '1.9' && owner == @object
+
+        false
       end
     end
   end
```

JonRowe:
If we do modify this with a conditional, simply adding:

```diff
diff --git a/lib/rspec/mocks/instance_method_stasher.rb b/lib/rspec/mocks/instance_method_stasher.rb
index 8a113db6..679bd0d1 100644
--- a/lib/rspec/mocks/instance_method_stasher.rb
+++ b/lib/rspec/mocks/instance_method_stasher.rb
@@ -138,7 +142,9 @@ module RSpec
         # end
         # The owner of M.b is the raw Module object, instead of the expected
         # singleton class of the module
-        owner == @klass || owner == @object || !(method_defined_on_klass?(owner))
+        return true if RUBY_VERSION < '1.9' && owner == @object
+        owner == @klass || !(method_defined_on_klass?(owner))
       end
     end
   end
```

Should suffice

segiddins:
@JonRowe done!

JonRowe:
Thanks!

PR #1147, opened 2017-03-22
Fix a test on Ruby trunk

yui-knk:
In Ruby 2.5, top-level constant look-up through a qualified constant
is not permitted.

see: https://github.com/ruby/ruby/commit/44a2576f798b07139adde2d279e48fdbe71a0148

### DIFF
--- a/spec/rspec/mocks/mutate_const_spec.rb
+++ b/spec/rspec/mocks/mutate_const_spec.rb
@@ -170,11 +170,17 @@ module RSpec
             expect(::Hash).to equal(top_level_hash)
           end
 
-          it 'does not affect the ability to access the top-level constant from nested contexts', :silence_warnings do
-            top_level_hash = ::Hash
+          # In Ruby 2.5, top-level constant look-up through a qualified constant
+          # is not permitted.
+          #
+          # see: https://github.com/ruby/ruby/commit/44a2576f798b07139adde2d279e48fdbe71a0148
+          if RUBY_VERSION.to_f < 2.5
+            it 'does not affect the ability to access the top-level constant from nested contexts', :silence_warnings do
+              top_level_hash = ::Hash
 
-            hide_const("TestClass::Hash")
-            expect(TestClass::Hash).to equal(top_level_hash)
+              hide_const("TestClass::Hash")
+              expect(TestClass::Hash).to equal(top_level_hash)
+            end
           end
         end
 


samphippen:
Hi @yui-knk, when we switch on Ruby interpreter behaviour, we typically use an RSpec module called `RubyFeatures` found [here](https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/ruby_features.rb). One of the things that's important to note is that `RubyFeatures` typically tries the behaviour, and doesn't switch on Ruby Versions e.g. [`supports_rebinding_module_methods`](https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/ruby_features.rb#L134-L144). Would you mind opening a pull requests which attempts to detect this qualified lookup behaviour switch on `rspec-support`?

yui-knk:
@samphippen Thanks for your advice. I created [PR](https://github.com/rspec/rspec-support/pull/310). But some test cases are broken. Do you have any idea to suppress these warnings?

myronmarston:
> Hi @yui-knk, when we switch on Ruby interpreter behaviour, we typically use an RSpec module called RubyFeatures found here.

We do that for our feature detection in our `lib` code, but if it's just to conditionally define a spec, it seems like overkill, IMO.  I think switching on ruby version in a spec is fine.

yui-knk:
I agree with myronmarston :)

spec/rspec/mocks/mutate_const_spec.rb
```diff
@@ -170,11 +170,17 @@ def change_const_value_to(value)
...
+          # is not permitted.
+          #
+          # see: https://github.com/ruby/ruby/commit/44a2576f798b07139adde2d279e48fdbe71a0148
+          if RUBY_VERSION.to_f < 2.5
+            it 'does not affect the ability to access the top-level constant from nested contexts', :silence_warnings do
+              top_level_hash = ::Hash
+
+              hide_const("TestClass::Hash")
+              expect(TestClass::Hash).to equal(top_level_hash)
+            end
```
JonRowe: Is there an alternative we should be checking on Ruby 2.5?

pirj:
Would you like to wrap this up @yui-knk ?

JonRowe:
I mean I think this must have changed as we have green builds on Rubies >= 2.5  so closing as stale for now?

PR #1148, opened 2017-03-23
Add a nice error message when calling stub_const with something other than a string

luispcosta:
Sometimes, I'm calling `.stub_const` so frequently that I don't even understand why it doesn't work only to later figuring out that I was calling it with the _actual_ constant to stub instead of a string with the namespace of the constant.

For example:

```ruby
module A
    CONST = 10
end

# Instead of calling
stub_const("A::CONST", 1)

# I sometimes do
stub_const(A::CONST, 1)

# Giving me the error message:
NoMethodError:
       undefined method `sub' for 10:Fixnum
```

Thank you for this awesome project btw!

### DIFF
--- a/lib/rspec/mocks/mutate_const.rb
+++ b/lib/rspec/mocks/mutate_const.rb
@@ -105,9 +105,9 @@ module RSpec
       #  so you can stub constants in other contexts (e.g. helper
       #  classes).
       def self.stub(constant_name, value, options={})
-        raise ArgumentError,
-              "Please provide a string as constant name" \
-              " You provided a #{constant_name.class.name}" unless constant_name.is_a?(String)
+        unless String === constant_name
+          raise ArgumentError, "`stub_const` requires a String, but you provided a #{constant_name.class.name}"
+        end
 
         mutator = if recursive_const_defined?(constant_name, &raise_on_invalid_const)
                     DefinedConstantReplacer
--- a/lib/rspec/mocks/mutate_const.rb
+++ b/lib/rspec/mocks/mutate_const.rb
@@ -105,6 +105,10 @@ module RSpec
       #  so you can stub constants in other contexts (e.g. helper
       #  classes).
       def self.stub(constant_name, value, options={})
+        raise ArgumentError,
+              "Please provide a string as constant name" \
+              " You provided a #{constant_name.class.name}" unless constant_name.is_a?(String)
+
         mutator = if recursive_const_defined?(constant_name, &raise_on_invalid_const)
                     DefinedConstantReplacer
                   else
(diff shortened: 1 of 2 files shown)


lib/rspec/mocks/mutate_const.rb
```diff
@@ -105,6 +105,10 @@ class ConstantMutator
       #  so you can stub constants in other contexts (e.g. helper
       #  classes).
       def self.stub(constant_name, value, options={})
+        raise ArgumentError,
+              "Please provide a string as constant name" \
+              " You provided a #{constant_name.class.name}" unless constant_name.is_a?(String)
```
myronmarston: I think it's confusing to put an `unless` after a multi-line expression -- instead, I think it would be more readable to put it first.

Also, we prefer doing type checks with `===`, e.g. `String === constant_name`.  The reason is that `constant_name` could be _anything_ and may not have a `is_a?` method or it may not work as we expect due to the user overriding it to behave differently -- but we expect `String.===` to always work correctly.  If you're monkeying with a core method on a core time we can't really support you.  A concrete example is a `BasicObject`. `BasicObject.new.is_a?(String)` fails with a `NoMethodError` but `String === BasicObject.new` correctly returns `false`.

Putting these together, this would be:

``` ruby
unless String === constant_name
  raise # ...
end
```

myronmarston: Also, the formatting on this can be improved, I think:

```
Please provide a string as constant name You provided a SomeClass
```

Instead, how about:

```
`stub_const` requires a String, but you provided a SomeClass
```

spec/rspec/mocks/mutate_const_spec.rb
```diff
@@ -218,6 +218,12 @@ def change_const_value_to(value)
           it_behaves_like "loaded constant stubbing", "TestClassThatDefinesSend::C"
         end
 
+        context "for a non string argument" do
+          it { expect { stub_const(10, 1) }.to raise_error(ArgumentError, /Please provide a string/i) }
+          it { expect { stub_const(nil, 1) }.to raise_error(ArgumentError, /Please provide a string/i) }
+          it { expect { stub_const(10.1, 1) }.to raise_error(ArgumentError, /Please provide a string/i) }
+        end
```
myronmarston: I don't think we need 3 specs for this--one is sufficient. And a descriptive doc string would be nice.  So, instead of a context, how about:

``` ruby
it "requires a string argument" do
  expect { ... }.to raise_error(...)
end
```

lib/rspec/mocks/mutate_const.rb
luispcosta: Damn. This is a good catch! Will fix it asap

luispcosta:
@myronmarston Changes made as requested!

myronmarston:
Thanks, @luispcosta!

PR #1167, opened 2017-07-07
Allow hash matchers to work with Hash descendants

### LINKED ISSUE #1166: Hash matchers do not work as expected with descendants of `Hash`
Currently using either `hash_including` or `hash_excluding` with an argument whose class inherits from `Hash` has unexpected behavior. While this can be most easily discovered with Rails's `ActiveSupport::HashWithIndifferentAccess`, it effects all descendants of `Hash`.

```ruby
hash_including({foo: 1}) === {foo: 1} #=> true
hash_including({foo: 1}.with_indifferent_access) === {foo: 1} #=> false
```

This is because `ArgumentMatchers.anythingize_lonely_keys` looks exclusively for hashes and treats a `Hash` descendant objects as key for an anything matcher.

```ruby
hash_including({foo: 1}) #=> #<...HashIncludingMatcher @expected={:foo=>1}>
# vs
hash_including({foo: 1}.with_indifferent_access) #=> #<...HashIncludingMatcher @expected={{"foo"=>1}=>#<...AnyArgMatcher>}>
```

Pull request open at #1167

agrberg:
Instead of checking if the object passed to ArgumentMatchers#anythingize_lonely_key _is_ strictly a Hash, check if it inherits from Hash using Case Equality inherited from `Module.===`

[fixes #1166]

### DIFF
--- a/lib/rspec/mocks/argument_matchers.rb
+++ b/lib/rspec/mocks/argument_matchers.rb
@@ -116,7 +116,7 @@ module RSpec
 
       # @private
       def self.anythingize_lonely_keys(*args)
-        hash = args.last.class == Hash ? args.delete_at(-1) : {}
+        hash = Hash === args.last ? args.delete_at(-1) : {}
         args.each { | arg | hash[arg] = AnyArgMatcher::INSTANCE }
         hash
       end
--- a/spec/rspec/mocks/hash_excluding_matcher_spec.rb
+++ b/spec/rspec/mocks/hash_excluding_matcher_spec.rb
@@ -28,6 +28,9 @@ module RSpec
             expect(hash_not_including(:a, :b, :c)).to be === { :d => 7 }
           end
 
+          it "matches against classes inheriting from Hash" do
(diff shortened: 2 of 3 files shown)


lib/rspec/mocks/argument_matchers.rb
```diff
@@ -116,7 +116,7 @@ def kind_of(klass)
 
       # @private
       def self.anythingize_lonely_keys(*args)
-        hash = args.last.class == Hash ? args.delete_at(-1) : {}
+        hash = args.last.is_a?(Hash) ? args.delete_at(-1) : {}
```
myronmarston: Please use `Hash === args.last`.  The reason being: RSpec winds up being used with all manner of user objects, including ones that do not define `is_a?` properly.  (One such example is `BasicObject`--`BasicObject.is_a?(Hash)` raises an error).  We trust that `Hash.===` is defined properly since it's a method on an object defined and owned by Ruby.

agrberg: 👍 good call! Fix coming shortly.

myronmarston:
Thanks!

PR #1181, opened 2017-09-22
Fix typo

chidoli:


myronmarston:
As the `# Typo in name` comment is trying to explain, the typo here is intentional.  One of the benefits of verifying doubles is that they catch typos like this, and this scenarios is trying to demonstrate that.

PR #1183, opened 2017-10-03
Prevent mutated args from failing have_received expectations

JonRowe:
This came up in rspec/rspec-expectations#1017 we need to duplicate args we capture for later replay onto expectations, in case they are mutated downstream. This seems like an "ok" way to fix the problem, but maybe someone can suggest a better one?

### DIFF
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -161,7 +161,19 @@ module RSpec
       # @private
       def record_message_received(message, *args, &block)
         @order_group.invoked SpecificMessage.new(object, message, args)
-        @messages_received << [message, args.map(&:dup), block]
+        @messages_received << [message, args.map { |arg| dup_if_mutable(arg) }, block]
+      end
+
+      # @private
+      def dup_if_mutable(arg)
+        case arg
+        when IO, StringIO # enumerable, but we don't want to dup it.
+          arg
+        when Enumerable, String
+          arg.dup
+        else
+          arg
+        end
       end
 
       # @private
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -161,7 +161,7 @@ module RSpec
       # @private
       def record_message_received(message, *args, &block)
         @order_group.invoked SpecificMessage.new(object, message, args)
-        @messages_received << [message, args, block]
+        @messages_received << [message, args.map(&:dup), block]
       end
 
       # @private
--- a/spec/rspec/mocks/stub_spec.rb
+++ b/spec/rspec/mocks/stub_spec.rb
@@ -110,7 +110,7 @@ module RSpec
             end
 
             def method_missing(name, *)
-              called_methods << name.to_s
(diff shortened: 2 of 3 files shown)


spec/rspec/mocks/stub_spec.rb
```diff
@@ -110,15 +110,17 @@ def called_methods
             end
 
             def method_missing(name, *)
-              called_methods << name
+              called_methods << name.to_s
```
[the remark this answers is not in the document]
JonRowe: So I could reliably compare the method name, regardless of Ruby version.

lib/rspec/mocks/proxy.rb
```diff
@@ -161,7 +161,7 @@ def has_negative_expectation?(message)
       # @private
       def record_message_received(message, *args, &block)
         @order_group.invoked SpecificMessage.new(object, message, args)
-        @messages_received << [message, args, block]
+        @messages_received << [message, args.map(&:dup), block]
```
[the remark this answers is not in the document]
JonRowe: Interesting, FYI, dup is a noop for those on 2.4.2

JonRowe:
Green bar jruby which is failing due to unrelated reasons, care to re-review @myronmarston ?

lib/rspec/mocks/proxy.rb
```diff
@@ -161,7 +161,19 @@ def has_negative_expectation?(message)
       # @private
       def record_message_received(message, *args, &block)
         @order_group.invoked SpecificMessage.new(object, message, args)
-        @messages_received << [message, args.map(&:dup), block]
+        @messages_received << [message, args.map { |arg| dup_if_mutable(arg) }, block]
+      end
+
+      # @private
+      def dup_if_mutable(arg)
```
mrageh: Do you think it's worth refactoring this method to make it clear that we only `dup`, `Enumerable` and `String`. And everything else we don't `dup`. 

In essence what I'm saying is remove the first `when IO, StringIO` and just refactor this code to say something like:

```
case arg
when Enumerable, String
  arg.dup 
else
  arg
end
```

xaviershay: Both `IO` and `StringIO` are _also_ `Enumerable` :(

mrageh: Thanks @xaviershay, I think that should be included in the commit message. 

Code comments normally expire, whereas commit messages are like living documentation for why some code was written in a certain way. Perhaps we could replace the comment with a commit message?

xaviershay: eh I don't have strong feelings, but bikeshedding fwiw.. :)

For me, commit message = explain why we made changes in the first place. Comments = explain why/how a particular bit of (small) code works. I don't see how this comment would "expire" since it's hyperlocal to the code it affects. This is also a pattern we need to use multiple times in the code base [citation needed, but I know I've been bitten by it a few times].

JonRowe: It's clear enough and matches how we do it elsewhere

PR #1207, opened 2018-03-05
Fixed mistype

panasyuk:
Fixed mistype  in declaration and usage of RSpec::Mocks::AnyInstance::StubChain::InvocationOrder constant.


xaviershay:
thanks!

review: I double checked that this constant was private (it is).

Issue #1210, opened 2018-03-13
Wrong negative expectation with once


SergiyYarinovskiy:
Hello. I have such tests:
```ruby
  class A; end
  it 'calls new' do
    expect(A).to receive(:new).once
    A.new
  end
```
It finishes successfully. Next one:
```ruby
  class A; end
  it 'calls new' do
    expect(A).not_to receive(:new)
    # A.new
  end
```
Also fine. But if I combine `.not_to` with `.once` it behaves like it expects **to receive**:
```ruby
  class A; end
  it 'calls new' do
    expect(A).not_to receive(:new).once
    # A.new
  end
```
```
     Failure/Error: expect(A).not_to receive(:new).once
     
       (A (class)).new(*(any args))
           expected: 1 time with any arguments
           received: 0 times with any arguments
```
I am not sure if this is a bug. I would expect that `.once` will be ignored and last test passes
`rspec-mocks (3.7.0)`

Thank you

myronmarston:
Thanks for reporting.  I'd consider this a bug.  I don't think we should support receive counts (e.g. `once`, `twice`, at_least(3).times`, etc) with `not_to receive`.  Want to take a stab at a fix?

SergiyYarinovskiy:
@myronmarston OK, I'll investigate :slightly_smiling_face:

SergiyYarinovskiy:
Since I don't have access to create branch in this repo, I forked it and created branch in my repo called `issue-1210`. Here is my fix proposal - https://github.com/SergiyYarinovskiy/rspec-mocks/commit/6420f56bfb9befc5f87a670bfce255eb1c3ad2ed
What is the process to create PR here?
Thanks

myronmarston:
Pushing a branch to your fork is the right way to do it.  You just need to open the PR here:

https://github.com/rspec/rspec-mocks/compare/master...SergiyYarinovskiy:issue-1210

mrageh:
I think this issue can was resolved in #1212, let's close it

myronmarston:
Yep, closing.

PR #1211, opened 2018-03-20
Fixes Issue #1210: Wrong negative expectation with once

SergiyYarinovskiy:
See details https://github.com/rspec/rspec-mocks/issues/1210

### DIFF
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -581,6 +581,7 @@ module RSpec
         end
 
         def set_expected_received_count(relativity, n)
+          return if negative?
           @at_least = (relativity == :at_least)
           @at_most  = (relativity == :at_most)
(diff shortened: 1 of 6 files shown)


lib/rspec/mocks/message_expectation.rb
```diff
@@ -581,6 +581,7 @@ def raise_already_invoked_error_if_necessary(calling_customization)
         end
 
         def set_expected_received_count(relativity, n)
+          return if negative?
```
JonRowe: Ideally it should warn when this happens...

myronmarston: Or should it raise an error?  Consider that it's invalid to set an expected count on a negative expectation, and has never worked (as demonstrated by #1210).

JonRowe: Would be equally happy in that scenario too.

Issue #1216, opened 2018-04-04
found typos in document.


dizy64:
There is nothing special about it.

I read a document and found a typographical error.

https://github.com/rspec/rspec-mocks/blob/master/features/verifying_doubles/partial_doubles.feature#L28

saave -> save

JonRowe:
It'd be great if you could open a PR to fix this

JonRowe:
Sorry! This typo is actually deliberate, my apologies I should have checked better, thank you very much for taking the time to help us out.

PR #1217, opened 2018-04-04
fixed typographical error of document

dizy64:
This is my first commit for open source.
I found a typo in the document and corrected it.

ISSUE: #1216 

Thank you.

### DIFF
--- a/features/verifying_doubles/partial_doubles.feature
+++ b/features/verifying_doubles/partial_doubles.feature
@@ -25,7 +25,7 @@ Feature: Partial doubles
       RSpec.describe '#save_user' do
         it 'renders message on success' do
           user = User.new
-          expect(user).to receive(:saave).and_return(true) # Typo in name
+          expect(user).to receive(:save).and_return(true) # Typo in name
           expect(save_user(user)).to eq("saved!")
         end
       end


JonRowe:
Ah actually this typo is deliberate as it proves a failure works. Thanks anyway though :)

mikegee:
I don't think this is the first PR opened to fix this typo. Perhaps it could be rewritten to avoid confusion?

```ruby
expect(user).to receive(:obviously_incorrect_method_name_here).and_return(true)
```

JonRowe:
@dizy64 Would you like to tackle that? (Renaming the method name to make it more obvious it's a deliberate typo?)

myronmarston:
Hmm, I thought the `# Typo on name` comment would be clear enough.  `:saave` has the advantage of being closer to the kind of typo you'd actually have in practice unlike `obviously_incorrect_method_name_here`.

mikegee:
@myronmarston I agree with your point there. Unfortunately, we have some evidence that people are confused by it. More concerning are the readers leaving confused without opening an issue or PR.

My particular suggestion may not be the best solution, but I’m convinced we should consider the options. 

Here’s that other PR from a few months ago: https://github.com/rspec/rspec-mocks/pull/1181

dizy64:
I am not good at English.
I often read most of my documents in code only.
So I did not understand it properly.

Sometimes there is code that can not be expressed as intended. This is also the case.

It was a Pull Request that started with my mistake, but it was an enjoyable experience for me. :)

dizy64:
If the opportunity, I would like to try again challenged.
But it seems the issue needs a lot of discussion.

Issue #1213, opened 2018-03-29
Stubbing prepended methods no longer works under Ruby 2.5.1


jhottenstein:
We recently upgraded to ruby 2.5.1 and had a regression in our test suite.  I isolated the failure to expecting a method to be called on a partial class double and wrote the following test case:

```
    RSpec.describe "A partial class mock" do
      it 'allows prepended methods to be mocked' do
        klass = Class.new do
          def self.meth1
            'hi'
          end
        end

        mod = Module.new do
          def meth1
            meth2
          end

          def meth2
            'hello'
          end
        end
        klass.singleton_class.prepend(mod)
        expect(klass).to receive(:meth2).and_call_original
        klass.meth1
      end
    end
```
```
$ rbenv local 2.5.0
$ bundle exec rspec spec/rspec/mocks/partial_double_spec.rb --example "A partial class mock allows prepended methods to be mocked"
Run options: include {:full_description=>/A\ partial\ class\ mock\ allows\ prepended\ methods\ to\ be\ mocked/}

Randomized with seed 51084

A partial class mock
  allows prepended methods to be mocked

Finished in 0.00733 seconds (files took 0.17655 seconds to load)
1 example, 0 failures

Randomized with seed 51084
```
```
$ rbenv local 2.5.1
$ bundle exec rspec spec/rspec/mocks/partial_double_spec.rb --example "A partial class mock allows prepended methods to be mocked"
Run options: include {:full_description=>/A\ partial\ class\ mock\ allows\ prepended\ methods\ to\ be\ mocked/}

Randomized with seed 20049

A partial class mock
  allows prepended methods to be mocked (FAILED - 1)

Failures:

  1) A partial class mock allows prepended methods to be mocked
     Failure/Error: object_singleton_class.__send__(@original_visibility, method_name)

     NameError:
       undefined method `meth2' for class `#<Class:#<Class:0x00007f82cd197538>>'
       Did you mean?  method
     # ./lib/rspec/mocks/method_double.rb:108:in `public'
     # ./lib/rspec/mocks/method_double.rb:108:in `restore_original_visibility'
     # ./lib/rspec/mocks/method_double.rb:87:in `restore_original_method'
     # ./lib/rspec/mocks/method_double.rb:118:in `reset'
     # ./lib/rspec/mocks/proxy.rb:319:in `block in reset'
     # ./lib/rspec/mocks/proxy.rb:319:in `each_value'
[...]

PR #1218, opened 2018-04-10
Fix stubbing prepended only methods

godfat:
Previously, we're assuming the method must be defined in the
singleton class. However this is not always true. Whenever the
method was only defined in the prepended module, then it's not
defined in the singleton class. We need to find the owner of
the method instead, which is the prepended module.

Closes https://github.com/rspec/rspec-mocks/issues/1213

### DIFF
--- a/lib/rspec/mocks/instance_method_stasher.rb
+++ b/lib/rspec/mocks/instance_method_stasher.rb
@@ -62,7 +62,12 @@ module RSpec
         # `#<MyClass:0x007fbb94e3cd10>`, rather than the expected `MyClass`.
         owner = owner.class unless Module === owner
 
-        owner == @klass || !(method_defined_on_klass?(owner))
+        owner == @klass ||
+          # When `extend self` is used, and not under `allow_any_instance_of`
+          # nor `expect_any_instance_of`.
+          (owner.singleton_class == @klass &&
+            !Mocks.space.any_instance_recorder_for(owner, true)) ||
+          !(method_defined_on_klass?(owner))
       end
     end
   end
--- a/lib/rspec/mocks/method_double.rb
+++ b/lib/rspec/mocks/method_double.rb
@@ -85,8 +85,11 @@ module RSpec
         return unless @method_is_proxied
 
         remove_method_from_definition_target
-        @method_stasher.restore if @method_stasher.method_is_stashed?
-        restore_original_visibility
+
+        if @method_stasher.method_is_stashed?
+          @method_stasher.restore
+          restore_original_visibility
+        end
 
         @method_is_proxied = false
       end
@@ -104,10 +107,7 @@ module RSpec
 
       # @private
       def restore_original_visibility
-        return unless @original_visibility &&
-          MethodReference.method_defined_at_any_visibility?(object_singleton_class, @method_name)
-
-        object_singleton_class.__send__(@original_visibility, method_name)
+        method_owner.__send__(@original_visibility, @method_name)
       end
 
       # @private
@@ -247,6 +247,12 @@ module RSpec
         RSpecPrependedModule.new.tap do |mod|
           @object.singleton_class.prepend mod
         end
+      end
+
+      def method_owner
+        @method_owner ||=
+          # We do this because object.method might be overridden.
+          ::RSpec::Support.method_handle_for(object, @method_name).owner
       end
 
       def remove_method_from_definition_target
--- a/spec/rspec/mocks/stub_spec.rb
+++ b/spec/rspec/mocks/stub_spec.rb
@@ -138,6 +138,10 @@ module RSpec
           def value
             "#{super}_prepended".to_sym
           end
+
+          def value_without_super
+            :prepended
+          end
         end
 
         it "handles stubbing prepended methods" do
@@ -163,6 +167,15 @@ module RSpec
           expect(object.value).to eq :original_prepended
           allow(object).to receive(:value) { :stubbed }
           expect(object.value).to eq :stubbed
+        end
+
+        it "handles stubbing prepending methods that were only defined on the prepended module" do
+          object = Object.new
+          object.singleton_class.send(:prepend, ToBePrepended)
+
+          expect(object.value_without_super).to eq :prepended
+          allow(object).to receive(:value_without_super) { :stubbed }
+          expect(object.value_without_super).to eq :stubbed
         end
 
         it 'does not unnecessarily prepend a module when the prepended module does not override the stubbed method' do
@@ -346,6 +359,21 @@ module RSpec
 
           allow(mod).to receive(:hello) { :stub }
           reset mod
+
+          expect(mod.hello).to eq(:hello)
+        end
+
+        it "correctly restores from allow_any_instance_of for self extend" do
+          mod = Module.new {
+            extend self
+            def hello; :hello; end
+          }
+
+          allow_any_instance_of(mod).to receive(:hello) { :stub }
+
+          expect(mod.hello).to eq(:stub)
+
+          reset_all
 
           expect(mod.hello).to eq(:hello)
         end
--- a/Changelog.md
+++ b/Changelog.md
@@ -21,6 +21,7 @@ Bug Fixes:
 
 * Support keyword argument semantics when constraining argument expectations using
   `with` on Ruby 3.0+ (Yusuke Endoh, #1394)
+* Fix stubbing of prepended-only methods. (Lin Jen-Shin, #1218)
 
 ### 3.10.2 / 2021-01-27
 [Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.10.1...v3.10.2)


godfat:
@xaviershay Thanks for the support. I realized that there's still a failing test:

`spec/rspec/mocks/partial_double_spec.rb:604`

Which is:

``` ruby
it 'uses the method signature from `#initialize` for arg verification' do
  pending "Failing on JRuby due to https://github.com/jruby/jruby/issues/2565" if RSpec::Support::Ruby.jruby?

  subclass = Class.new(klass) do
    private_class_method :new
  end

  prevents(/arguments/) { allow(subclass).to receive(:new).with(1) }
  allow(subclass).to receive(:new).with(1, 2)
end
```

It's actually not failing, but it would mark the `new` method as `private` for the other class, causing issues. I am still figuring out how to deal with this edge case.

lib/rspec/mocks/method_double.rb
```diff
@@ -47,6 +47,13 @@ def object_singleton_class
         class << @object; self; end
       end
 
+      # @private
+      def method_owner
+        if Object.instance_method(:respond_to?).bind(object).call(@method_name, true)
+          Object.instance_method(:method).bind(object).call(@method_name).owner
+        end
+      end
```
myronmarston: A couple things...

* It looks like this is only used from `restore_original_visibility`, below.  Can it be actually `private` instead of just being hidden from the YARD docs via `@private`?  We use `@private` only for methods that need to be publicly accessible for other parts of rspec-mocks to be able to call, but that we consider to not be part of our public API.
* What's the reason for the `instance_method(...).bind(...).call(...)` business?  (It'd probably be obvious if I had worked in rspec-mocks recently, but I haven't).  Alternately, we do have `RSpec::Support.method_handle_for(object, method_name).call(*args)` that you could use that does something similar--it might be worth using here?

lib/rspec/mocks/method_double.rb
```diff
@@ -102,10 +109,10 @@ def show_frozen_warning
 
       # @private
       def restore_original_visibility
-        return unless @original_visibility &&
-          MethodReference.method_defined_at_any_visibility?(object_singleton_class, @method_name)
+        return unless method_owner && @original_visibility &&
+          MethodReference.method_defined_at_any_visibility?(method_owner, @method_name)
 
-        object_singleton_class.__send__(@original_visibility, method_name)
+        method_owner.__send__(@original_visibility, @method_name)
```
myronmarston: This method calls `method_owner` 3 different times.  The value shouldn't change between calls, and it seems inefficient to do that logic 3 times.  Can you store the value of `method_owner` in a variable so it's called only once?

spec/rspec/mocks/stub_spec.rb
```diff
@@ -154,6 +158,15 @@ def object.value; :original; end
           expect(object.value).to eq :stubbed
         end
 
+        it "handles stubbing prepended methods without originally defined on an object's singleton class" do
```
myronmarston: I don't understand what "without originally defined" means.  Do you mean this?

``` ruby
it "handles stubbing prepending methods that were only defined on the prepended module"
```

myronmarston (dismissed):
Thanks @godfat!

myronmarston:
Woops, with the slow wi-fi on this bus I didn't see your comment there @xaviershay as I was reviewing :).

lib/rspec/mocks/method_double.rb
godfat: Yes, good calls, thanks.

As for `instance_method(...).bind(...).call(...)`, it's needed because the method `method` and `respond_to?` might be stubbed, and when that happened obviously it's not working.

I tried `RSpec::Support.method_handle_for` but then `spec/rspec/mocks/and_call_original_spec.rb:257` would fail with:

```
  2) and_call_original on a partial double that overrides #method works for a singleton method
     Failure/Error: RSpec::Support.method_handle_for(object, :method).call(@method_name).owner

     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # /Users/godfat/project/fork/rspec-mocks/lib/rspec/mocks/method_double.rb:56:in `method'
```

All the other tests except the still failing one is passed. Any idea? Maybe it's a bug in `RSpec::Support.method_handle_for`?

spec/rspec/mocks/stub_spec.rb
godfat: I guess so. Sorry for my English :P Will update after I settled all the tests.

godfat:
I pushed the current progress. Now the only failing test is:

``` ruby
it "correctly restores the visibility of methods whose visibility has been tweaked on the singleton class" do
  # hello is a private method when mixed in, but public on the module
  # itself
  mod = Module.new {
    extend self
    def hello; :hello; end

    private :hello
    class << self; public :hello; end;
  }

  expect(mod.hello).to eq(:hello)

  allow(mod).to receive(:hello) { :stub }
  reset mod

  expect(mod.hello).to eq(:hello)
end
```

Due to:

```
  1) A method stub #rspec_reset correctly restores the visibility of methods whose visibility has been tweaked on the singleton class
     Failure/Error: expect(mod.hello).to eq(:hello)

     NoMethodError:
       private method `hello' called for #<Module:0x00007f832828b560>
     # ./spec/rspec/mocks/stub_spec.rb:352:in `block (3 levels) in <module:Mocks>'
```

It's not passing because `InstanceMethodStasher#method_owned_by_klass?` wasn't properly detecting that the method was defined by its own. This is an edge case due to `extend self`, which makes it think that it's not defined on its own, but it actually is...

That method was a bit complex and I am still thinking a general way which could cover this edge case.

lib/rspec/mocks/method_double.rb
```diff
@@ -83,8 +83,11 @@ def restore_original_method
         return unless @method_is_proxied
 
         remove_method_from_definition_target
-        @method_stasher.restore if @method_stasher.method_is_stashed?
-        restore_original_visibility
+
+        if @method_stasher.method_is_stashed?
+          @method_stasher.restore
+          restore_original_visibility
```
godfat: Actually, we could remove all the checks in `restore_original_visibility` when this is only called when `@method_stasher.method_is_stashed?`

godfat:
Oh, Ruby 2.3.7+ passed but all the other older Rubies broke. We're really supporting very old rubies. I'll revisit this later as I don't have older rubies installed locally.

Any help would be appreciated though, as I am personally not too interested in older rubies. `singleton_class` could be easy to fix, but `method_owned_by_klass?` and `Object.instance_method(:method).bind(object)` could be tough.

I'll not force push too much from now as I don't want to pollute the related issue anymore.

godfat:
Is there any plan to drop support for older Rubies?

JonRowe:
In RSpec 4 we will drop some but not necessarily all older Rubies.

godfat:
@JonRowe Thanks. What about 2.2.10? According to the [test](https://travis-ci.org/rspec/rspec-mocks/builds/364730977), this works on 2.3.7+

JonRowe:
Given the ruby 2 branch is mostly trouble free it is unlikely to be dropped

godfat:
Thanks. I could probably see if I could fix one more older ruby, but if that won't fix even older rubies, I think I am not looking into them then...

lib/rspec/mocks/method_double.rb
godfat: I updated `method_owner` to memorize as `@method_owner` now.

lib/rspec/mocks/method_double.rb
```diff
@@ -261,6 +261,15 @@ def definition_target
 
     private
 
+      def method_owner
+        @method_owner ||=
+          if object.kind_of?(Object)
+            Object.instance_method(:method).bind(object).call(@method_name).owner
```
godfat, line 267: This condition fixes for Ruby 1.9.2, 1.9.3, 2.0.0, 2.1.0, and 2.2.10. This is probably because `SimpleDelegator` is a `BasicObject` and cannot bind to `Object` in those cases. Not sure why it did work for Ruby 2.3+ though, even when that's a good thing.

pirj: Do you mind adding a code comment to clarify that?
We'll [drop support for old Rubies this year](https://github.com/rspec/rspec/issues/61#issuecomment-687290045), so maybe this workaround won't be needed anymore.

godfat: Yes, I actually added a comment after I left this comment, and after seeing your comment I tried to make the comment even more clear, so thanks for asking.

In https://github.com/rspec/rspec-mocks/pull/1218#issuecomment-392540214 @JonRowe also said that Ruby 2.2 is unlikely to be dropped so we probably still need this patch for a long time.

On the other hand, I really can't get Ruby 1.8.7 work so at least that should be dropped. Now I really want to get this merged so I also started looking into how to get JRuby pass, and I am running this version of JRuby and everything actually just passed:

```
jruby 9.2.13.0 (2.5.7) 2020-08-03 9a89c94bcc OpenJDK 64-Bit Server VM 14.0.2+12 on 14.0.2+12 +jit [linux-x86_64]
```

If we're ok to upgrade to newer JRuby, I hope we can just do that because it's likely older JRuby bugs and I really don't want to add even more workaround :(

lib/rspec/mocks/instance_method_stasher.rb
```diff
@@ -139,7 +139,12 @@ def method_owned_by_klass?
         # The owner of M.b is the raw Module object, instead of the expected
         # singleton class of the module
         return true if RUBY_VERSION < '1.9' && owner == @object
-        owner == @klass || !(method_defined_on_klass?(owner))
+
+        owner == @klass ||
+          # When `extend self` is used, and not under any instance of
+          (owner.singleton_class == @klass &&
```
godfat, line 145: I think it's ok to use `singleton_class` now because we can drop Ruby 1.8.7 and all later Rubies do support it.

I can't get Ruby 1.8.7 to work anyway... even with `class << owner; self; end` there are still other wild failures in Ruby 1.8.7.

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -604,8 +604,6 @@ class << self
 
         context "on a class with a private `new`" do
           it 'uses the method signature from `#initialize` for arg verification' do
-            pending "Failing on JRuby due to https://github.com/jruby/jruby/issues/2565" if RSpec::Support::Ruby.jruby?
```
godfat, line 607: jruby 9.2.13.0 can pass this so removing it, otherwise RSpec is complaining this is actually passing.

spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
```diff
@@ -165,8 +165,6 @@ class << self
 
         context "on a class with a private `new`" do
           it 'uses the method signature from `#initialize` for arg verification' do
-            pending "Failing on JRuby due to https://github.com/jruby/jruby/issues/2565" if RSpec::Support::Ruby.jruby?
```
godfat, line 168: jruby 9.2.13.0 can pass this so removing it, otherwise RSpec is complaining this is actually passing.

spec/rspec/mocks/stub_spec.rb
```diff
@@ -350,6 +363,21 @@ class << self; public :hello; end;
...
+          mod = Module.new {
+            extend self
+            def hello; :hello; end
+          }
+
+          allow_any_instance_of(mod).to receive(:hello) { :stub }
+
+          expect(mod.hello).to eq(:stub)
+
+          reset_all
```
godfat, line 376: Probably due to `allow_any_instance_of`, this has to be `reset_all` and `reset mod` doesn't clean everything.

spec/rspec/mocks/stub_spec.rb
```diff
@@ -350,6 +363,21 @@ class << self; public :hello; end;
           expect(mod.hello).to eq(:hello)
         end
 
+        it "correctly restores from allow_any_instance_of for self extend" do
```
godfat, line 366: This test can catch the need for `!Mocks.space.any_instance_recorder_for(owner, true)` in the other condition. It was discovered in https://gitlab.com/gitlab-org/gitlab/-/merge_requests/44417#note_429658296 thanks to @engwan !

lib/rspec/mocks/instance_method_stasher.rb
JonRowe: > I think it's ok to use singleton_class now because we can drop Ruby 1.8.7

Ruby 1.8.7 support will not be dropped until RSpec 4.

godfat: > Ruby 1.8.7 support will not be dropped until RSpec 4.

That is fine as long as it's scheduled. Before then we'll just pick this patch into our codebase.

JonRowe: Just be aware that this won't be merged at this time without Ruby 1.8.7 support.

godfat:
Guess I'll give up on JRuby 9.1.7.0 for now. I don't understand the error which doesn't even have backtrace in Ruby, but Java, and it looks like JRuby internal bug. Not that we can't fix that but 9.2.13.0 is all green locally for me, and Travis doesn't like 9.2.13.0, which got stuck forever upon installing Rubygems 🤷 

So essentially what's broken are:

* JRuby 9.1.7.0 due to JRuby bug
* Any Ruby 1.8.7 families (REE, JRuby 1.8 mode, etc). Too many issues, didn't try to look deep

And we're good for:

* JRuby 1.7
* JRuby 9.2.13.0
* Ruby 1.9.2 + 1.9.3
* Ruby 2.0.0
* Ruby 2.1.0
* Ruby 2.2.10
* Ruby 2.3.8
* Ruby 2.4.10
* Ruby 2.5.8
* Ruby 2.6.6
* Ruby 2.7.1
* Ruby 3 preview
* Ruby HEAD (trunk?)

headius:
I commented on jruby/jruby#6449 but repost some info here...

There's no strong reason to keep trying to support JRuby 9.1.x. Every version lower than 9.2 is considered EOL and we do not recommend using them due to many missing compatibility and security fixes.

The stall when running 9.2.13 *might* be due to the build machine running out of entropy. If so, installing the `haveged` service might fix it.

If it's not entropy, I'm out of ideas for the moment. Try that and I'll keep thinking about the problem.

JonRowe:
Our versioning policy dictates we don't drop support for Ruby versions except in major versions, with JRuby it gets a bit more complicated because JRuby versions don't match up with Ruby versions which are enforced by the gemspec... so I guess we could shift the build provided we still test the older modes?

Travis seems to be having issues with out builds in general of late, I'm wondering if switching to github actions would solve that?

godfat:
For a quick update, I tried installing `haveged` service and the result for JRuby 9.2.13.0 was the same. More can be found at https://github.com/jruby/jruby/issues/6449#issuecomment-718722793

> so I guess we could shift the build provided we still test the older modes?

@JonRowe That sounds great to me. Actually, other than 1.8 mode, it's just JRuby 9.1.7.0 didn't work, and which we should be able to just upgrade to 9.2.13.0 because they're sharing the same mode?

> Travis seems to be having issues with out builds in general of late, I'm wondering if switching to github actions would solve that?

I never used GitHub Actions so I cannot comment on that. I tried to move from travis-ci.org to travis-ci.com yesterday for my other project, and sadly it's even worse, now it just got stuck forever.

I am guessing they're in the process of migration so things are not stabilizing yet. This is so frustrating and I am not sure if waiting for their migration will work or not. I personally will probably try to migrate all my CI to GitLab (disclaimer: I am employed by GitLab), and which has free pipeline quota I am not sure if it would be enough for my personal projects. Maybe yes maybe not, but that would be my first attempt for an alternative. (and being able to just use Docker images can give me better confidence about reproducibility to trouble shooting)

On the other hand, [open source projects can request Gold accounts](https://about.gitlab.com/solutions/open-source/join/) which I assume it'll have much larger pipeline quota. GitLab is certainly using RSpec and that's also why I am trying to make this patch here, so I would be happy to connect if RSpec would like something from GitLab.

I am also assuming that we can just use GitLab CI instead of the whole GitLab, and integrate GitLab CI with GitHub, just like how Travis was integrated with GitHub. I don't know how feasible is that and it's just my assumption though.

I am not trying to push anything, just throwing out ideas because I got so frustrated with Travis for the last few days. I hope this is not offending 🙏

.travis.yml
```diff
@@ -1,8 +1,6 @@
 # This file was generated on 2020-09-28T21:53:54+02:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
-# In order to install old Rubies, we need to use old Ubuntu distibution.
-dist: trusty
```
godfat, line 5: This is the line where it got JRuby 9.2.13.0 stuck. I have no idea why but assuming this is only needed for Ruby 1.8 and this merge request is not aiming for Ruby 1.8 that should be fine.

I want to get this ready except anything related to Ruby 1.8. I know we wouldn't edit this file directly but this is a demonstration that everything beside that should work.

godfat:
Regular tests passed with JRuby 9.2.13.0, but now it's giving an error on running `cucumber` with JRuby 9.2.13.0:

```
/home/travis/build/rspec/rspec-mocks/bin/cucumber
uninitialized constant Aruba::Platforms::WindowsPlatform::FFI (NameError)
org/jruby/RubyModule.java:3760:in `const_missing'
/home/travis/build/rspec/bundle/jruby/2.5.0/gems/aruba-0.14.14/lib/aruba/platforms/windows_platform.rb:23:in `match?'
org/jruby/RubyEnumerable.java:690:in `find'
/home/travis/build/rspec/bundle/jruby/2.5.0/gems/aruba-0.14.14/lib/aruba/platform.rb:13:in `<module:Aruba>'
/home/travis/build/rspec/bundle/jruby/2.5.0/gems/aruba-0.14.14/lib/aruba/platform.rb:11:in `<main>'
org/jruby/RubyKernel.java:974:in `require'
/home/travis/.rvm/rubies/jruby-9.2.13.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:72:in `require'
/home/travis/build/rspec/bundle/jruby/2.5.0/gems/aruba-0.14.14/lib/aruba/api.rb:7:in `<main>'
org/jruby/RubyKernel.java:974:in `require'
/home/travis/.rvm/rubies/jruby-9.2.13.0/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:72:in `require'
/home/travis/build/rspec/bundle/jruby/2.5.0/gems/aruba-0.14.14/lib/aruba/cucumber.rb:3:in `<main>'
```

I have no idea why and it looks like a bug in `aruba` or `cucumber`, so I guess I'll just leave it for now.

godfat:
Actually, that is an FFI issue: https://github.com/ffi/ffi/pull/763/files#r514498961

I guess have to try to downgrade it.

godfat:
@xaviershay @JonRowe I just rebased this without changing anything, and surprise! All tests passed on GitHub Actions?? I have no idea why but I really didn't change anything.

Could any of you review and merge if there's no other concerns?

/cc @texpert

pirj:
@godfat Good news indeed! Thanks for moving this forward.

Travis MRI 1.8.7/1.9.2/ree/JRuby 1.7 build failure is due to:
```
$ gem install bundler -v '< 2'

ERROR:  Could not find a valid gem 'bundler' (< 2), here is why:

          Unable to download data from https://rubygems.org/ - hostname was not match with the server certificate (https://rubygems.org/specs.4.8.gz)
```
or
```
          Unable to download data from https://rubygems.org/ - Received fatal alert: handshake_failure (https://api.rubygems.org/specs.4.8.gz)
```
or
```
          Unable to download data from https://rubygems.org/ - hostname was not match with the server certificate (https://rubygems.org/specs.4.8.gz)
```

We're going to release RSpec 4 soon after Ruby 3 release (end of December, or beginning of January 2021), which will drop support for older Rubies. Do you mind to delay this PR for a month, and merge it totally clean to become part of 4.0.1?

BTW `@xaviershay` might not be following RSpec development that closely lately. It's ok not to carbon copy anyone, we're looking closely at all issue/pr comments anyway.

godfat:
> Do you mind to delay this PR for a month, and merge it totally clean to become part of 4.0.1?

No. It's fine to delay it. The schedule sounds good to me.

> BTW `@xaviershay` might not be following RSpec development that closely lately. It's ok not to carbon copy anyone, we're looking closely at all issue/pr comments anyway.

Thanks for letting me know! Great to know that you're looking at them and anyone might jump in :D That's terrific.

fledman:
@godfat @JonRowe @pirj 
since it has been 2 months, I assume the RSpec 4 release was delayed
is there a new timeline to merge this?

pirj:
You're right, we were overly-optimistic about the amount of work for the RSpec 4 release, and it takes longer.

We've merged a couple of things while working on the RSpec 4 release and would have to merge anyway, so I personally have no objections on merging this now. @JonRowe, @benoittgt WDYT?

pirj:
I've rebased this to include the changes we've made to the CI. Expect the build to be all green this time.

JonRowe:
@pirj this PR doesn't support 1.8.7 so it can't be merged to `main` at this time, it can be merged to `4-0-dev` although I'm uncomfortable about bug fixes only existing there.

benoittgt:
I am in favor of merging this to 4-0-dev and move on.

fledman:
@benoittgt @godfat @JonRowe @pirj 
so is this ready to merge into `4-0-dev` then?

JonRowe:
If it was rebased from that branch and build passed, yes.

godfat:
I rebased against `4-0-dev` but there's no CI running.

pirj:
Push-forced to trigger CI. Running.

lib/rspec/mocks/method_double.rb
```diff
@@ -249,6 +249,21 @@ def new_rspec_prepended_module
         end
       end
 
+    private
```
pirj, line 252: Agree with RuboCop, there's yet another `private` in line 210.

lib/rspec/mocks/method_double.rb
```diff
@@ -249,6 +249,19 @@ def new_rspec_prepended_module
         end
       end
 
+      def method_owner
+        @method_owner ||=
+          if Object === object
+            # This works in Ruby 2.3+, so we can just use this after we
+            # drop Ruby 2.2 and earlier.
```
pirj, line 256: We've dropped Ruby 2.2 in the base `4-0-dev` branch.

pirj: Maybe this will help to resolve [the conversation about `method_handle_for`](https://github.com/rspec/rspec-mocks/pull/1218#discussion_r180487361).

fledman: FYI @godfat
you can remove the fallback logic for older rubies

godfat: Thanks, I updated to use the one works in Ruby 2.3+

godfat:
It doesn't seem like CI is running unless some maintainers approve this. Let me know what else needs to be done because I am not clear what we're missing.

pirj:
Approved, running.

godfat:
Thanks. Looks like it's all green. Do we need anything else?

lib/rspec/mocks/instance_method_stasher.rb
```diff
@@ -62,7 +62,11 @@ def method_owned_by_klass?
         # `#<MyClass:0x007fbb94e3cd10>`, rather than the expected `MyClass`.
         owner = owner.class unless Module === owner
 
-        owner == @klass || !(method_defined_on_klass?(owner))
+        owner == @klass ||
+          # When `extend self` is used, and not under any instance of
+          (owner.singleton_class == @klass &&
```
pirj, line 67: Is this comment complete?

lib/rspec/mocks/method_double.rb
```diff
@@ -249,6 +249,11 @@ def new_rspec_prepended_module
         end
       end
 
+      def method_owner
+        @method_owner ||=
+          Object.instance_method(:method).bind(object).call(@method_name).owner
```
pirj, line 254: It took me a while to figure out why we have to use `Object#method` instead of using `object.method` directly. For the future reference: it is to tolerate `method` method from being overridden, like here:
```ruby
let(:request_klass) { Struct.new(:method, :url) }
```

lib/rspec/mocks/method_double.rb
```diff
@@ -249,6 +249,11 @@ def new_rspec_prepended_module
         end
       end
 
+      def method_owner
+        @method_owner ||=
```
pirj, line 253: Does it affect the performance?

pirj (approved):
Thank you for all the hard work!

@JonRowe Should we merge to 4-0-dev, or wait until we release 4.0, and release this in 4.0.1?

lib/rspec/mocks/instance_method_stasher.rb
godfat: Yes, by "any instance of" I meant `allow_any_instance_of` or `expect_any_instance_of`. I'll update it.

lib/rspec/mocks/method_double.rb
godfat: Yes, I added a comment.

lib/rspec/mocks/method_double.rb
godfat: It's memoized so I don't think we'll see anything noticeable in practice. If we disregard memoization and run simple benchmark, `object.method(method_name).owner` is about 3 times faster:

```
Warming up --------------------------------------
 object.method.owner   261.517k i/100ms
       bind and call    91.219k i/100ms
Calculating -------------------------------------
 object.method.owner      2.661M (± 0.9%) i/s -     13.337M in   5.011938s
       bind and call    914.184k (± 1.0%) i/s -      4.652M in   5.089434s
```

From running this:

``` ruby

require 'benchmark/ips'

Benchmark.ips do |x|
  object = Object.new

  x.report('object.method.owner') do
    object.method(:to_s).owner
  end

  x.report('bind and call') do
    Object.instance_method(:method).bind(object).call(:to_s).owner
  end
end
```

Well, or we can remove the test and calling out that do not override `method`, which I fully support :P

pirj: I'm clearly on your side with you that one shouldn't override `Kernel`'s internals with unrelated application-level implementations. However, someone's code may break with a weird message, and they will open a ticket, and we'll have to address it, since even though this might not be a good style to override `method`, it's not our call to enforce that in any way.

godfat: Yeah, of course, that's why I was half joking there :)

Half joking, half serious, because there are a lot of other methods we're relying on, and of course we won't want to be so defensive to do it this way for most of the built-in methods? (or a rule of thumb is, whenever someone is filing a ticket 🤷 This is also relying on someone isn't trying to break RSpec intentionally)

I see `method` or `send` is more commonly overridden with a completely different behaviour (if it's the same behaviour like some people will override `method_missing` also with `respond_to_missing?` properly, that would be mostly alright, but issues might still happen especially for mocks implementation which often talks about proxies), and that's why we have `__send__` we can use, sadly not for `method` and it may be confusing that `__method__` is actually a completely different method! Not like `__send__` is an alias of `send`.

Sorry that this is probably falling down to a rant. To be more on the topic, since the test is already there, I don't think a regression is acceptable. However if we never added that test, I would think it's fine to just say "don't override `method`". Maybe RSpec 5, or maybe it doesn't matter after all. It's just one method right now. We don't have an army of `Object.instance_method(method_name).bind(object)` yet. Or do we? Maybe it's inside one of the support libraries already? 😅 To be honest I won't be surprised!

pirj: We defend against a number of quite odd things:
```
# This class protects us against Mutex.new stubbed out within tests
```
https://github.com/rspec/rspec-core/pull/2886, https://github.com/rspec/rspec-support/pull/431 and in a few other places.

> Maybe it's inside one of the support libraries already? 😅

Good point, it is! 😄 
https://github.com/rspec/rspec-support/blob/d63133f478408c1d965e673b96ad10ef5a5d183f/lib/rspec/support.rb#L53

godfat: > We defend against a number of quite odd things:

Oh my, I was guilty here. I am sure I was once lazy and stubbed `File.read` without explicit about the arguments, and it caused some issues. However being explicit about a specific argument did the trick for me.

I wonder if this kinds of things will need a more general approach. For example, will this work?

``` ruby
RSpec::Support::Source::File = ::File.dup
```

> Good point, it is! 😄

Ok, thank you for spotting that. I updated the code to use it instead.

pirj: > kinds of things

I can think of three cases we've touched:
1. Ruby core method stubbed (`File.read`, `File.expand_path`, `Mutex.new`)
2. Ruby core method overridden in a subclass (`Kernel.method`)
3. Attempt to define a method with memoized helpers that would break `Example` (`to_s`, `initialize`)

Not sure if there could be a common solution to all of this.

godfat: > Ruby core method stubbed (File.read, File.expand_path, Mutex.new)

I just tried this does work:

``` ruby
PreservedFile = File.dup

def File.read
  'break you'
end

PreservedFile.read('README.md')
```

We can have a list of preserved classes.

``` ruby
module RSpec
  module Preserved
    File = ::File.dup
    Mutex = ::Mutex.dup
  end
end
```

> Ruby core method overridden in a subclass (Kernel.method)

We can always do it via the same trick.

``` ruby
module RSpec
  module Preserved
    module_function
    def send_kernel(object, message, *args, **kargs)
      ::Kernel.instance_method(message).bind(object).call(*args, **kargs)
    end
  end
end

method_owner = RSpec::Preserved.send_kernel(object, :method).owner
```

> Attempt to define a method with memoized helpers that would break Example (to_s, initialize)

For `to_s` and `initialize` I think we should just raise an error like now, indeed. It would be too much work for those to be overridden. Even if we can stick with `__to_s__` or `__initialize__` and play around with `allocate`, it's easily broken with other integration.

> Not sure if there could be a common solution to all of this.

I didn't mean a common solution to all of above. I meant a common solution for the same issue. For example, we use the same way to resolve `File.read` or `Mutex.new` (a module method) got overridden. We use the same way to resolve a particular instance method got overridden.

JonRowe: Our policy is to do the method work around for cases as they occur

JonRowe:
> @JonRowe Should we merge to 4-0-dev, or wait until we release 4.0, and release this in 4.0.1?

@pirj I'm happy for you to make that decision, personally I'd lean towards merging into `4-0-dev` if you're happy with it.

pirj:
@godfat, thanks a lot for getting this to the finish line, and for an interesting conversation.

fledman:
@JonRowe @pirj 
is there somewhere to follow along how rspec 4 development is going?
(besides just scrolling through the lists of open PRs)

pirj:
@fledman There's this ticket https://github.com/rspec/rspec/issues/61.
In a nutshell, RSpec 4 is pretty much done. There are 2 open PRs that introduce deprecation warnings for the future RSpec 3.99, and one is WIP.
You can also look at `RSpec 4` milestones in repos (e.g. https://github.com/rspec/rspec-mocks/milestone/11).

This PR was not exactly planned for the release, but it turned out to be easier to target it for 4.0 rather than support all the older Ruby versions.

If you wish to give RSpec 4 a try, you can do it right away:
```ruby
# Gemfile
%w[rspec rspec-core rspec-expectations rspec-mocks rspec-support].each do |lib|
  gem "#{lib}", github: "rspec/#{lib}", branch: '4-0-dev'
end
```
Any pre-release bug report from a real-world project is highly appreciated.

fledman:
@pirj I got a relatively large codebase running successfully

the most troublesome adjustment was the change to `strict_predicate_matchers`

it was very unexpected that `to be_XXX` and `not_to be_XXX` are not inverses: both can fail
particularly when the underlying value is `nil` e.g. `not_to (be|have)_XXX` is effectively an `eql(false)` check

also ruby built-ins no longer being assertable this way e.g. `be_infinite` on positive and negative infinity

otherwise it was just getting the Gemfile to resolve all the conflicts due to it being a pre-release:
* I had to set `ENV['RSPEC_CI'] = '1'` and `ENV['RSPEC_VERSION'] = '4.0.0.pre'` in my Gemfile itself
* I had to pull `rspec-rails` directly from github so those vars would get picked up by its gemspec
* I had to fork a few other gems that had restrictive gemspec dependencies (`rspec-sidekiq` and `rspec_junit_formatter`)

fledman:
I suppose it would have been helpful if there had been deprecation warnings for the switch from`if:`/`unless:` to `skip:`

pirj:
@fledman Thanks a lot for putting your time into this!

We'll add deprecation warnings in 3.99, e.g. [for `:if`/:unless`](https://github.com/rspec/rspec-core/pull/2880/files#diff-796fd1d47b2b1c77641a959b3119518deea3fd49dc93513871be66c371d8f756R98).

I really hope strict predicate matchers are for the good.
> some Ruby built-ins no longer being assertable this way e.g. be_infinite on positive and negative infinity

```ruby
Float::INFINITY.infinite? # => 1
```
Oh, that is true. Well, truthy. There were a couple of similar ones, and it's mostly surprising that they don't return `true`/`false`.

fledman:
@pirj maybe you can have both strict (true/false) and loose (truthy/falsey) forms of the be/have matchers?

that way, regardless of which is the "default", it is easy to do inline adjustment

some syntax possibilities:
1. picking a symbol to append e.g. `be_infinite` vs `be_infinite?` or `have_foo` vs `have_foo!`
2. adding an option to the matcher e.g. `be_ready(strict: false)` 
3. coming up with another prefix (couldn't think of one)

alternatively there could be an explicitly truthy/falsey matcher
`expect(number).to reply(:infinite?)`
or
`expect(number).to be_truthy(:infinite?)`

PR #1221, opened 2018-05-11
Added `duck_type_including` ArgumentMatcher

xtrasimplicity:
# Background
If you are working with a hash, you can use:
```
allow(obj).to receive(:message).with(hash_including(name: 'Fred', last_name: 'Flintstone'))
```
If you are working with objects, you can use:
```
allow(obj).to receive(:message).with(duck_type(:name, :last_name))
```

But what if you want to combine these, and allow any object that **both** responds to `name` and `last_name` **and** returns `Fred` and `Flintstone`, respectively? That's where the `duck_type_including` argument matcher comes in. :)

# Implementation
```
allow(obj).to receive(:message).with(duck_type_including(name: 'Fred', last_name: 'Flintstone'))
```

### DIFF
--- a/lib/rspec/mocks/argument_matchers.rb
+++ b/lib/rspec/mocks/argument_matchers.rb
@@ -52,6 +52,15 @@ module RSpec
         DuckTypeMatcher.new(*args)
       end
 
+      # Matches if the actual argument responds to the specified messages, and the values match.
+      #
+      # @example
+      #   expect(object).to receive(:message).with(duck_type_including(name: 'Fred'))
+      #   expect(object).to receive(:message).with(duck_type_including(name: 'Fred', last_name: 'Flintstone'))
+      def duck_type_including(**args)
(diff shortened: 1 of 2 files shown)


JonRowe:
I'm pretty sure you can already do this with `allow(myObj).to receive(:message).with(have_attributes(name: 'Fred', last_name: 'Flintstone'))`

mikegee:
`having_attributes` is also aliased as `an_object_having_attributes`, which reads better in this case.

xtrasimplicity:
Thanks all! I swear I've tried those and failed, but I'll take another look when I get a chance.

xtrasimplicity:
These work perfectly. Thanks @JonRowe and @mikegee! I've submitted [a PR](https://github.com/rspec/rspec-mocks/pull/1222) to have the ArgumentMatcher documentation updated. :)

PR #1228, opened 2018-06-19
Support temporarily suppressing any instance expectations

### LINKED ISSUE #1225: without_partial_double_verification incorrectly fails allow_any_instance_of
We're upgrading gems, and came across this issue. We have some legacy code that unfortunately requires `allow_any_instance_of`. We have been using a backport of `without_partial_double_verification` which has been working fine. We've just upgraded to rspec 3.7 and now our legacy tests are failing because it looks like `without_partial_double_verification` doesn't work with `allow_any_instance_of`. Example spec code:

```ruby
    before do
      without_partial_double_verification do
        allow(Object).to receive(:success)
        allow_any_instance_of(LegacyModule).to receive(:success)
      end
    end
```

With the back port, `allow` and `allow_any_instance_of` both work. With RSpec 3.7, the `allow` works but `allow_any_instance_of` fails with `LegacyModule does not implement #success`.

Here's our current `without_partial_double_verification` that works as expected.

```ruby
def without_partial_double_verification(&block)
  original_state = RSpec::Mocks.configuration.verify_partial_doubles?
  RSpec::Mocks.configuration.verify_partial_doubles = false
  yield
ensure
  RSpec::Mocks.configuration.verify_partial_doubles = original_state
end
```

JonRowe:
Apparently this was missed previously.

Fixes #1225

### DIFF
--- a/lib/rspec/mocks/any_instance/recorder.rb
+++ b/lib/rspec/mocks/any_instance/recorder.rb
@@ -242,7 +242,7 @@ module RSpec
         def observe!(method_name)
           allow_no_prepended_module_definition_of(method_name)
 
-          if RSpec::Mocks.configuration.verify_partial_doubles?
+          if RSpec::Mocks.configuration.verify_partial_doubles? && !Mocks.configuration.temporarily_suppress_partial_double_verification
             unless public_protected_or_private_method_defined?(method_name)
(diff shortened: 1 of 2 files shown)


pacoguzman:
Hi all! any plan on releasing at least a version that includes this change?

JonRowe:
Yes absolutely, but no timeline at the moment

PR #1231, opened 2018-06-25
Improve error message for negative any_instance spy expectations

Darhazer:
Fixes https://github.com/rspec/rspec-expectations/issues/920

PR #1232, opened 2018-06-27
Follow up to #1231

JonRowe:
My apologies after merging #1231 (whilst writing the changelog!) I noticed the spec is actually wrong, the implementation is correct but the spec was calling `to` and not `to_not`. This fixes that.


JonRowe:
No apology needed @benoittgt I missed it too :p

PR #1233, opened 2018-06-27
Remove duplicated examples from specs

Darhazer:
* Removed duplicated specs
  - One(old syntax for any_instance) became obsolete since 411743175724d04885367f821e515824b747e781
  - The other (matches the same array,  specified without square brackets) is duplicated since it's very existance
* Updated couple of test descrirption to reflect the actual test

### DIFF
--- a/spec/rspec/mocks/array_including_matcher_spec.rb
+++ b/spec/rspec/mocks/array_including_matcher_spec.rb
@@ -14,6 +14,10 @@ module RSpec
 
         context "passing" do
           it "matches the same array" do
+            expect(array_including([1, 2, 3])).to be === [1, 2, 3]
+          end
+
+          it "matches the same array, specified without square brackets" do
             expect(array_including(1, 2, 3)).to be === [1, 2, 3]
           end
 
--- a/spec/rspec/mocks/null_object_double_spec.rb
+++ b/spec/rspec/mocks/null_object_double_spec.rb
@@ -76,7 +76,7 @@ module RSpec
         expect(val).to equal(@double)
       end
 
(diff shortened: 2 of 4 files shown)


spec/rspec/mocks/array_including_matcher_spec.rb
```diff
@@ -17,10 +17,6 @@ module ArgumentMatchers
             expect(array_including(1, 2, 3)).to be === [1, 2, 3]
           end
 
-          it "matches the same array,  specified without square brackets" do
```
JonRowe: So there is a mistake here, the spec above should be `array_including([1, 2, 3])` and this one left in, if you could change that, that would be grand! (Although you could remove the double space in the title :))

spec/rspec/mocks/null_object_double_spec.rb
```diff
@@ -76,7 +76,7 @@ module Mocks
         expect(val).to equal(@double)
       end
 
-      it 'allows unexpected message sends using `send`' do
+      it 'allows unexpected message sends using `__send__`' do
```
JonRowe: Can you leave this as it was? `__send__` is the same as `send` but we use the double underscore version "just to be sure" TM

Darhazer: I find it confusing having the same description, but :ok:, I'll revert

JonRowe: Oh sorry I take it back, please keep!

Darhazer:
Updated

JonRowe:
Thanks!

PR #1235, opened 2018-09-12
Any instance expectations with super/sub classes and nested blocks leak stubs.

JonRowe:
When using nested method calls and blocks, stubbing a super class method will leak into the child class for the duration of the block. First reported on rspec/rspec-core#2563 (with reproduction over at https://github.com/jcavalieri/rspec_issue).

Attached commit demonstrates the issue.


JonRowe:
@benoittgt This PR is just demonstrating the issue, it still needs fixing.

Issue #673, opened 2014-05-19
More thread safety fixes needed


myronmarston:
Ran across this gist that demonstrates some thread safety issues with 3.0.0.beta2 (and 2.14, of course):

https://gist.github.com/ordinaryzelig/f3a48b2456eac28bbbca

Thought it would be useful for when we dig more into this.

trejkaz:
> I think I'd be ok with making a call that rspec-mocks does not support parallel threaded execution.
> The only time you should be using it is in unit specs, which should be hella fast anyways, right?

What if you're testing a component which internally uses threads?

    context "when a callback is provided"
      it "calls the callback once per item"
        callback = double(:callback)
        expect(callback).to receive(:found).exactly(24).times
        component.search('query', callback)
      end
    end

With rspec-mocks 3.4.1, this fails for me claiming that the callback has not been called 24 times, even though I can see from the debugger that it is called exactly 24 times. My guess would be that the maps it's using to keep track of this are not synchronised. If you modify the code to print to the console for every time it's called, the issue mysteriously stops happening because console output forces a sync. Jamming a sleep in after the call also fixes the test. Both of these are of course cheating.

JonRowe:
Given what you've described, it's more likely the expectation is being verified before your method call is "complete".

trejkaz:
What actually happens is that the tasks which are spun off in the background all complete on their different threads before the method call returns, but the threads themselves are in a pool, so the threads themselves don't complete. But I think from the point of view of concurrency, these calls to the callback don't "happen-before" the check rspec does at the end of the test, because it isn't synchronising whatever structures it's storing this information in, so effectively it isn't being verified "after" the calls have been made.

I plan to try and make a standalone test for it. I'm just not sure yet whether such a test would reproduce the issue if written entirely in Ruby, or whether I have to call the callback from Java code on JRuby to get this behaviour.

JonRowe:
The moment `component.search('query', callback)` returns to RSpec, the expectation will be verified, the number of calls that RSpec has received by then will be used. You can fix this by adding a check to check for when your thread pool is drained.

trejkaz:
All the tasks submitted to the thread pool complete before the method call returns. This I already know, because the same unit test written in Java passes.

JonRowe:
That doesn't apply here, RSpec can't see beyond it's own thread AFAIK

Issue #380, opened 2013-07-25
thread safety


knewter:
Using class methods and ivars to track mocks isn't thread safe. Can generate a trivial test case that fails for this later. Opened now to begin discussion. See twitter thread here: https://twitter.com/myronmarston/status/360261658788114432

PR #1236, opened 2018-09-13
Concurrent execution vs mocks

yaauie:
On JRuby, we're seeing concurrency errors while validating message expectations when the code under test does work in threads.

There are two separate errors that show up, exemplified with a spec:

1. expected count doesn't always match the number of times the message was actually received
2. JRuby can throw a `ConcurrencyError` exception with message `Detected invalid array contents due to unsynchronized modifications with concurrent users` when an array is modified by one thread while it is being iterated over in another thread.

This Pull-Request is a proof-of-concept to show the issue and a potential solution.

It adds mutexes to synchronize _writes_ to `RSpec::Mocks::MessageExpectation::ImplementationDetail`'s `@actual_received_count` instance variable and to synchronize _all_ access to `RSpec::Mocks::Proxy`'s `@messages_received` array.

I am not particularly thrilled with the implementation, since it disperses the details of synchronisation across each of the classes and would be pretty easy to accidentally circumvent these mutexes.

### DIFF
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -357,6 +357,7 @@ module RSpec
       # some collaborators it delegates to for this stuff but for now this was
       # the simplest way to split the public from private stuff to make it
       # easier to publish the docs for the APIs we want published.
+      # rubocop:disable Metrics/ModuleLength
       module ImplementationDetails
         attr_accessor :error_generator, :implementation
         attr_reader :message
@@ -366,6 +367,11 @@ module RSpec
 
         # @private
         attr_reader :type
+
+        unless defined?(Mutex)
+          Support.require_rspec_support 'mutex'
+          Mutex = Support::Mutex
+        end
 
         # rubocop:disable Metrics/ParameterLists
         def initialize(error_generator, expectation_ordering, expected_from, method_double,
@@ -378,6 +384,7 @@ module RSpec
           @orig_object = @method_double.object
           @message = @method_double.method_name
           @actual_received_count = 0
+          @actual_received_count_write_mutex = Mutex.new
           @expected_received_count = type == :expectation ? 1 : :any
           @argument_list_matcher = ArgumentListMatcher::MATCH_ALL
           @order_group = expectation_ordering
@@ -536,7 +543,9 @@ module RSpec
         end
 
         def increase_actual_received_count!
-          @actual_received_count += 1
+          @actual_received_count_write_mutex.synchronize do
+            @actual_received_count += 1
+          end
         end
 
       private
@@ -567,7 +576,9 @@ module RSpec
             parent_stub.invoke(nil, *args, &block)
           end
         ensure
-          @actual_received_count += increment
+          @actual_received_count_write_mutex.synchronize do
+            @actual_received_count += increment
+          end
         end
 
         def has_been_invoked?
@@ -626,6 +637,7 @@ module RSpec
           nil
         end
       end
+      # rubocop:enable Metrics/ModuleLength
 
       include ImplementationDetails
     end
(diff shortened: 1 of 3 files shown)


spec/rspec/mocks/stubbed_message_expectations_spec.rb
```diff
@@ -23,6 +23,23 @@
     reset dbl
   end
 
+  it 'handles concurrent validation of expectations' do
+    dbl = double('double', :foo => true)
+    concurrency = 100
+    repetition = 100
```
JonRowe: Could we knock this down a touch, seems a tad overkill for our CI...

lib/rspec/mocks/proxy.rb
```diff
@@ -1,3 +1,5 @@
+RSpec::Support.require_rspec_support 'mutex'
```
JonRowe: So the point of this is to be loaded where the core `Mutex` is not available, I'd prefer we replace this with: 
```
unless defined?(Mutex)
  RSpec::Support.require_rspec_support 'mutex'
  Mutex = RSpec::Support::Mutex
end
```

JonRowe:
Hi, thanks for this, we're definitely not as thread safe as I'd like! I have a few suggestions above and this needs to pass our rubocop checks, (either refactor to remove some lines or mark as skipping the check)

lib/rspec/mocks/proxy.rb
yaauie: 👍 I'm glad to do this. The fall-through to stdlib makes sense to me.

All of the other invocations of `RSpec::Support#require_rspec_support` in this project follow the pattern I initially used, so it may be worth filing a ticket to clean them up too.

JonRowe: The `require_rspec_support` is just a convenience helper to load a file from our gem, its not always used to polyfill, mostly its actual code we own, it's just this one specific instance, which is only used to create our `ReEntrantMutex` as far as I'm aware.

xaviershay:
some potentially useful comments for background on overall thread-safety (or lack thereof) in  https://github.com/rspec/rspec-mocks/issues/673 and https://github.com/rspec/rspec-mocks/issues/380 if you hadn't seen them.

yaauie:
I've refactored to fall-through to stdlib's `Mutex` if it is present, and have disabled the rubocop `ModuleLength` for the offending module, since I couldn't find a simple way to refactor it to fit under the current complexity limitation.

yaauie:
Appveyor build is failing because the MRI-1.9.3 job can't seem to contact rubygems.org: https://ci.appveyor.com/project/rspec/rspec-mocks/build/803/job/rkny5fsmc3kvbw9l

JonRowe:
Great, if you could just bump down the iterations in the test a bit I think this is good to go

lib/rspec/mocks/message_expectation.rb
```diff
@@ -367,6 +368,11 @@ module ImplementationDetails
         # @private
         attr_reader :type
 
+        unless defined?(Mutex)
+          Support.require_rspec_support 'mutex'
+          Mutex = Support::Mutex
```
JonRowe: Can we refactor this slightly sorry, can we do this:
```Can we refactor this slightly sorry, can we do this:
```
if defined?(Mutex)
  Support::Mutex = Mutex
else
  Support.require_rspec_support 'mutex'
end
```

And change all the calls to `Support::Mutex`

The reasoning being this is a lib file and we don't want to poison the older Rubies. (It would make it seem like `Mutex` is available in code during tests, and then fail in reality).

JonRowe:
If you rebase this the appveyor build will pass too, if you don't have time to do this let me know and I'll attempt to tackle it this week.

JonRowe:
Merged in #1244, thank you very much for contributing this!

PR #1237, opened 2018-09-17
Add argument matcher for expecting a block.

JonRowe:
This is @kaiwren's `a_block` argument matcher with fleshed out sad path specs, and a refactoring of the match logic to allow mixing with other arguments. Review please @kaiwren @ioquatix @benoittgt.

### DIFF
--- a/features/setting_constraints/matching_arguments.feature
+++ b/features/setting_constraints/matching_arguments.feature
@@ -122,3 +122,28 @@ Feature: Matching arguments
       """
     When I run `rspec responding_differently_spec.rb`
     Then the examples should all pass
+
+  Scenario: Expecting a method invocatin with a block
+    Given a file named "expect_method_with_block.rb" with:
+    """ruby
+    RSpec.describe "Expecting a message with a block" do
+      let(:dbl) { double }
+      before { expect(dbl).to receive(:foo).with(a_block) }
+
+      it "passes when the method is invoked with a block" do
+        dbl.foo { |a| :block }
+      end
+
+      it "fails when the method is invoked without a block" do
+        dbl.foo
+      end
+    end
+    """
+  When I run `rspec expect_method_with_block.rb`
+  Then it should fail with the following output:
+    | 2 examples, 1 failure                                           |
+    |                                                                 |
+    | Failure/Error: dbl.foo                                          |
+    |   #<Double (anonymous)> received :foo with unexpected arguments |
+    |     expected: (a block)                                         |
+    |          got: (no args)                                         |
--- a/lib/rspec/mocks/argument_list_matcher.rb
+++ b/lib/rspec/mocks/argument_list_matcher.rb
@@ -54,8 +54,13 @@ module RSpec
       # position of the arguments passed to `new`.
       #
       # @see #initialize
-      def args_match?(*args)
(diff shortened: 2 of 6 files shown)


JonRowe:
Some further improvements could be made here, do we want to check for blocks in other places? e.g. is `expect(obj).to receive(:msg).with(a_block, a_block)` valid? And if we don't should we reject the method expectation out of hand?

benoittgt:
Maybe we can say this is good enough for a first step? Ready the tests I found it already a good improvement.

Good job

kaiwren:
A quick question - IMO ArgumentListMatcher and associated code could use a little refactoring to better express intent, and to include blocks in the args verification in a more consistent way (that's one of the reasons I stopped at a_block). I'm happy to attempt this if it makes sense to do so, and of course, it goes without saying that I'll keep the changes small, and will completely understand if they are rejected. Would that be useful?

JonRowe:
You are welcome to have a go at refactoring it if you think you can make it better, please make sure you understand the matcher protocol before changing method names, e.g. it's important and non optional that `===` is used as the match method, as it's part of our composability and is what lets argument matchers and other matchers be combined.

Thinking about this, theres now ambiguity as to what we mean when, @myronmarston I'm leaning towards adding `a_block` to match a block "anywhere", and then adding `.with_block` to mean the block argument in order to be ambiguous, meaning `with(...)`s arguments will remain "real" argument. Effectively `.with_block` would just check for the presence of `&block`, or could check for a specific block?

ioquatix:
Procs/Lambdas as an argument and blocks are different things. I suggest consider `a_block` and `a_proc`/`a_lambda`/`a_callable`?

kaiwren:
@JonRowe yep, well noted, I will keep that in mind.

Yep agree that bound and unbound blocks be treated differently. I was unaware of `and_yield` when I picked up this ticket, but +1 to @JonRowe's point IMO the contracts and use cases for `with(a_block)` `and_yield` and `with(a_block,  ..., a_block)` should be defined before proceeding further to avoid ambiguous apis.

JonRowe:
They're not different enough, and can be passed around... Consider:
```
2.5.1 :001 > def a(&block)
2.5.1 :002?>   puts block.inspect
2.5.1 :003?>   block
2.5.1 :004?> end
 => :a
2.5.1 :005 > a =  a { |a| }
#<Proc:0x00007fd290a01440@(irb):4>
 => nil
2.5.1 :006 > b = a(&proc { |b| })
#<Proc:0x00007fd2909e1c58@(irb):5>
 => nil
2.5.1 :007 > c =  a(&lambda { |c| })
#<Proc:0x00007fd2909d0548@(irb):6 (lambda)>
 => nil
2.5.1 :008 > Proc === a
 => true
2.5.1 :009 > Proc === b
 => true
2.5.1 :010 > Proc === c
 => true
```

kaiwren:
I've done some work to try to tease out `ReceivedMessage` as a concrete class from `Proxy`. I'm not too happy with it, but wanted to circle back to check on my direction with the team before moving further in this direction. Do let me know if this makes sense, or if I should move in a different direction. https://github.com/rspec/rspec-mocks/compare/master...kaiwren:refactor-extract-ReceivedMessage-from-Proxy

JonRowe:
I'm not sure why this was never merged, migrated to the monorepo as rspec/rspec#114

PR #1244, opened 2018-10-16
Concurrent execution vs mocks without falsely creating Mutex

JonRowe:
Finish up #1236 by handling the `Mutex` edge case.


benoittgt:
Thanks! 👍

PR #1275, opened 2019-05-03
Prevent error formatting from causing recursive errors

jamesdabbs-procore:
This comes from tracking down https://github.com/rspec/rspec-rails/issues/2049 wherein

```
it 'should show only 2 times' do
  expect(@student).to     receive(:email).once.and_call_original
  @student.email
  @student.email
end
```

overflows. There `Devise` has defined an `inspect` method which calls `email`, so when the second call to `email` errors, the [error generator's `inspect` call](https://github.com/rspec/rspec-mocks/blob/55f0b116dc8c15a7acef9e7619bb0a3925bd18fa/lib/rspec/mocks/error_generator.rb#L230) goes into a loop.

I've added a spec that I think describes the desired behavior, but which is currently failing against master.

The included `lib` change fixes the spec, but this is my first time digging into the RSpec codebase, and it's entirely possible there's a better place to make that change.

Please let me know if I've mis-understood the desired behavior, or if there's anything else you'd like to see changed. Thanks!

### DIFF
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -390,7 +390,7 @@ module RSpec
           @expectation_type = type
           @ordered = false
           @at_least = @at_most = @exactly = nil
-          @raised = false
+          @failed = false
 
           # Initialized to nil so that we don't allocate an array for every
           # mock or stub. See also comment in `and_yield`.
@@ -425,7 +425,7 @@ module RSpec
         end
 
         def invoke(parent_stub, *args, &block)
-          if @raised
+          if @failed
             invoke_incrementing_actual_calls_by(0, false, parent_stub, *args, &block)
           else
             invoke_incrementing_actual_calls_by(1, true, parent_stub, *args, &block)
@@ -561,7 +561,7 @@ module RSpec
           args.unshift(orig_object) if yield_receiver_to_implementation_block?
 
           if negative? || (allowed_to_fail && (@exactly || @at_most) && (@actual_received_count == @expected_received_count))
-            @raised = true
+            @failed = true
             # args are the args we actually received, @argument_list_matcher is the
             # list of args we were expecting
             @error_generator.raise_expectation_error(
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -390,6 +390,7 @@ module RSpec
           @expectation_type = type
           @ordered = false
           @at_least = @at_most = @exactly = nil
+          @raised = false
 
           # Initialized to nil so that we don't allocate an array for every
           # mock or stub. See also comment in `and_yield`.
@@ -424,7 +425,11 @@ module RSpec
         end
 
         def invoke(parent_stub, *args, &block)
-          invoke_incrementing_actual_calls_by(1, true, parent_stub, *args, &block)
+          if @raised
+            invoke_incrementing_actual_calls_by(0, false, parent_stub, *args, &block)
+          else
+            invoke_incrementing_actual_calls_by(1, true, parent_stub, *args, &block)
+          end
         end
 
         def invoke_without_incrementing_received_count(parent_stub, *args, &block)
@@ -556,6 +561,7 @@ module RSpec
           args.unshift(orig_object) if yield_receiver_to_implementation_block?
 
           if negative? || (allowed_to_fail && (@exactly || @at_most) && (@actual_received_count == @expected_received_count))
+            @raised = true
             # args are the args we actually received, @argument_list_matcher is the
             # list of args we were expecting
             @error_generator.raise_expectation_error(
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -390,7 +390,7 @@ module RSpec
           @expectation_type = type
           @ordered = false
           @at_least = @at_most = @exactly = nil
-          @failed = false
+          @invoking = false
 
           # Initialized to nil so that we don't allocate an array for every
           # mock or stub. See also comment in `and_yield`.
@@ -425,11 +425,15 @@ module RSpec
         end
 
         def invoke(parent_stub, *args, &block)
-          if @failed
-            invoke_incrementing_actual_calls_by(0, false, parent_stub, *args, &block)
+          if @invoking
+            safe_invoke_without_incrementing_received_count(parent_stub, *args, &block)
           else
             invoke_incrementing_actual_calls_by(1, true, parent_stub, *args, &block)
           end
+        end
+
+        def safe_invoke_without_incrementing_received_count(parent_stub, *args, &block)
+          invoke_incrementing_actual_calls_by(0, false, parent_stub, *args, &block)
         end
 
         def invoke_without_incrementing_received_count(parent_stub, *args, &block)
@@ -560,8 +564,9 @@ module RSpec
         def invoke_incrementing_actual_calls_by(increment, allowed_to_fail, parent_stub, *args, &block)
           args.unshift(orig_object) if yield_receiver_to_implementation_block?
 
+          @invoking = true
+
           if negative? || (allowed_to_fail && (@exactly || @at_most) && (@actual_received_count == @expected_received_count))
-            @failed = true
             # args are the args we actually received, @argument_list_matcher is the
             # list of args we were expecting
             @error_generator.raise_expectation_error(
@@ -580,6 +585,7 @@ module RSpec
             parent_stub.invoke(nil, *args, &block)
           end
         ensure
+          @invoking = false
           @actual_received_count_write_mutex.synchronize do
             @actual_received_count += increment
           end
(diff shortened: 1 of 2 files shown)


spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -152,6 +152,34 @@ def self.inspect
...
+          end
+        end
+
+        object = klass.new('foo')
+
+        expect(object).to receive(:name).once.and_return('bar')
+
+        object.name
+
+        expect do
```
JonRowe: Can you use `{`/`}` here in line with the specs around

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -152,6 +152,34 @@ def self.inspect
...
+          object.name
+        end.to raise_error(
+          RSpec::Mocks::MockExpectationError,
+          %|(<Inspector(bar)>).name(no args)\n    expected: 1 time with any arguments\n    received: 2 times|
+        )
+
+        expect do
+          verify object
+        end.to raise_error(
+          RSpec::Mocks::MockExpectationError,
```
JonRowe: Can you use the `fail_with` the shortcut here.

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -152,6 +152,34 @@ def self.inspect
...
+        klass = Struct.new(:name) do
+          def inspect
+            "<Inspector(#{name})>"
+          end
+        end
+
+        object = klass.new('foo')
+
+        expect(object).to receive(:name).once.and_return('bar')
+
```
JonRowe: Remove ws similar to spec above.

lib/rspec/mocks/message_expectation.rb
```diff
@@ -556,6 +561,7 @@ def invoke_incrementing_actual_calls_by(increment, allowed_to_fail, parent_stub,
           args.unshift(orig_object) if yield_receiver_to_implementation_block?
 
           if negative? || (allowed_to_fail && (@exactly || @at_most) && (@actual_received_count == @expected_received_count))
+            @raised = true
```
JonRowe: I think I prefer `@failed` for the name, as we've note yet raised, and there are circumstances where we don't until later.

lib/rspec/mocks/message_expectation.rb
```diff
@@ -424,7 +425,11 @@ def safe_invoke(parent_stub, *args, &block)
         end
 
         def invoke(parent_stub, *args, &block)
-          invoke_incrementing_actual_calls_by(1, true, parent_stub, *args, &block)
+          if @raised
+            invoke_incrementing_actual_calls_by(0, false, parent_stub, *args, &block)
```
JonRowe: You don't need to do this at all.

jamesdabbs-procore: I'm not entirely sure what you're saying here. Are you saying this could just be?:

```
def invoke(...)
  return if @failed 
  invoke_incrementing_actual_calls_by(1, true, ...)
end
```

If so, that breaks the new spec as written b/c `inspect` is expecting to receive the correct `name` post-failure. I also considered extracting an "invoke directly" from [around here](https://github.com/rspec/rspec-mocks/blob/55f0b116dc8c15a7acef9e7619bb0a3925bd18fa/lib/rspec/mocks/message_expectation.rb#L571) for clarity, or unsetting the method proxy once it failed, but those seemed like larger changes. If any of those approaches sound better, or if I'm misunderstanding the comment, please let me know.

spec/rspec/mocks/partial_double_spec.rb
jamesdabbs-procore: I _think_ I made the changes you're expecting here, but please let me know if not.

lib/rspec/mocks/message_expectation.rb
JonRowe: Well wouldn't `invoke_without_incrementing_received_count` be a better choice?

samphippen:
@jamesdabbs-procore would you be able to finish this one off and close it out?

jamesdabbs-procore:
Ah, sorry. Lost this thread. Definitely, although it might be sometime this weekend.

lib/rspec/mocks/message_expectation.rb
jamesdabbs-procore: So after digging a bit, I'm not entirely sure what's best here.

* `invoke_without_incrementing_received_count` is still [`allowed_to_fail` here](https://github.com/rspec/rspec-mocks/blob/bccc110309745a08c2c49fe44cbabd24c280489a/lib/rspec/mocks/message_expectation.rb#L560), so [this](https://github.com/rspec/rspec-mocks/blob/bccc110309745a08c2c49fe44cbabd24c280489a/spec/rspec/mocks/partial_double_spec.rb#L166) hits the same stack overflow
* `safe_invoke`, on the other hand, stops the error, but increments the count so [this](https://github.com/rspec/rspec-mocks/blob/bccc110309745a08c2c49fe44cbabd24c280489a/spec/rspec/mocks/partial_double_spec.rb#L174) error message ends up being `expected: 1 time with any arguments\n    received: 3 times with any arguments` instead, which is a little counter-intuitive for end users.

It seems like `invoke_incrementing_actual_calls_by(0, false, ...)` _is_ what needs to happen here, unless you're envisioning larger changes elsewhere.

pirj:
@jamesdabbs-procore ping!

jamesdabbs-procore:
Oh, sorry. I posted a threaded response earlier - https://github.com/rspec/rspec-mocks/pull/1275#discussion_r295124926 - and could use some feedback there.

lib/rspec/mocks/message_expectation.rb
JonRowe: Thanks for your patience here, I'd the code expressed what it does, if we have to change `invoke_without_incrementing_received_count` so we can send the false thats ok, we might just need two methods, personally I think the whole matcher implementation needs refactoring a bit breaking apart the methods so we don't have to pass these counts and a boolean flag...

lib/rspec/mocks/message_expectation.rb
```diff
@@ -424,7 +425,11 @@ def safe_invoke(parent_stub, *args, &block)
         end
 
         def invoke(parent_stub, *args, &block)
-          invoke_incrementing_actual_calls_by(1, true, parent_stub, *args, &block)
+          if @failed
```
JonRowe: I want to remove this and replace with `@invoking`

lib/rspec/mocks/message_expectation.rb
```diff
@@ -390,6 +390,7 @@ def initialize(error_generator, expectation_ordering, expected_from, method_doub
           @expectation_type = type
           @ordered = false
           @at_least = @at_most = @exactly = nil
+          @failed = false
```
JonRowe: I want to remove this and replace with `@invoking`

lib/rspec/mocks/message_expectation.rb
```diff
@@ -556,6 +561,7 @@ def invoke_incrementing_actual_calls_by(increment, allowed_to_fail, parent_stub,
           args.unshift(orig_object) if yield_receiver_to_implementation_block?
 
           if negative? || (allowed_to_fail && (@exactly || @at_most) && (@actual_received_count == @expected_received_count))
+            @failed = true
```
JonRowe: I want to remove this, and replace it with an `@invoking` check further down.

e.g
```
@invoking = true
if implementation.present?
  implementation.call(*args, &block)
elsif parent_stub
  parent_stub.invoke(nil, *args, &block)
end
@invoking = false
```

JonRowe: I think thats safer and I think theres a risk of false positives as is, as it won't record further invocations?

lib/rspec/mocks/message_expectation.rb
jamesdabbs-procore: Totally! Thanks for the feedback. I should have a little bit of downtime over the Thanksgiving weekend, so I'll take a look and aim to have something up next week.

jamesdabbs-procore:
@JonRowe - I think that last implements the changes you were requesting. I also don't love the proliferation of flags / method-name combinations. Are the existing `safe_invoke` and `invoke_without_incrementing_received_count` intended to be part of the publicly supported API here? (Is `invoke` itself?) If not, we could standardize on something like

```ruby
def invoke(parent_stub, *args, increment: 1, allow_failure: !@invoking, &block)
```

and update the existing callers.

What were you thinking in terms of refactoring the existing `invoke`?

benoittgt:
I find the current implementation pretty clean and readable.

I don't think we want this to be accessible to the public API. Do you see a reason to have it public? I am not against your proposal but it is little bit more complicated. :)

JonRowe:
Ok I'm happy to merge this once CI is green, if I feel that strongly about the method names I'll refactor it myself later 😂, can you give it a rebase?

JonRowe:
I'm not sure why this never got merged, I think it just needed a rebase and then got forgotten, sorry about that! I've migrated it to the monorepo rspec/rspec#116 and will aim to get it merged.

PR #1277, opened 2019-05-24
Do not change method visibility when stubbing private class methods.

kcboschert:
This fixes an issue where stubbing a private class method creates the stub as a public class method.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,10 @@ Enhancements:
   `expect(Class).to receive(:method).exactly(1).time`.
   (Pistos, Benoit Tigeot, #1271)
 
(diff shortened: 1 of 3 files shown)


kcboschert:
Suggestions for how to proceed with this build failure?

https://travis-ci.org/rspec/rspec-mocks/jobs/536863839#L5207
> Gem::RuntimeRequirementNotMetError: i18n requires Ruby version >= 2.3.0. The current ruby version is 2.2.0.

Seems unrelated to this change.

JonRowe:
Thanks, I'm not sure how this was missed previously, I did wonder if it was deliberate but I can't find evidence of this.

PR #1279, opened 2019-06-16
Update Aruba to version 0.14.10

mvz:
- Bump dependency version
- Fix deprecations
- Limit version of transitive dependency on contracts gem on old Rubies


JonRowe:
Thanks! 💚

PR #1297, opened 2019-11-02
Pin childprocess on older rubies

JonRowe:


JonRowe:
Build fixed itself, idk

PR #1309, opened 2019-12-30
#1308: Call when_declaring_verifying_double for any_instance partial …

### LINKED ISSUE #1308: Call when_declaring_verifying_double for any_instance partial mocks
### Using allow_ or expect_any_instance_of does not call the when_declaring_verifying_double callbacks

See https://github.com/rspec/rspec-rails/issues/1357.

The when_declaring_verifying_double callbacks only get called when instance_double and its ilk are called. In the case of any_instance, these callbacks are never called. The ramification of this is that rspec-rails will fail to dynamically define methods in this case and rspec-mocks will claim that the method doesn't exist.

### Your environment
* Ruby version: 2.6
* rspec-mocks version: 3.10.0-pre

### Steps to reproduce

Set verify_partial_mocks to true, then use `allow_any_instance_of` or `expect_any_instance_of` with a class that uses the callback to define dynamic methods.

```ruby
# frozen_string_literal: true

begin
  require "bundler/inline"
rescue LoadError => e
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rspec", "3.7.0" # Activate the gem and version you are reporting the issue against.
end

require 'rspec/autorun'

class MyClass
  def self.define_attribute_methods!
    unless method_defined?(:some_method_defined_dynamically)
      define_method(:some_method_defined_dynamically) { true }
    end
  end
end

RSpec.configuration.mock_with(:rspec) do |config|
  config.verify_partial_doubles = true
  config.when_declaring_verifying_double do |reference|
    reference.target.define_attribute_methods! if reference.target.is_a?(Class)
  end
end

RSpec.describe 'dynamic methods' do
  expect_any_instance_of(MyClass).to receive(:some_method_defined_dynamically)
end
```

### Expected behavior
The callback should be called and the method should exist.

### Actual behavior
The callback is not called and the method does not exist and raises an error.

dorner:
…mocks.

Fixes #1308.

### DIFF
--- a/lib/rspec/mocks/targets.rb
+++ b/lib/rspec/mocks/targets.rb
@@ -101,14 +101,6 @@ module RSpec
 
     # @private
     class AnyInstanceAllowanceTarget < TargetBase
-      def initialize(klass)
-        super
-        return unless RSpec::Mocks.configuration.verify_partial_doubles?
-        RSpec::Mocks.configuration.verifying_double_callbacks.each do |block|
-          block.call(ObjectReference.for(klass))
-        end
-      end
-
       def expression
         :allow_any_instance_of
       end
@@ -120,14 +112,6 @@ module RSpec
 
     # @private
     class AnyInstanceExpectationTarget < TargetBase
-      def initialize(klass)
-        super
-        return unless RSpec::Mocks.configuration.verify_partial_doubles?
-        RSpec::Mocks.configuration.verifying_double_callbacks.each do |block|
-          block.call(ObjectReference.for(klass))
-        end
-      end
-
       def expression
         :expect_any_instance_of
       end
--- a/lib/rspec/mocks/targets.rb
+++ b/lib/rspec/mocks/targets.rb
@@ -101,6 +101,14 @@ module RSpec
 
     # @private
     class AnyInstanceAllowanceTarget < TargetBase
+      def initialize(klass)
+        super
+        return unless RSpec::Mocks.configuration.verify_partial_doubles?
+        RSpec::Mocks.configuration.verifying_double_callbacks.each do |block|
+          block.call(ObjectReference.for(klass))
+        end
+      end
+
       def expression
         :allow_any_instance_of
       end
@@ -112,6 +120,14 @@ module RSpec
 
     # @private
     class AnyInstanceExpectationTarget < TargetBase
+      def initialize(klass)
+        super
+        return unless RSpec::Mocks.configuration.verify_partial_doubles?
+        RSpec::Mocks.configuration.verifying_double_callbacks.each do |block|
+          block.call(ObjectReference.for(klass))
+        end
+      end
+
       def expression
         :expect_any_instance_of
       end
--- a/spec/rspec/mocks/verifying_doubles/construction_spec.rb
+++ b/spec/rspec/mocks/verifying_doubles/construction_spec.rb
@@ -7,7 +7,9 @@ module RSpec
 
       class ClassThatDynamicallyDefinesMethods
         def self.define_attribute_methods!
-          define_method(:some_method_defined_dynamically) { true }
+          unless method_defined?(:some_method_defined_dynamically)
+            define_method(:some_method_defined_dynamically) { true }
+          end
         end
       end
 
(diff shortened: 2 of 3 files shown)


lib/rspec/mocks/targets.rb
```diff
@@ -101,6 +101,14 @@ def expression
 
     # @private
     class AnyInstanceAllowanceTarget < TargetBase
+      def initialize(klass)
+        super
+        return unless RSpec::Mocks.configuration.verify_partial_doubles?
+        RSpec::Mocks.configuration.verifying_double_callbacks.each do |block|
+          block.call(ObjectReference.for(klass))
+        end
```
JonRowe: This will cause the callback to be executed for every invocation of `allow_any_instance_of` which is different to their existing behaviour (their use is memoized to be once per class thanks to the space collection), I think tis should be place in `AnyInstance::Recorder#initialize` instead which is the same pattern as the normal proxies...

spec/rspec/mocks/verifying_doubles/construction_spec.rb
```diff
@@ -7,7 +7,9 @@ module Mocks
 
       class ClassThatDynamicallyDefinesMethods
         def self.define_attribute_methods!
-          define_method(:some_method_defined_dynamically) { true }
+          unless method_defined?(:some_method_defined_dynamically)
```
JonRowe: I think if you make the refactor suggested above you won't need this..

spec/rspec/mocks/verifying_doubles/construction_spec.rb
```diff
@@ -60,6 +62,43 @@ module CustomModule
...
           instance_double(test_class, :some_method_defined_dynamically => true)
         end
+
+        describe 'any_instance' do
+          let(:test_class) { Class.new(ClassThatDynamicallyDefinesMethods) }
+          before(:each) do
+            RSpec.configuration.mock_with(:rspec) do |config|
+              config.verify_partial_doubles = true
+              config.when_declaring_verifying_double do |reference|
+                reference.target.define_attribute_methods! if reference.target.is_a?(Class)
```
JonRowe: I mean... technically it should always be a Class, is it ever not?

spec/rspec/mocks/verifying_doubles/construction_spec.rb
```diff
@@ -60,6 +62,43 @@ module CustomModule
...
+              end
+            end
+          end
+
+          it 'calls the callback for expect_any_instance_of' do
+            expect_any_instance_of(test_class).to receive(:some_method_defined_dynamically)
+            expect { expect_any_instance_of(test_class).to receive(:some_invalid_method) }.
+              to raise_error(RSpec::Mocks::MockExpectationError,
+                             "#{test_class.to_s} does not implement #some_invalid_method")
```
JonRowe: `to_s` is implied

spec/rspec/mocks/verifying_doubles/construction_spec.rb
```diff
@@ -60,6 +62,43 @@ module CustomModule
...
+            expect(test_class.new.some_method_defined_dynamically).to eq(nil)
+          end
+
+          it 'calls the callback for allow_any_instance_of' do
+            allow_any_instance_of(test_class).to receive(:some_method_defined_dynamically)
+            expect { allow_any_instance_of(test_class).to receive(:some_invalid_method) }.
+              to raise_error(RSpec::Mocks::MockExpectationError,
+                             "#{test_class.to_s} does not implement #some_invalid_method")
```
JonRowe: `to_s` is implied

spec/rspec/mocks/verifying_doubles/construction_spec.rb
```diff
@@ -60,6 +62,43 @@ module CustomModule
...
+              config.verify_partial_doubles = true
+              config.when_declaring_verifying_double do |reference|
+                reference.target.define_attribute_methods! if reference.target.is_a?(Class)
+              end
+            end
+          end
+
+          it 'calls the callback for expect_any_instance_of' do
+            expect_any_instance_of(test_class).to receive(:some_method_defined_dynamically)
+            expect { expect_any_instance_of(test_class).to receive(:some_invalid_method) }.
```
JonRowe: Typically we format such invocations like this:
```
expect {
  expect_any_instance_of(test_class).to receive(:some_invalid_method)
}.to raise_error(RSpec::Mocks::MockExpectationError, ...)
```

I think I'd prefer that, and assign the message to a let given its the same...

spec/rspec/mocks/verifying_doubles/construction_spec.rb
```diff
@@ -60,6 +62,43 @@ module CustomModule
...
+
+          it 'should not call the callback if verify_partial_doubles is off' do
+            RSpec.configuration.mock_with(:rspec) do |config|
+              config.verify_partial_doubles = false
+            end
+
+            expect(test_class.method_defined?(:some_method_defined_dynamically)).to be_falsey
+          end
+        end
+
```
JonRowe: No need for a line here.

JonRowe:
This LGTM if it passes 🤞

dorner:
@JonRowe super quick response - thanks a ton for your helpful feedback!

dorner:
Thanks @JonRowe! What's the process for bumping up a release version with this? Do I need to snoop the commit history or is there a way to be notified when a version is released with this fix?

JonRowe:
Released @dorner

dorner:
Awesomesauce! 😁

pirj:
@dorner Perfect bug reports, research and execution. Bravo! 👍

dorner:
@pirj it helps that this project has such great guidelines and tooling set up. Made it super easy to contribute! ❤️

PR #1324, opened 2020-04-05
Handle keyword arguments separately for and_call_original in supported rubies

### LINKED ISSUE #1306: Keyword argument warning on Ruby 2.7 for #and_call_original
### Subject of the issue
`#and_call_original` with keyword arguments causes the following warning:

> ~/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/rspec-mocks-3.9.0/lib/rspec/mocks/message_expectation.rb:101: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

### Your environment
* Ruby version: 2.7.0
* rspec-expectations version: 3.9.0

### Steps to reproduce
```ruby
require 'rspec'

class Foo
  def initialize(bar: nil); end
end

RSpec.describe Foo do
  it 'works' do
    expect(Foo).to receive(:new).with(bar: :qux).and_call_original
    Foo.new(bar: :qux)
  end
end

RSpec::ExampleGroups::Foo.run
```

### Expected behavior
No warning is emitted.

### Actual behavior
A warning is emitted.

JonRowe:
Fixes #1306

### DIFF
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -98,7 +98,7 @@ module RSpec
       #   expect(counter.count).to eq(original_count + 1)
       def and_call_original
         wrap_original(__method__) do |original, *args, &block|
-          original.call(*args, &block)
+          __call_original(original, *args, &block)
         end
       end
 
@@ -353,6 +353,25 @@ module RSpec
         "#<#{self.class} #{error_generator.intro}.#{message}#{args_description}>"
       end
       alias inspect to_s
+
+      private
+
+      if RSpec::Support::RubyFeatures.kw_args_supported?
+        def __call_original(original, *args, &block)
+          if RSpec::Support::MethodSignature.new(original).has_kw_args_in?(args)
+            binding.eval(<<-CODE, __FILE__, __LINE__)
+            kwargs = args.pop
+            original.call(*args, **kwargs, &block)
+            CODE
+          else
+            original.call(*args, &block)
+          end
+        end
+      else
+        def __call_original(original, *args, &block)
+          original.call(*args, &block)
+        end
+      end
 
       # @private
       # Contains the parts of `MessageExpectation` that aren't part of
--- a/.rubocop_rspec_base.yml
+++ b/.rubocop_rspec_base.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2020-03-30T16:31:25+02:00 from the rspec-dev repo.
+# This file was generated on 2020-04-03T18:53:23+03:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 # This file contains defaults for RSpec projects. Individual projects
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2020-03-30T16:31:25+02:00 from the rspec-dev repo.
+# This file was generated on 2020-04-03T18:53:23+03:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 # In order to install old Rubies, we need to use old Ubuntu distibution.
@@ -22,10 +22,10 @@ rvm:
   - 2.1
   - 2.2.10
   - 2.3.8
-  - 2.4.9
-  - 2.5.7
-  - 2.6.5
-  - 2.7.0
+  - 2.4.10
+  - 2.5.8
+  - 2.6.6
+  - 2.7.1
   - ruby-head
   - ree
   - rbx-3
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,8 @@ matrix:
   include:
     - rvm: jruby-1.7
       env: JRUBY_OPTS='--dev --1.8'
+    - rvm: 2.7.1
+      env: DIFF_LCS_VERSION='~> 1.3.0'
   allow_failures:
     - rvm: jruby-head
     - rvm: ruby-head
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,8 @@ Enhancements:
 * Add the ability to set a custom error generator in `MessageExpectation`.
   This will allow rspec-expectations to inject a custom failure message.
   (Benoit Tigeot and Nicolas Zermati, #1312)
+* Return the result of the block passed to `RSpec::Mocks.with_temporary_scope`
+  when block run. (@expeehaa, #1329)
 
 ### 3.9.1 / 2019-12-31
 [Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.9.0...v3.9.1)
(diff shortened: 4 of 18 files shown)


pirj:
I'm still getting a warning running this:
```ruby
class Foo
  def initialize(bar: nil); end
end

RSpec.describe Foo do
  it 'works' do
    expect(Foo).to receive(:new).with(bar: :qux).and_call_original
    Foo.new(bar: :qux)
  end
end
```

JonRowe:
Ah thats a slightly difference scenario, I think we need more tests around this.

pirj:
That's from #1306 🤷‍♂

JonRowe:
Yeah I had written a similar spec assuming it was `and_call_original` to blame, but seems like we'll need more specs considering we have so little surrounding keyword argument usage in these places.

lib/rspec/mocks/message_expectation.rb
```diff
@@ -354,6 +354,25 @@ def to_s
       end
       alias inspect to_s
 
+      private
+
+      if RSpec::Support::RubyFeatures.kw_args_supported?
+        def __call_original(original, *args, &block)
+          if RSpec::Support::MethodSignature.new(original).has_kw_args_in?(args)
+            binding.eval(<<-CODE, __FILE__, __LINE__)
```
benoittgt, line 362: Still not super fan of using this. But we already had a discussion, and it seems the best solution. :)

benoittgt:
Does https://github.com/rspec/rspec-mocks/pull/1324#issuecomment-609468331 will be handle in this PR too?

JonRowe:
Yep I'll be continuing some work over in rspec support later, and then using that with more scenarios to fix this.

imanel:
Any update on this?

imanel:
@pirj https://github.com/rspec/rspec-support/pull/419 should solve the case you mentioned, although it's a bit hacky so I'm not sure if it will get accepted.

pirj:
Pushed a spec to cover the original request.
It passes nicely with https://github.com/rspec/rspec-support/pull/419.
Do you want me to move the conditional from `rspec-support` here?

pirj:
@JonRowe It seems we either have to merge https://github.com/rspec/rspec-support/pull/419, or to close it and pull the `method_or_class_initialize` to `rspec-mocks`.
What would you prefer?

benoittgt:
Quick win while waiting for the patch. Bump to Ruby 2.7.2 released this week. No more warnings.

pirj:
> Bump to Ruby 2.7.2 released this week. No more warnings.

That would be just perfect if we didn't have to do anything tricky with RSpec to be kwargs-compliant and no warnings were issued!

Wondering if this means there's a performance impact in Ruby 3.
Do you think it's possible to take one of our previous kwargs-related fixes and benchmark the difference?

I'm more inclined to let postpone those fixes if the performance gain is marginal rather than maintain this burden up to the point when we drop support for Ruby 2.x (for three more years to come or so).

JonRowe:
The warnings just become opt-in in 2.7.2, they don't change the fact we need to change these things before Ruby 3.0 is released.

benoittgt:
> The warnings just become opt-in in 2.7.2, they don't change the fact we need to change these things before Ruby 3.0 is released.

Yes of course. :)

pirj:
@benoittgt I'm available tomorrow in the afternoon. Lost track of what's going on, so you lead, me image

godfat:
I am providing two more failures for `and_wrap_original` and `block implementation`. This is just a demonstration so I omitted `binding.eval` and we might want to put those in the corresponding tests.

``` diff
diff --git a/spec/rspec/mocks/and_call_original_spec.rb b/spec/rspec/mocks/and_call_original_spec.rb
index 25c5a85c..4459142a 100644
--- a/spec/rspec/mocks/and_call_original_spec.rb
+++ b/spec/rspec/mocks/and_call_original_spec.rb
@@ -88,6 +88,18 @@ RSpec.describe "and_call_original" do
         expect(instance.initialize_arg).to eq :initialize_arg
         CODE
       end
+
+      it 'works with keyword arguments for and_wrap_original' do
+        expect(instance).to receive(:meth_3).and_wrap_original { |m, keyword_arg:| m.call(keyword_arg: keyword_arg.to_s) }
+
+        expect(instance.meth_3(keyword_arg: :original_arg)).to eq 'original_arg'
+      end
+
+      it 'works with keyword arguments for block implementation' do
+        expect(instance).to receive(:meth_4) { |keyword_arg:| keyword_arg }
+
+        expect(instance.meth_4(keyword_arg: :original_arg)).to eq :original_arg
+      end
     end
 
     it 'errors when you pass through the wrong number of args' do
```

Or in Ruby:

``` ruby
it 'works with keyword arguments for and_wrap_original' do
  expect(instance).to receive(:meth_3).and_wrap_original { |m, keyword_arg:| m.call(keyword_arg: keyword_arg.to_s) }

  expect(instance.meth_3(keyword_arg: :original_arg)).to eq 'original_arg'
end

it 'works with keyword arguments for block implementation' do
  expect(instance).to receive(:meth_4) { |keyword_arg:| keyword_arg }

  expect(instance.meth_4(keyword_arg: :original_arg)).to eq :original_arg
end
```

godfat:
This fixed `and_wrap_original`

``` diff
diff --git a/lib/rspec/mocks/message_expectation.rb b/lib/rspec/mocks/message_expectation.rb
index d999dc0f..643b7ca5 100644
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -115,7 +115,9 @@ module RSpec
       #     original_method.call(*args, &block).first(10)
       #   end
       def and_wrap_original(&block)
-        wrap_original(__method__, &block)
+        wrap_original(__method__) do |original, *args, &supplied_block|
+          __call_original(block, original, *args, &supplied_block)
+        end
       end
 
       # @overload and_raise
```

bryanp:
@JonRowe What's the status on getting this finished up and merged? `and_call_original` is currently broken on Ruby 3 when the original method accepts keyword arguments. Willing to help, but not sure what all is left here.

pirj:
@bryanp According to the plan, quite a lot of method have to be marked with `ruby2_keywords`.

I took a stab at a similar `rspec-core` issue today, but could barely handle half. There's an uncertainty if the whole chain should be marked as such or something is not right with the approach yet. On a synthetic example, `ruby2_keywords` do help, but marking just a few methods down the call chain doesn't.

You can certainly help. There is this https://github.com/rspec/rspec-mocks/issues/1306#issuecomment-752047914 that you can turn into a failing spec like this https://github.com/rspec/rspec-mocks/pull/1383/files#diff-ee4bec83cefa8d926951324b0059b58f31c016628981f29a0c391cd9b8a51315R15. After that, apply the approach described here https://github.com/rspec/rspec-expectations/issues/1221#issuecomment-716133588 and (in brief) here https://github.com/rspec/rspec-expectations/issues/1241#issuecomment-749694359.

If you decide to do so, please open a draft PR so we can avoid double work and intervene early in the process.

bryanp:
@pirj Here's an initial PR that resolves the issue I hit today: https://github.com/rspec/rspec-mocks/pull/1385.

Issue #1328, opened 2020-05-11
RSpec::Mocks.with_temporary_scope does not return the block result


expeehaa:
### Subject of the issue
When using a temporary scope I want to get a return value and work with it. However, the method `RSpec::Mocks.with_temporary_scope` does not return the result of the block given to it.

### Your environment
* Ruby version: ruby-2.5.3
* rspec-mocks version: 3.8.0

### Steps to reproduce
Run the following code:
```ruby
result = RSpec::Mocks.with_temporary_scope do
  5
end
```

### Expected behavior
`result` is `5`.

### Actual behavior
`result` is an empty hash `{}`.

JonRowe:
I think this is a reasonable request, but its not something thats expected to work (we don't document a return value), if you feel like working up a PR this is something that could be included in the future.

As a work around for now you can use Ruby scoping semantics to escape the closure:
```
result = nil
RSpec::Mocks.with_temporary_scope do
  result = 5
end
```

benoittgt:
Closed by https://github.com/rspec/rspec-mocks/pull/1329

PR #1329, opened 2020-05-11
Update RSpec::Mocks.with_temporary_scope to return the result of the passed block.

expeehaa:
This implements #1328.


JonRowe:
Thanks!

PR #1331, opened 2020-06-21
Updates from rspec-dev (2020-06-21)

JonRowe:

### DIFF
--- a/script/clone_all_rspec_repos
+++ b/script/clone_all_rspec_repos
@@ -1,5 +1,5 @@
 #!/bin/bash
-# This file was generated on 2019-12-26T17:20:33+00:00 from the rspec-dev repo.
+# This file was generated on 2020-06-21T22:22:12+01:00 from the rspec-dev repo.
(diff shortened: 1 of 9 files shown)


script/clone_all_rspec_repos
```diff
@@ -12,7 +12,11 @@ if is_mri; then
   clone_repo "rspec-core"
   clone_repo "rspec-expectations"
   clone_repo "rspec-mocks"
-  clone_repo "rspec-rails"
+  if requires_rspec_rails_maintenance_branch; then
+    clone_repo "rspec-rails" "4-0-maintenance"
+  else
+    clone_repo "rspec-rails"
+  fi
```
pirj, line 19: Maybe clone both and run against both?

JonRowe: The whole point of this is that the build fails against rspec-rails 3-9-maintenance due to the config predicate bug fix. 3.9.x is no longer maintained.

PR #1340, opened 2020-08-27
Improve (legacy) `any_instance` documentation example

saveriomiroddi:
Using the variable name `instance` keeps the general nature of the name, while making it more intuitive, due to proximity to the call `any_instance_of`.

See discussion (starting) at https://git.io/JUk98.


JonRowe (approved):

pirj:
Thank you! It looks much better.

PR #1347, opened 2020-10-06
Fix Ruby 2.7 warning for AndWrapOriginalImplementation#call

igor-drozdov:


lib/rspec/mocks/message_expectation.rb
```diff
@@ -736,8 +736,8 @@ def inner_action
         true
       end
 
-      def call(*args, &block)
-        @block.call(@method, *args, &block)
+      def call(*args, **kwargs, &block)
```
JonRowe, line 739: This is a parse error on some of our supported Rubies, so it can't be done like this. We have taken to conditionally defining such methods with use of string_eval where necessary, which is a shame but a necessary (pardon the pun) eval here.

It will also need tests demonstrating the use with keyword arguments.

JonRowe (changes requested):
Thanks for trying to tackle this, but it's not that simple.

PR #1349, opened 2020-11-15
Drop Ruby < 2.3 support

pirj:
Better viewed ignoring indentation changes https://github.com/rspec/rspec-mocks/pull/1349/files?diff=split&w=1

Sibling PRs:
 - https://github.com/rspec/rspec-support/pull/436
 - https://github.com/rspec/rspec-expectations/pull/1231
 - https://github.com/rspec/rspec-core/pull/2787

As per rspec/rspec#61 (comment)

 - [x] Is this needed in spec_helper.rb
  config.extend RSpec::Support::RubyFeatures
  config.include RSpec::Support::RubyFeatures
 - [x] RSpec::Support.require_rspec_support 'ruby_features'
 - [x] Usage of RubyFeatures
 - [x] clean predicate_functions
 - [-] check respond_to?
 - [x] check method_defined?
 - [x] check send (where we call private methods that later became public)
 - [x] check if we require_rspec_support ruby_features where RubyFeatures are not used
 - [x] check define_method where def would suffice
 - [x] check 1.8, 1.9, 2.0, 2.1, jruby, 9.0, 9.1, mri
 - [x] check eval
 - [x] clean up if RUBY_VERSION < "2.3" || backed_up_method_owner[method_name.to_sym] == self
 - [x] change old ruby induced rubocop settings, fix offences
 - [#] https://github.com/jruby/jruby/issues/1398 - only fixed in JRuby 9.2
 - [x] Can work without this check?  Gemfile|28| 4:if RUBY_VERSION <= '2.3.0' && !!(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/)
 - [x] skip "This intermittently fails on JRuby" if RUBY_PLATFORM == 'java'
 - [x] remove handle_restoration_failures?
 - [x] remove expect_output_warning_on_ruby_lt_2
 - [x] check "jruby"
 - [x] clean?  return Object unless Object.ancestors.last == ::BasicObject
 - [x] clean if defined?(BasicObject)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+### Development (unreleased)
+
+Breaking Changes:
+
+* Ruby < 2.3 is no longer supported. (Phil Pirozhkov, #1349)
+
 ### 3.10.0 / 2020-10-30
 [Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.9.1...v3.10.0)
 
--- a/Gemfile
+++ b/Gemfile
@@ -27,10 +27,6 @@ if RUBY_VERSION >= '2.4' && RUBY_ENGINE == 'ruby'
   gem 'rubocop', "~> 0.52.1"
 end
 
-if RUBY_VERSION <= '2.3.0' && !!(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/)
-  gem "childprocess", "< 1.0.0"
-end
-
 # Version 5.12 of minitest requires Ruby 2.4
 if RUBY_VERSION < '2.4.0'
   gem 'minitest', '< 5.12.0'
--- a/Gemfile
+++ b/Gemfile
@@ -12,19 +12,13 @@ branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
   end
 end
 
-if RUBY_VERSION < '1.9.3'
-  gem 'rake', '< 11.0.0' # rake 11 requires Ruby 1.9.3 or later
-elsif RUBY_VERSION < '2.0.0'
-  gem 'rake', '< 12.0.0' # rake 12 requires Ruby 2.0.0 or later
-else
-  gem 'rake', '> 12.3.2'
-end
-
 if ENV['DIFF_LCS_VERSION']
   gem 'diff-lcs', ENV['DIFF_LCS_VERSION']
 else
   gem 'diff-lcs', '~> 1.4', '>= 1.4.3'
 end
+
+gem 'ffi', '> 1.9.24' # prevent Github security vulnerability warning
 
 gem 'yard', '~> 0.9.24', :require => false
 
@@ -33,23 +27,8 @@ if RUBY_VERSION >= '2.4' && RUBY_ENGINE == 'ruby'
   gem 'rubocop', "~> 0.52.1"
 end
 
-# allow gems to be installed on older rubies and/or windows
-if RUBY_VERSION < '2.2.0' && !!(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/)
-  gem 'ffi', '< 1.10'
-elsif RUBY_VERSION < '1.9'
-  gem 'ffi', '< 1.9.19' # ffi dropped Ruby 1.8 support in 1.9.19
-elsif RUBY_VERSION < '2.0'
-  gem 'ffi', '< 1.11.0' # ffi dropped Ruby 1.9 support in 1.11.0
-else
-  gem 'ffi', '> 1.9.24' # prevent Github security vulnerability warning
-end
-
 if RUBY_VERSION <= '2.3.0' && !!(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/)
   gem "childprocess", "< 1.0.0"
-end
-
-if RUBY_VERSION < '1.9.2'
-  gem 'contracts', '~> 0.15.0' # is a dependency of aruba
 end
 
 # Version 5.12 of minitest requires Ruby 2.4
@@ -65,21 +44,6 @@ end
 
 gem 'simplecov', '~> 0.8'
 
-if RUBY_VERSION < '2.0.0' || RUBY_ENGINE == 'java'
-  gem 'json', '< 2.0.0' # this is a dependency of simplecov
-else
-  gem 'json', '> 2.3.0'
-end
-
-platforms :jruby do
-  if RUBY_VERSION < '1.9.0'
-    # Pin jruby-openssl on older J Ruby
-    gem "jruby-openssl", "< 0.10.0"
-    # Pin child-process on older J Ruby
-    gem "childprocess", "< 1.0.0"
-  else
-    gem "jruby-openssl"
-  end
-end
+gem "jruby-openssl", platforms: [:jruby]
 
 eval File.read('Gemfile-custom') if File.exist?('Gemfile-custom')
--- a/Gemfile
+++ b/Gemfile
@@ -46,4 +46,4 @@ gem 'simplecov', '~> 0.8'
 
 gem "jruby-openssl", platforms: [:jruby]
 
-eval File.read('Gemfile-custom') if File.exist?('Gemfile-custom')
+eval_gemfile 'Gemfile-custom' if File.exist?('Gemfile-custom')
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ else
   gem 'diff-lcs', '~> 1.4', '>= 1.4.3'
 end
 
-gem 'ffi', '> 1.9.24' # prevent Github security vulnerability warning
+gem 'ffi', '~> 1.12.0'
 
 gem 'yard', '~> 0.9.24', :require => false
 
--- a/Gemfile-custom.sample
+++ b/Gemfile-custom.sample
@@ -3,17 +3,10 @@ group :development do
   gem 'relish', '~> 0.6.0'
   gem 'guard-rspec', '~> 1.2.1'
   gem 'growl', '1.0.3'
-  gem 'spork', '0.9.0'
 
   platform :mri do
     gem 'rb-fsevent', '~> 0.9.0'
     gem 'ruby-prof', '~> 0.10.0'
-
-    case RUBY_VERSION
-    when /^1.8/
-      gem 'ruby-debug'
-    when /^1.9/
-      gem 'debugger'
-    end
+    gem 'pry-byebug'
   end
 end
--- a/lib/rspec/mocks.rb
+++ b/lib/rspec/mocks.rb
@@ -1,7 +1,6 @@
 require 'rspec/support'
 RSpec::Support.require_rspec_support 'caller_filter'
 RSpec::Support.require_rspec_support 'warnings'
-RSpec::Support.require_rspec_support 'ruby_features'
 
 RSpec::Support.define_optimized_require_for_rspec(:mocks) { |f| require_relative f }
 
--- a/lib/rspec/mocks/any_instance/recorder.rb
+++ b/lib/rspec/mocks/any_instance/recorder.rb
@@ -257,6 +257,7 @@ module RSpec
           @observed_methods << method_name
           backup_method!(method_name)
           recorder = self
+          # In Ruby 2.4 and earlier, `define_method` is private
           @klass.__send__(:define_method, method_name) do |*args, &blk|
             recorder.playback!(self, method_name)
             __send__(method_name, *args, &blk)
--- a/lib/rspec/mocks/any_instance/recorder.rb
+++ b/lib/rspec/mocks/any_instance/recorder.rb
@@ -275,18 +275,12 @@ module RSpec
           end
         end
 
-        if Support::RubyFeatures.module_prepends_supported?
-          def allow_no_prepended_module_definition_of(method_name)
-            prepended_modules = RSpec::Mocks::Proxy.prepended_modules_of(@klass)
-            problem_mod = prepended_modules.find { |mod| mod.method_defined?(method_name) }
-            return unless problem_mod
+        def allow_no_prepended_module_definition_of(method_name)
+          prepended_modules = RSpec::Mocks::Proxy.prepended_modules_of(@klass)
+          problem_mod = prepended_modules.find { |mod| mod.method_defined?(method_name) }
+          return unless problem_mod
 
-            AnyInstance.error_generator.raise_not_supported_with_prepend_error(method_name, problem_mod)
-          end
-        else
-          def allow_no_prepended_module_definition_of(_method_name)
-            # nothing to do; prepends aren't supported on this version of ruby
-          end
+          AnyInstance.error_generator.raise_not_supported_with_prepend_error(method_name, problem_mod)
         end
       end
     end
--- a/lib/rspec/mocks/any_instance/recorder.rb
+++ b/lib/rspec/mocks/any_instance/recorder.rb
@@ -266,6 +266,7 @@ module RSpec
         def mark_invoked!(method_name)
           backup_method!(method_name)
           recorder = self
+          # In Ruby 2.4 and earlier, `define_method` is private
           @klass.__send__(:define_method, method_name) do |*_args, &_blk|
             invoked_instance = recorder.instance_that_received(method_name)
             inspect = "#<#{self.class}:#{object_id} #{instance_variables.map { |name| "#{name}=#{instance_variable_get name}" }.join(', ')}>"
--- a/lib/rspec/mocks/any_instance/recorder.rb
+++ b/lib/rspec/mocks/any_instance/recorder.rb
@@ -200,24 +200,17 @@ module RSpec
             remove_method method_name
 
             # A @klass can have methods implemented (see Method#owner) in @klass
-            # or inherited from a superclass. In ruby 2.2 and earlier, we can copy
-            # a method regardless of the 'owner' and restore it to @klass after
-            # because a call to 'super' from @klass's copied method would end up
-            # calling the original class's superclass's method.
+            # or inherited from a superclass.
             #
-            # With the commit below, available starting in 2.3.0, ruby changed
-            # this behavior and a call to 'super' from the method copied to @klass
+            # A call to 'super' from the method copied to @klass
             # will call @klass's superclass method, which is the original
             # implementer of this method!  This leads to very strange errors
             # if @klass's copied method calls 'super', since it would end up
             # calling itself, the original method implemented in @klass's
             # superclass.
             #
-            # For ruby 2.3 and above, we need to only restore methods that
-            # @klass originally owned.
-            #
-            # https://github.com/ruby/ruby/commit/c8854d2ca4be9ee6946e6d17b0e17d9ef130ee81
-            if RUBY_VERSION < "2.3" || backed_up_method_owner[method_name.to_sym] == self
+            # We need to only restore methods that @klass originally owned.
+            if backed_up_method_owner[method_name.to_sym] == self
               alias_method method_name, alias_method_name
             end
             remove_method alias_method_name
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -1,4 +1,5 @@
 RSpec::Support.require_rspec_support "object_formatter"
+RSpec::Support.require_rspec_support 'ruby_features'
 
 module RSpec
   module Mocks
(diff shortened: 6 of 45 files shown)


lib/rspec/mocks/method_double.rb
```diff
@@ -205,6 +206,8 @@ def raise_method_not_stubbed_error
         RSpec::Mocks.error_generator.raise_method_not_stubbed_error(method_name)
       end
 
+      private
+
       # In Ruby 2.0.0 and above prepend will alter the method lookup chain.
```
benoittgt, line 211: Maybe the description should be changed?

lib/rspec/mocks/syntax.rb
```diff
@@ -179,20 +179,6 @@ def self.expect_enabled?(syntax_host=::RSpec::Mocks::ExampleMethods)
...
-        if defined?(JRUBY_VERSION) && JRUBY_VERSION == '1.7.4' && RUBY_VERSION.to_f > 1.8
-          return ::BasicObject
-        end
-
-        # On 1.8.7, Object.ancestors.last == Kernel but
-        # things blow up if we include `RSpec::Mocks::Methods`
-        # into Kernel...not sure why.
-        return Object unless defined?(::BasicObject)
-
         # MacRuby has BasicObject but it's not the root class.
```
benoittgt, line 182: Maybe this could be removed too?

script/functions.sh
```diff
@@ -12,10 +12,8 @@ SPECS_HAVE_RUN_FILE=specs.out
 MAINTENANCE_BRANCH=`cat maintenance-branch`
 
 # Don't allow rubygems to pollute what's loaded. Also, things boot faster
-# without the extra load time of rubygems. Only works on MRI Ruby 1.9+
-if is_mri_192_plus; then
-  export RUBYOPT="--disable=gem"
-fi
+# without the extra load time of rubygems.
+export RUBYOPT="--disable=gem"
```
benoittgt, line 16: Should we remove it too? Because I read "Only works on MRI Ruby 1.9+"? 🤔

benoittgt: Also those change should come from rspec-dev no?

benoittgt (commented):
Ouhaoo. So happy with those changes. I have few questions.

lib/rspec/mocks/syntax.rb
pirj: Sure!
I'm still actively working on this. Let me quickly update the description with my current TODO.

script/functions.sh
pirj: 1.9+ means 2.3-2.7 and above, so let's keep it for now. I know it went problematic for `rspec-rails`, but probably will still work for other repos.

On a funny note `rspec-core` had this in `spec/spec_helper.rb:
```ruby
require 'rubygems' if RUBY_VERSION.to_f < 1.9
```

pirj: >  Also those change should come from rspec-dev no?

Yes, absolutely. But I'm experimenting here of what can be stripped off, so we can find what needs to remain because it's still used in some of the repos and what can be removed in `rspec-dev`. I'll extract this and will send a follow-up PR to `rspec-dev`and to other repos against `4-0-dev`.

Gemfile-custom.sample
```diff
@@ -3,17 +3,11 @@ group :development do
   gem 'relish', '~> 0.6.0'
   gem 'guard-rspec', '~> 1.2.1'
   gem 'growl', '1.0.3'
-  gem 'spork', '0.9.0'
+  gem 'spring'
```
JonRowe, line 6: This is a sample, don't encourage spring 😂

lib/rspec/mocks/any_instance/recorder.rb
```diff
@@ -149,8 +149,8 @@ def stop_observing!(method_name)
           restore_method!(method_name)
           @observed_methods.delete(method_name)
           super_class_observers_for(method_name).each do |ancestor|
-            ::RSpec::Mocks.space.
-              any_instance_recorder_for(ancestor).stop_observing!(method_name)
+            ::RSpec::Mocks.space
+              .any_instance_recorder_for(ancestor).stop_observing!(method_name)
```
JonRowe, line 153: We don't need to make these changes, just keep the rubocop rule as it was...

lib/rspec/mocks/proxy.rb
```diff
@@ -101,7 +96,7 @@ def replay_received_message_on(expectation, &block)
             next unless expectation.matches?(actual_method_name, *args)
 
             expectation.safe_invoke(nil)
-            block.call(*args, &received_block) if block
+            block&.call(*args, &received_block)
```
JonRowe, line 99: I think it was clearer previously, just because something is now available doesn't mean we have to use it.

.travis.yml
```diff
@@ -15,33 +15,24 @@ before_install:
...
   - 2.5.8
   - 2.6.6
-  - 2.7.1
+  - 2.7.2
   - ruby-3.0.0-preview1
   - ruby-head
   - ree
   - rbx-3
-  - jruby-9.1.7.0 # pin JRuby to this until travis/rvm can install later versions
+  - jruby-9.2
```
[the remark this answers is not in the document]
pirj: Not sure yet. I was too aggressive with some changes even 2.3/2.4 started failing. I intend to make them green and check JRuby afterwards.

rspec-mocks.gemspec
```diff
@@ -45,7 +45,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency "diff-lcs", ">= 1.2.0", "< 2.0"
 
-  s.add_development_dependency 'rake',     '> 10.0.0'
+  s.add_development_dependency 'rake',     '> 12.3.2'
```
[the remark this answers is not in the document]
pirj: [`> 12.3.2` comes from our current `Gemfile`](https://github.com/rspec/rspec-mocks/blob/7fc7bf823646c87c6fb342ee1b8e03d7fb29da3b/Gemfile#L20):
```ruby
  gem 'rake', '> 12.3.2'
```
We can update it to the latest separately.

Gemfile-custom.sample
pirj: This for sure needs to be further cleanup 😄

maintenance-branch
```diff
@@ -1 +1 @@
-4-0-dev
\ No newline at end of file
+drop-old-rubies
```
pirj, line 1: ⚠️ This change is temporary and needs to be removed before merging, or soon after ⚠️

pirj:
```
============= Starting rspec-support specs ===============
Running specs for rspec-support
~/work/rspec-mocks/rspec-support ~/work/rspec-mocks/rspec-mocks
cat: .travis.yml: No such file or directory
```

Gemfile
```diff
@@ -12,46 +12,21 @@ branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
...
-  gem 'rake', '> 12.3.2'
-end
-
 if ENV['DIFF_LCS_VERSION']
   gem 'diff-lcs', ENV['DIFF_LCS_VERSION']
 else
   gem 'diff-lcs', '~> 1.4', '>= 1.4.3'
 end
 
+gem 'ffi', '> 1.9.24' # prevent Github security vulnerability warning
```
JonRowe, line 21: Does this not need to be `1.12` due to our lack of rubygems

Changelog.md
```diff
@@ -1,3 +1,7 @@
+# Development (unreleased)
+
+* Drop support for Ruby < 2.3. (Phil Pirozhkov, #1349)
```
JonRowe, line 3:
```suggestion
## Breaking Changes

* Ruby < 2.3 is no longer supported. (Phil Pirozhkov, #1349)
```

lib/rspec/mocks.rb
```diff
@@ -1,7 +1,6 @@
 require 'rspec/support'
 RSpec::Support.require_rspec_support 'caller_filter'
 RSpec::Support.require_rspec_support 'warnings'
-RSpec::Support.require_rspec_support 'ruby_features'
```
JonRowe, line 4: Honestly I would have left this here.

lib/rspec/mocks/error_generator.rb
```diff
@@ -1,4 +1,5 @@
 RSpec::Support.require_rspec_support "object_formatter"
+RSpec::Support.require_rspec_support 'ruby_features'
```
JonRowe, line 2: See other comment, now if we need this again elsewhere we'll have to remember to require it / move it back.

lib/rspec/mocks/method_double.rb
```diff
@@ -44,6 +43,8 @@ def visibility
 
       # @private
       def object_singleton_class
+        # We can't use @object.singleton_class because this method
+        # creates a new singleton class if the object doesn't have one.
```
JonRowe, line 47: Doesn't this apply to the previous swaps?

maintenance-branch
JonRowe: Why is this required?

spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
```diff
@@ -165,7 +165,7 @@ class << self
 
         context "on a class with a private `new`" do
           it 'uses the method signature from `#initialize` for arg verification' do
-            if RSpec::Support::Ruby.jruby? && RSpec::Support::Ruby.jruby_version < '9.2.1.0'
+            if RSpec::Support::Ruby.jruby? && JRUBY_VERSION < '9.2.1.0'
```
JonRowe, line 168: Revert

spec/spec_helper.rb
```diff
@@ -1,8 +1,7 @@
 require 'rspec/support/spec'
-require 'rspec/support/ruby_features'
 
 RSpec::Support::Spec.setup_simplecov do
-  minimum_coverage 93
+  minimum_coverage 90
```
JonRowe, line 4: :(

JonRowe (changes requested):
Great work, much fewer quibbles :joy:

Gemfile
pirj: Probably. I tried to keep the changes scoped to the removal of support of Ruby < 2.3
It's the way [it is for Ruby 2.3+ in 4-0-dev](https://github.com/rspec/rspec-mocks/blame/4-0-dev/Gemfile#L44).
I'll be happy to fix it here, and the reason why I didn't is because I find it problematic to foresee what you'd like to keep untouched and what to change.

I don't mind this change, and I clearly understand the problem with ffi 1.13.1 and long-unreleased 1.13.2.

lib/rspec/mocks.rb
pirj: Couldn't quickly find the issue, but I recall a weird case when `rspec-mocks` were required partially.
As an example, [this](https://github.com/rspec/rspec-mocks/blob/461d7f6f7f688f69154e410dcd9c7690120e7dbf/lib/rspec/mocks/argument_list_matcher.rb#L4):
```ruby
require 'rspec/mocks/argument_matchers'
```
Where [`lib/rspec/mocks/argument_list_matcher.rb `](https://github.com/rspec/rspec-mocks/blob/fa45affa48d2dadcd67d15644d403dedd601dfd5/lib/rspec/mocks/argument_matchers.rb#L4) don't require `rspec/mocks`.

I've moved `RSpec::Support.require_rspec_support 'ruby_features'` where `RubyFeatures` are used.

lib/rspec/mocks/error_generator.rb
pirj: That's what I did, went through all our files and added/removed it to only be required where used.

lib/rspec/mocks/method_double.rb
pirj: This statement still applies to other cases where I've replaced `(class << ... ; self; end)` but leaves specs green.
Do you think it makes sense to use our implementation of `singleton_method` that doesn't accidentally create a singleton class if it doesn't exist?
Would you consider adding a helper to `rspec-support`? AFAIR, we use this quite extensively here and in `rspec-expectations`.

maintenance-branch
pirj: I want to make sure that all those changes to our four repos work together.
4-0-dev branches have almost no differences (apart from Gemfile and CI files) from 3.x.
Because of
```
# pin to major/minor ignoring patch
```
I assume that compatibility of 4.0 with 4.0.x of our other gems is more important than with 3.x (that 4-0-dev didn't diverge much from at this moment).

It would be a nice bonus if `rspec-mocks` worked with `rspec-support` 3.9, and it probably does, but I was not too concerned about this compatibility.

spec/spec_helper.rb
pirj: Pretty significant drop, I agree. And it's just because of one file:
image

This might be coincidental. [`simplecov` 0.20.0 has been released a couple of weeks ago](https://rubygems.org/gems/simplecov/versions/0.20.0). It might have improved its detection, and our coverage % dropped.

There's a ticket somewhere Myron filed to increase the coverage, surely worth addressing.

spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
pirj: I see that a naïve string comparison is inferior to our `rspec-support`'s implementation of `jruby_version`, e.g.:
```ruby
'3.2' < '3.10' # => false
'2.3.rc1' < '2.3' # => false
```

However, here, in `rspec-mocks` we exclusively compare JRuby version in specs.
And in both cases, we compare with `'9.2.1.0'`.
I could not figure out a version number that RVM would install (and it only provides stable versions) that would be incorrectly compared. Can you?

I understand your concern, however, and I will revert it in `rspec-support`
On the one hand, it would be safe for future changes to keep the correct version comparison.
On the other hand, I remember you mentioned that we're aiming to drop specific Ruby implementation support, and "support Rubies that are MRI-compatible".

Just for the reference, this is a comprehensive list of our JRuby version comparisons. We only compare to "safe to compare" `9.2.0.0` and `9.2.1.0`.
```
$ rg JRUBY_VERSION
spec/integration/suite_hooks_errors_spec.rb
11:  if RSpec::Support::Ruby.jruby? && JRUBY_VERSION > '9.2.0.0'
99:      if RSpec::Support::Ruby.jruby? && JRUBY_VERSION > '9.2.0.0'

spec/rspec/core/formatters/exception_presenter_spec.rb
509:            if RSpec::Support::Ruby.jruby? && JRUBY_VERSION < '9.2.0.0'
526:            if RSpec::Support::Ruby.jruby? && JRUBY_VERSION < '9.2.0.0'

spec/rspec/support/ruby_features_spec.rb
125:          if Ruby.jruby? && JRUBY_VERSION < '9.2.1.0'

lib/rspec/support/ruby_features.rb
68:      if Ruby.rbx? || (Ruby.jruby? && JRUBY_VERSION < '9.2.1.0')

spec/rspec/expectations/failure_aggregator_spec.rb
70:        pending "This is broken on 9.2.x.x" unless JRUBY_VERSION < '9.2.0.0'

lib/rspec/expectations/failure_aggregator.rb
55:      if RSpec::Support::Ruby.jruby? && JRUBY_VERSION < '9.2.0.0'

spec/rspec/mocks/partial_double_spec.rb
596:            if RSpec::Support::Ruby.jruby? && JRUBY_VERSION < '9.2.1.0'

spec/rspec/mocks/verifying_doubles/class_double_with_class_loaded_spec.rb
168:            if RSpec::Support::Ruby.jruby? && JRUBY_VERSION < '9.2.1.0'
```

Please do not get me wrong, I do not resist bringing it back, but I'd like to know why exactly we're keeping it.

spec/rspec/mocks/verifying_doubles/naming_spec.rb
```diff
@@ -45,37 +45,37 @@ def error_message_for(_)
 
       describe "instance_double" do
         it_behaves_like "a named verifying double", "InstanceDouble" do
-          alias :create_double :instance_double
+          alias_method :create_double, :instance_double
```
pirj, line 48: This is something RuboCop insisted on. And there were JRuby failures that made me think might be related do this.
I don't mind reverting this and shutting up the cop.

pirj:
> Great work, much fewer quibbles 😂

After this comment, I've started experiencing an immense cognitive resistance at looking at your comments in other repos 😆

maintenance-branch
JonRowe: Removing old ruby support should be entirely compatible with 3.9, we're not adding anything so working with the `main` branch is confidence boosting and is a benefit.

pirj: It was 4-0-dev actually in the past. Reverted.
Hoping for a green build.

Gemfile
JonRowe: I'd make changes like this only to versions known to work..

pirj: I haven't made any change here 😆 

image

JonRowe: `main` is locked to `1.12`

spec/rspec/mocks/verifying_doubles/naming_spec.rb
JonRowe: I do not mind this change

JonRowe:
> After this comment, I've started experiencing an immense cognitive resistance at looking at your comments in other repos 😆

Sorry! Large PRs get large feedbacks :(

pirj:
Updates:
 - reverted `jruby_version` changes
 - pinned `ffi` to `~> 1.12.0`
 - accepted Changelog "breaking changes" section proposed change

Changelog.md
```diff
@@ -1,3 +1,9 @@
+# Development (unreleased)
+
+## Breaking Changes
```
JonRowe, line 3:
```suggestion
### Development (unreleased)

Breaking Changes:
```

lib/rspec/mocks/method_double.rb
JonRowe: I'd just revert the other changes, it's a well known technique

pirj:
Updates:
 - accepted proposed changes (Changelog)
 - reverted `class << ...; self; end` -> `singleton_class` changes

lib/rspec/mocks/method_double.rb
```diff
@@ -215,72 +218,39 @@ def raise_method_not_stubbed_error
...
-        # saving us from the cost of the expensive exception, but this error is
-        # extremely rare (it was discovered on 2014-12-30, only happens on
-        # RUBY_VERSION < 2.0 and our spec suite only hits this condition once),
-        # so we'd rather avoid the cost of that check for every method double,
-        # and risk the rare situation where this exception will get raised.
-        RSpec.warn_with(
-          "WARNING: RSpec could not fully restore #{@object.inspect}." \
-          "#{@method_name}, possibly because the method has been redefined " \
-          "by something outside of RSpec."
-        )
```
JonRowe, line 283: Restore this and amend the comment? I still think it is a valid protection even if the specific example provided doesn't happen anymore?

pirj: Strongly agree 👍
A tricky situation that is out of our control.

pirj:
Update: reverted the removal of `rescue NameError` from `lib/rspec/mocks/method_double.rb`.

pirj:
I'm combining ReentrantMutex and Mutex in `rspec-support`, so going to change those two `require` statements:
```
rspec-mocks/lib/rspec/mocks/message_expectation.rb:RSpec::Support.require_rspec_support 'mutex'
rspec-mocks/lib/rspec/mocks/proxy.rb:RSpec::Support.require_rspec_support "mutex"
```

lib/rspec/mocks/method_double.rb
```diff
@@ -271,11 +259,7 @@ def remove_method_from_definition_target
         #
         # Note: we could avoid rescuing this by checking
         # `definition_target.instance_method(@method_name).owner == definition_target`,
-        # saving us from the cost of the expensive exception, but this error is
-        # extremely rare (it was discovered on 2014-12-30, only happens on
-        # RUBY_VERSION < 2.0 and our spec suite only hits this condition once),
-        # so we'd rather avoid the cost of that check for every method double,
-        # and risk the rare situation where this exception will get raised.
+        # but this error is extremely rare, so we'd rather rescue this exception.
```
JonRowe, line 262:
```suggestion
        # saving us from the cost of the expensive exception, but this error is
        # extremely rare so we'd rather avoid the cost of that check for every
        # method double, and risk the rare situation where this exception will
        # get raised. This was originally discovered in the core library of older
        # unsupported Rubies, (< 2.0) but could happen in code under test
        # during meta-programming.
```

pirj: Brilliant, thanks!

JonRowe (approved):

PR #1357, opened 2020-12-06
raise ArgumentError if object cannot be proxied

### LINKED ISSUE #1356: TypeError: can't define singleton
### Subject of the issue
Try to mock class not by an object, but by a symbol and get very uninformative error message.
Maybe we can improve such behaviour and leave for programmer better error message, like
> Tried to mock symbol, but object expected.

Of course we can go to [stackoverflow](https://stackoverflow.com/questions/47080835/typeerror-cant-define-singleton-rspec) with some googling, but on my opinion this error message very unfriendly and error-phone.

### Your environment
* Ruby version: 2.7.0
* rspec-mocks version: 2.14.6

### Steps to reproduce
```ruby
allow(:subject).to receive(:foo).and_return(:baz)
```

### Expected behavior
Some error raised

```ruby
raise ArgumentError, 'Passed not an object'
```

### Actual behavior
```
Failure/Error: allow(:action).to receive(:foo).and_return(:baz)
     TypeError:
       can't define singleton
```

zhisme:
fixes #1356

### DIFF
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -38,8 +38,13 @@ module RSpec
         return unless object.frozen?
         return if object.nil?
 
-        raise ArgumentError, 'Cannot proxy frozen object.' \
-          ' Seems like you passed symbol but object expected'
+        msg = "Cannot proxy frozen objects"
+        if Symbol === object
+          msg << ". Symbols such as #{object} cannot be mocked or stubbed."
+        else
+          msg << ", rspec-mocks relies on proxies for method stubbing and expectations."
+        end
+        raise ArgumentError, msg
       end
 
       # @private
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -35,7 +35,12 @@ module RSpec
 
       # @private
       def ensure_can_be_proxied!(object)
-        return unless object.frozen?
+        if RUBY_VERSION.to_f < 2.0
+          return if !object.is_a?(Symbol) && !object.frozen?
+        else
+          return unless object.frozen?
+        end
+
         return if object.nil?
 
         msg = "Cannot proxy frozen objects"
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -35,12 +35,7 @@ module RSpec
 
       # @private
       def ensure_can_be_proxied!(object)
-        if RUBY_VERSION.to_f < 2.0
-          return if !object.is_a?(Symbol) && !object.frozen?
-        else
-          return unless object.frozen?
-        end
-
+        return unless object.is_a?(Symbol) || object.frozen?
         return if object.nil?
 
         msg = "Cannot proxy frozen objects"
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -21,6 +21,8 @@ module RSpec
 
       # @private
       def initialize(object, order_group, options={})
+        ensure_can_be_proxied!(object)
+
         @object = object
         @order_group = order_group
         @error_generator = ErrorGenerator.new(object)
@@ -29,6 +31,15 @@ module RSpec
         @options = options
         @null_object = false
         @method_doubles = Hash.new { |h, k| h[k] = MethodDouble.new(@object, k, self) }
+      end
+
+      # @private
+      def ensure_can_be_proxied!(object)
+        return unless object.frozen?
+        return if object.nil?
+
+        raise ArgumentError, 'Cannot proxy frozen object.' \
+          ' Seems like you passed symbol but object expected'
       end
 
       # @private
--- a/spec/rspec/mocks/space_spec.rb
+++ b/spec/rspec/mocks/space_spec.rb
@@ -232,6 +232,13 @@ module RSpec::Mocks
       }.to raise_error(RSpec::Expectations::ExpectationNotMetError, /Diff/)
     end
 
+    it 'raises ArgumentError if object cannot be proxied' do
+      space1 = Space.new
+
+      expect { space1.proxy_for(:subject) }.to raise_error(ArgumentError)
+      expect { space1.proxy_for("subject".freeze) }.to raise_error(ArgumentError)
+    end
+
     def in_new_space_scope
       RSpec::Mocks.setup
       yield
--- a/spec/rspec/mocks/space_spec.rb
+++ b/spec/rspec/mocks/space_spec.rb
@@ -232,11 +232,20 @@ module RSpec::Mocks
       }.to raise_error(RSpec::Expectations::ExpectationNotMetError, /Diff/)
     end
 
-    it 'raises ArgumentError if object cannot be proxied' do
+    it 'raises ArgumentError with message if object is symbol' do
       space1 = Space.new
+      object = :subject
+      expected_message = "Cannot proxy frozen objects. Symbols such as #{object} cannot be mocked or stubbed."
 
-      expect { space1.proxy_for(:subject) }.to raise_error(ArgumentError)
-      expect { space1.proxy_for("subject".freeze) }.to raise_error(ArgumentError)
+      expect { space1.proxy_for(object) }.to raise_error(ArgumentError, expected_message)
+    end
+
+    it 'raises ArgumentError with message if object is frozen' do
+      space1 = Space.new
+      object = "subject".freeze
+      expected_message = "Cannot proxy frozen objects, rspec-mocks relies on proxies for method stubbing and expectations."
+
+      expect { space1.proxy_for(object) }.to raise_error(ArgumentError, expected_message)
     end
 
     def in_new_space_scope
--- a/spec/rspec/mocks/syntax_agnostic_message_matchers_spec.rb
+++ b/spec/rspec/mocks/syntax_agnostic_message_matchers_spec.rb
@@ -43,11 +43,9 @@ module RSpec
         reset subject
       end
 
-      context 'when subject cannot be proxied' do
-        let(:subject) { :subject }
-
+      context 'when target cannot be proxied' do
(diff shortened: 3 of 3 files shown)


spec/rspec/mocks/space_spec.rb
```diff
@@ -232,6 +232,13 @@ def create_generations
       }.to raise_error(RSpec::Expectations::ExpectationNotMetError, /Diff/)
     end
 
+    it 'raises ArgumentError if object cannot be proxied' do
+      space1 = Space.new
+
+      expect { space1.proxy_for(:subject) }.to raise_error(ArgumentError)
+      expect { space1.proxy_for("subject".freeze) }.to raise_error(ArgumentError)
```
JonRowe, line 239: These should assert on part of the message raised.

zhisme:
@pirj can you take a look?

however I cannot understand why travis CI failed, I will try to reproduce with the given ruby versions on localhost to identify what's wrong

lib/rspec/mocks/proxy.rb
```diff
@@ -31,6 +33,15 @@ def initialize(object, order_group, options={})
         @method_doubles = Hash.new { |h, k| h[k] = MethodDouble.new(@object, k, self) }
       end
 
+      # @private
+      def ensure_can_be_proxied!(object)
+        return unless object.frozen?
+        return if object.nil?
+
+        raise ArgumentError, 'Cannot proxy frozen object.' \
+          ' Seems like you passed symbol but object expected'
```
JonRowe, line 42:
```suggestion
        msg = "Cannot proxy frozen objects"
        if Symbol === object
          msg << ". Symbols such as #{object} cannot be mocked or stubbed."
        else
          msg << ", rspec-mocks relies on proxies for method stubbing and expectations."
        end
        raise ArgumentError, msg
```

spec/rspec/mocks/syntax_agnostic_message_matchers_spec.rb
```diff
@@ -42,6 +42,14 @@ module Mocks
 
         reset subject
       end
+
+      context 'when subject cannot be proxied' do
+        let(:subject) { :subject }
+
+        it 'raises ArgumentError with message' do
+          expect { ::RSpec::Mocks.allow_message(subject, :foo) { :a } }.to raise_error(ArgumentError)
```
JonRowe, line 50:
```suggestion
      context 'when target cannot be proxied' do
        it 'raises ArgumentError with message' do
          expect { ::RSpec::Mocks.allow_message(:subject, :foo) { :a } }.to raise_error(ArgumentError)
```

spec/rspec/mocks/syntax_agnostic_message_matchers_spec.rb
```diff
@@ -91,6 +99,14 @@ module Mocks
 
         verify subject
       end
+
+      context 'when subject cannot be proxied' do
+        let(:subject) { :subject }
+
+        it 'raises ArgumentError with message' do
+          expect { ::RSpec::Mocks.expect_message(subject, :foo) { :a } }.to raise_error(ArgumentError)
+        end
```
JonRowe, line 108:
```suggestion
      context 'when target cannot be proxied' do
        it 'raises ArgumentError with message' do
          expect { ::RSpec::Mocks.allow_message(:subject, :foo) { :a } }.to raise_error(ArgumentError)
```

JonRowe (changes requested):
Thanks for tackling this, some feedback

zhisme:
thanks @JonRowe will update

zhisme:
I couldn't find changelog for ruby 2.0.0, but surely in previous ruby versions symbols where not frozen by default, that's why travis was failing for older ruby versions.
consider this example, I reproduced and fixed with ruby 1.9.3

```ruby
irb(main):001:0> RUBY_VERSION
=> "1.9.3"
irb(main):002:0> :subject.frozen?
=> false
```

zhisme:
@pirj @JonRowe and I see on another PR #1349 you planning to drop ruby < 2.3, maybe there is no need to adjust this code to be compatible with older ruby versions. what do you think?

JonRowe:
@zhisme If you wish to target this off `4-0-dev` for release in `4.0.0` you are welcome to drop the compatibility code, we plan to release around the launch of Ruby 3.0.0 but there is no guaranteed timeline for it. I will probably backport this to main however for release in a bugpatch in the 3.x series.

lib/rspec/mocks/proxy.rb
```diff
@@ -31,6 +33,25 @@ def initialize(object, order_group, options={})
         @method_doubles = Hash.new { |h, k| h[k] = MethodDouble.new(@object, k, self) }
       end
 
+      # @private
+      def ensure_can_be_proxied!(object)
+        if RUBY_VERSION.to_f < 2.0
+          return if !object.is_a?(Symbol) && !object.frozen?
+        else
+          return unless object.frozen?
+        end
```
JonRowe, line 42: I'm ok with this check all the time to be honest.
```suggestion
        return unless object.is_a?(Symbol) || object.frozen?
```

zhisme: applied your variant

zhisme:
@JonRowe  any idea of travis-ci build error?
And other seems like failed due to this one. Need your help
```notext
Failure/Error: require 'thread_order'
LoadError:
  cannot load such file -- thread_order
```

JonRowe:
Thanks, don't worry about Travis we're phasing it out as its become really unreliable since they essentially dropped open source.

keegangroth:
Hi, I've just pulled this in and am getting test failures in areas where I've stubbed destroyed ActiveRecord objects, which freezes them. I'm not understanding why rspec should refuse to stub these objects and I'm not seeing any explanation here, but hoping someone could explain it to me?

JonRowe:
@keegangroth Frozen objects cannot be modified, which is required to stub them.

keegangroth:
OK, I think there's something specific to ActiveRecord here, because old and new versions of rspec both throw errors for Hash objects. I'll see if I can come up with a trivial example for activerecord, here's something that behaves more or less the same both before and after this change.

```
describe 'frozen stubbing' do
  let(:h) { Hash.new }
  let(:error) { StandardError.new 'mine' }

  context 'expect_any_instance_of' do
    before do
      expect_any_instance_of(Hash).to receive(:map).and_raise(error)
    end

    # fails with either
    # expected #<StandardError: mine>, got #<ArgumentError: Cannot proxy frozen objects, rspec-mocks relies on proxies for method stubbing and expectations. Hash: {}>
    # or 
    # expected #<StandardError: mine>, got #<FrozenError: can't modify frozen class/module>
    it 'fails' do
      h.freeze
      expect{h.map}.to raise_error(error)
    end
  end

  context 'instance stub' do
    before do
      expect(h).to receive(:map).and_raise(error)
    end

    it 'works' do
      h.freeze
      expect{h.map}.to raise_error(error)
    end
  end
end
```

keegangroth:
The behavior has definitely changed for destroyed ActiveRecord Models. I don't know what makes their version of freezing different from hash. Trivial example using rails to do the db setup:

```
class Human < ApplicationRecord
  def my_method
    10
  end
end


RSpec.describe Human, type: :model do
  let(:h) { described_class.create! name: 'Steve' }
  let(:error) { StandardError.new 'mine' }

  context 'expect_any_instance_of' do
    before do
      expect_any_instance_of(described_class).to receive(:my_method).and_raise(error)
    end

    it 'fails' do
      h.destroy!
      expect{h.my_method}.to raise_error(error)
    end
  end

  context 'instance stub' do
    before do
      expect(h).to receive(:my_method).and_raise(error)
    end

    it 'works' do
      h.destroy!
      expect{h.my_method}.to raise_error(error)
    end
  end
end
```

result on rspec-mocks 3.9.1: 
```
Finished in 0.0497 seconds (files took 0.82981 seconds to load)
2 examples, 0 failures
```
result on rspec-mocks 3.10.1:
```
Failures:

  1) Human expect_any_instance_of fails
     Failure/Error: expect{h.my_method}.to raise_error(error)

       expected #<StandardError: mine>, got #<ArgumentError: Cannot proxy frozen objects, rspec-mocks relies on proxies for method stubbing and expectations.> with backtrace:
         # ./spec/models/human_spec.rb:14:in `block (4 levels) in <top (required)>'
         # ./spec/models/human_spec.rb:14:in `block (3 levels) in <top (required)>'
     # ./spec/models/human_spec.rb:14:in `block (3 levels) in <top (required)>'

Finished in 0.04758 seconds (files took 0.87952 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./spec/models/human_spec.rb:12 # Human expect_any_instance_of fails
```

Probably warrants it's own (new) issue? Kind of niche, but it's now a major headache to test certain kinds of functionality in active record (I actually haven't yet figured out the work around for my particular case).

flugsio:
The activerecord models are not really frozen, they override that method and delegates it to the attributes, which are frozen when the record is destroyed. https://github.com/rails/rails/blob/914caca2d31bd753f47f9168f2a375921d9e91cc/activerecord/lib/active_record/core.rb#L618
```ruby
def frozen?
  @attributes.frozen?
end
```

I'm not sure if there's a better way to check the real frozen status for the model itself.

JonRowe:
Feel free to raise that as a bug with the Rails team, if an object isn't actually frozen it probably shouldn't return that it is.

keegangroth:
I don't mind raising this with rails, but does any really think they're going to change their interface for this? Seems like there are some pretty simple alternatives here in rspec such as just catching the FrozenError (which looks like it was always raised in these cases) and re-raising with the appropriate message.

JonRowe:
They might not be aware they are breaking an expected Ruby behaviour, they might have a reason for it being the way it is, I don't know, but currently these objects don't behave as they should by the standards Ruby sets, its worth finding out if this is blocking you, equally If you'd like to create an alternative implementation for this warning feel free.

keegangroth:
#1401 for what it's worth

PR #1365, opened 2020-12-18
Scrape out monkey patching

pirj:
Sibling PRs:
 - https://github.com/rspec/rspec-expectations/pull/1245
 - https://github.com/rspec/rspec-core/pull/2803
 - https://github.com/rspec/rspec-support/pull/462

Prerequisite PRs:
 - https://github.com/rspec/rspec-rails/pull/2420
 - https://github.com/rspec/rspec-support/pull/457
 - https://github.com/rspec/rspec-rails/pull/2425

- [ ] Remove temporary commit. CI is expected to fail on Core/Mocks/Support/Rails steps, as `disable_monkey_patching!` calls methods that were removed with no replacement. I'll use the `maintenance-branch` temporary commit hack to fraternize those sibling PRs.

http://rspec.info/blog/2013/07/the-plan-for-rspec-3/#zero_monkey_patching_mode

> we do want to encourage people to switch to the new syntax, so we plan to make RSpec 3 print a warning on first usage of any the old syntax methods (should, should_not, should_receive, etc) unless the should syntax has been explicitly enabled. This should nudge folks towards the new syntax while keeping RSpec friendly to new users and will pave the way for the old syntax to be disabled by default in RSpec 4.

> zero-monkey-patching mode for RSpec...  We plan for these config options to become the defaults in RSpec 4.0, so that RSpec 4.0 will have zero monkey patching out of the box.

As for "disabled by default" vs "completely removed" and "default, out of the box" vs "impossible" I can only say that RSpec 4 was probably planned to be released earlier, as:

> we'll probably be dropping support for 1.8.7 in RSpec 4

but we've also dropped 1.9, 2.0, 2.1 and 2.2

https://github.com/rspec/rspec-core/issues/2301#issuecomment-234612382

> In RSpec 4, we plan to extract all monkey patching from RSpec and move it into a separate gem, so that monkey patching is opt-in instead of opt-out and users have to explicitly install and load a gem to get it.

`rspec-should` (or `rspec-monkey` as it's also about exposing example group DSL in the top-level/Module?) will be released later.

Those using the monkey-patched `should` syntax are not encouraged to update to RSpec 4 until this gem is extracted.


JonRowe (approved):

maintenance-branch
```diff
@@ -1 +1 @@
-4-0-dev
\ No newline at end of file
+remove-monkey-patching
```
JonRowe, line 1: Whats this dependant on?

pirj: I
Core should be merged first.
Core's `4-0-dev` has this:
```ruby
        RSpec::Mocks.configuration.tap do |config|
          config.syntax = :expect
```
and would fail if this was merged first.
Core's `remove-monkey-patching`'s specs pass when run against `4-0-dev` of Mock and Expectations.

Mocks and Expectations' PRs can be merged in any order.

JonRowe: Ah ok, this explanation resolves a point I had on core, in future it'd be useful if this sort of explicit information is in the PR description, or commented on the branch like this so people know why, I have an alternate solution.

JonRowe:
Merging as build failure is rspec-rails on Ruby 3.1 (head) due to no available minitest version.

PR #1385, opened 2020-12-29
Resolve an issue using method doubles with keyword arguments on Ruby 3

bryanp:
This captures and resolves an issue I encountered when running my test suite on Ruby 3. It may resolve more issues than just with `and_call_original` but that needs more investigation. Happy to fill in more test cases here but I haven't gone through to see what else uses `MethodDouble` yet.

### DIFF
--- a/lib/rspec/mocks/method_double.rb
+++ b/lib/rspec/mocks/method_double.rb
@@ -63,6 +63,7 @@ module RSpec
           define_method(method_name) do |*args, &block|
             method_double.proxy_method_invoked(self, *args, &block)
           end
+          ruby2_keywords(method_name)
           __send__(visibility, method_name)
         end
 
--- a/lib/rspec/mocks/method_double.rb
+++ b/lib/rspec/mocks/method_double.rb
@@ -63,7 +63,7 @@ module RSpec
           define_method(method_name) do |*args, &block|
             method_double.proxy_method_invoked(self, *args, &block)
           end
-          ruby2_keywords(method_name) if respond_to?(:ruby2_keywords, true)
+          ruby2_keywords(method_name) if Module.private_method_defined?(:ruby2_keywords)
           __send__(visibility, method_name)
         end
 
--- a/lib/rspec/mocks/method_double.rb
+++ b/lib/rspec/mocks/method_double.rb
@@ -63,7 +63,7 @@ module RSpec
           define_method(method_name) do |*args, &block|
             method_double.proxy_method_invoked(self, *args, &block)
           end
-          ruby2_keywords(method_name)
+          ruby2_keywords(method_name) if respond_to?(:ruby2_keywords, true)
           __send__(visibility, method_name)
         end
 
--- a/spec/rspec/mocks/and_call_original_spec.rb
+++ b/spec/rspec/mocks/and_call_original_spec.rb
@@ -185,10 +185,14 @@ RSpec.describe "and_call_original" do
       expect(klazz.alternate_new).to be_an_instance_of(klazz)
     end
 
-    it "works for methods that accept keyword arguments" do
-      def instance.foo(bar:); bar; end
-      expect(instance).to receive(:foo).and_call_original
-      expect(instance.foo(bar: "baz")).to eq("baz")
+    if RSpec::Support::RubyFeatures.kw_args_supported?
+      binding.eval(<<-CODE, __FILE__, __LINE__)
+      it "works for methods that accept keyword arguments" do
+        def instance.foo(bar:); bar; end
+        expect(instance).to receive(:foo).and_call_original
+        expect(instance.foo(bar: "baz")).to eq("baz")
+      end
+      CODE
     end
 
     context 'on an object that defines method_missing' do
--- a/spec/rspec/mocks/and_call_original_spec.rb
+++ b/spec/rspec/mocks/and_call_original_spec.rb
@@ -188,7 +188,7 @@ RSpec.describe "and_call_original" do
     if RSpec::Support::RubyFeatures.kw_args_supported?
       binding.eval(<<-CODE, __FILE__, __LINE__)
       it "works for methods that accept keyword arguments" do
-        def instance.foo(bar:); bar; end
+        def instance.foo(bar: nil); bar; end
         expect(instance).to receive(:foo).and_call_original
         expect(instance.foo(bar: "baz")).to eq("baz")
       end
--- a/spec/rspec/mocks/and_call_original_spec.rb
+++ b/spec/rspec/mocks/and_call_original_spec.rb
@@ -185,6 +185,12 @@ RSpec.describe "and_call_original" do
       expect(klazz.alternate_new).to be_an_instance_of(klazz)
     end
 
+    it "works for methods that accept keyword arguments" do
+      def instance.foo(bar:); bar; end
+      expect(instance).to receive(:foo).and_call_original
+      expect(instance.foo(bar: "baz")).to eq("baz")
+    end
+
     context 'on an object that defines method_missing' do
       before do
         klass.class_exec do


bryanp:
Well this approach clearly has some issues for Ruby < 2.7.

bryanp:
https://github.com/rspec/rspec-mocks/pull/1385/commits/1f2d12b3c10d7e0de3cf7c4ce0910fd64039a0c6 fixes some of the test failures, but I'd prefer to use the [ruby2_keywords](https://rubygems.org/gems/ruby2_keywords) shim. Are we okay adding that dependency?

lib/rspec/mocks/method_double.rb
```diff
@@ -63,6 +63,7 @@ def define_proxy_method
           define_method(method_name) do |*args, &block|
             method_double.proxy_method_invoked(self, *args, &block)
           end
+          ruby2_keywords(method_name) if respond_to?(:ruby2_keywords, true)
```
JonRowe, line 66: Aha! Thats where its needed.

spec/rspec/mocks/and_call_original_spec.rb
```diff
@@ -185,6 +185,12 @@ class << self
       expect(klazz.alternate_new).to be_an_instance_of(klazz)
     end
 
+    it "works for methods that accept keyword arguments" do
+      def instance.foo(bar:); bar; end
+      expect(instance).to receive(:foo).and_call_original
+      expect(instance.foo(bar: "baz")).to eq("baz")
+    end
```
JonRowe, line 192: This needs to be wrapped in an eval for our older ruby test suite, and skipped when theres no keyword argument support. (RSpec 3 supports 1.8.7 up).

pirj: An example https://github.com/rspec/rspec-expectations/blob/955151da74fd6356e614e62addb3da26ea1ccaf6/spec/rspec/matchers/built_in/has_spec.rb#L20

pirj:
Triggered a dummy build to figure out if those failures are due to the change or unrelated https://github.com/rspec/rspec-mocks/pull/1386

lib/rspec/mocks/method_double.rb
pirj: @bryanp 👏 

Just wondering how did you figure it out? 2.7.1's warning traces are kind of unhelpful.

bryanp: @pirj I found a test case that raised an exception, so there was a pretty strong hint in the backtrace:

```
     ArgumentError:
       wrong number of arguments (given 1, expected 0; required keyword: bar)
     # ./spec/rspec/mocks/and_call_original_spec.rb:189:in `foo'
     # ./lib/rspec/mocks/message_expectation.rb:101:in `call'
     # ./lib/rspec/mocks/message_expectation.rb:101:in `block in and_call_original'
     # ./lib/rspec/mocks/message_expectation.rb:740:in `call'
     # ./lib/rspec/mocks/message_expectation.rb:572:in `invoke_incrementing_actual_calls_by'
     # ./lib/rspec/mocks/message_expectation.rb:427:in `invoke'
     # ./lib/rspec/mocks/proxy.rb:217:in `message_received'
     # ./lib/rspec/mocks/proxy.rb:361:in `message_received'
     # ./lib/rspec/mocks/method_double.rb:78:in `proxy_method_invoked'
     # ./lib/rspec/mocks/method_double.rb:64:in `block (2 levels) in define_proxy_method'
     # ./spec/rspec/mocks/and_call_original_spec.rb:191:in `block (3 levels) in <top (required)>'
```

spec/rspec/mocks/and_call_original_spec.rb
bryanp: Ah, makes sense. Did that here: https://github.com/rspec/rspec-mocks/pull/1385/commits/1433a4299d726e236ab3d661fc54c73518b96e80

bryanp:
Addressed feedback so far but still getting failures that _seem_ unrelated to my change. Ideas?

lib/rspec/mocks/method_double.rb
JonRowe: Ah yes, sadly 2.7 is less helpful

JonRowe:
So the build failure is from this change, I can reproduce it on a clean build with only that line changing (spec failed 3.0 and 2.7 but passed 2.5 2.6, add that one line and 3.0 and 2.7 pass but 2.5 2.6 fails)

bryanp:
Have you reproduced the failures locally? I've only seen them on CI so far.

JonRowe:
@bryanp yes, they are from rspec-rails when combined with this branch

bryanp:
@JonRowe What's the best way to locally run the `rspec-rails` specs against this branch?

lib/rspec/mocks/method_double.rb
mame: Sorry for the inconvenience 🙏
Just a tip: if you run it on Ruby 3.0, it will raise an exception instead of a warning, which will show a more helpful backtrace.

bryanp:
Aha, finally reproduced the failure. I'm looking specifically at this one: 

```
rspec ./spec/rspec/rails/example/view_example_group_spec.rb:114 # RSpec::Rails::ViewExampleGroup routes helpers collides with asset helpers uses routes helpers
```

Here's the important part of the failure:

```
TypeError:
       can't create instance of singleton class
     # /Users/bryanp/.gem/ruby/2.6.6/gems/railties-6.0.3.4/lib/rails/railtie.rb:167:in `new'
     # /Users/bryanp/.gem/ruby/2.6.6/gems/railties-6.0.3.4/lib/rails/railtie.rb:167:in `instance'
     # /Users/bryanp/.gem/ruby/2.6.6/gems/railties-6.0.3.4/lib/rails/application.rb:100:in `instance'
     # /Users/bryanp/.gem/ruby/2.6.6/gems/railties-6.0.3.4/lib/rails/railtie.rb:183:in `respond_to_missing?'
     # /Users/bryanp/Code/Forks/rspec-mocks/lib/rspec/mocks/method_double.rb:67:in `respond_to?'
     # /Users/bryanp/Code/Forks/rspec-mocks/lib/rspec/mocks/method_double.rb:67:in `block in define_proxy_method'
```

Couple of observations:

1. `ruby2_keywords` calls `respond_to?`.
2. The context we call `respond_to?` on is a `RSpecRails::Application` class, which is a `Railtie`.
3. `Railtie::respond_to_missing?` tries to create an instance, which we can't do because the context is a `Singleton`.

I don't yet understand how `RSpecRails::Application` is created, but an idea is to redefine `::instance` to behave like `Singleton::instance` rather than `Railtie::instance`. There may be a more obvious solution that I'm missing here.

pirj:
@bryanp Interesting!
We call `if respond_to?` ourselves there, it's not `ruby2_keywords` itself.
Do you think we can add the check for `respond_to?` somewhere up the chain? Rubies 2.5 & 2.6 don't have it.

Or probably use `Module.respond_to?(:ruby2_keywords, true)` and rely on the assumption that if `Module` responds to it, other classes do as well?

bryanp:
@pirj lol, you're right we do call `respond_to?`. I should have finished my coffee before diving into this today :)

I like the `Module.respond_to?(:ruby2_keywords, true)` idea, but maybe we should use `TOPLEVEL_BINDING.receiver` instead? That's the approach used in the `ruby2_keywords` gem [here](https://github.com/ruby/ruby2_keywords/blob/23981c5296aec6c5dbe104b8adc7ca0e85cb4313/lib/ruby2_keywords.rb#L10-L15).

pirj:
I have some serious problem understanding what is going on in that gem while looking at [this line](https://github.com/ruby/ruby2_keywords/blob/23981c5296aec6c5dbe104b8adc7ca0e85cb4313/lib/ruby2_keywords.rb#L29) :smile:

I don't mind either option. To me, `Module.respond_to?` is more compact.

@mame, can you please advise what is the best option for our case?
We are trying to call `ruby2_keywords`, but only if the object responds to it. [`Railtie#respond_to_missing?`](https://github.com/rails/rails/blob/5710128a84d5d03646dca6118db0e3a5715904ec/railties/lib/rails/railtie.rb#L199) is calling `instance` and fails.

In order to avoid that `respond_to?` call on the current object, is it ok to call it on `Module`, `Object`, or `TOPLEVEL_BINDING.receiver` instead?

JonRowe:
We do use the "unbind the real method and use that" to work around monkey patching in various places, maybe we need that trick in `rspec-support` for defining keyword support?

mame:
Hello!

I think `Module.private_method_defined?(:ruby2_keywords)` is the best option. If directly calling `ruby2_keywords` on Railtie is troublesome, `Module.instance_method(:ruby2_keywords).bind(self).call(method_name)` may work, as @JonRowe said.

Sorry, [the check in `ruby2_keywords` gem](https://github.com/ruby/ruby2_keywords/blob/23981c5296aec6c5dbe104b8adc7ca0e85cb4313/lib/ruby2_keywords.rb#L2) looks wrong. `Module.respond_to?(:ruby2_keywords)` does NOT check if `Module#ruby2_keywords` is available. Fortunately, it works well because the check returns true in Ruby 2.7 or later due to the toplevel `ruby2_keywords` method. And thank you for letting me know about the syntactic error 🤣 I will talk with @nobu and will fix.

bryanp:
Thank you @mame for you input. I've made the suggested change.

This resolved the issues in Ruby 2.* but 1.8.7 and 1.9.2 are still failing in Travis. I haven't looked into why yet.

mame:
Looks like this issue is unrelated to this PR, this time for sure?

https://travis-ci.org/github/rspec/rspec-mocks/jobs/753288703
```
$ gem install bundler -v '< 2'
ERROR:  Could not find a valid gem 'bundler' (< 2), here is why:
          Unable to download data from https://rubygems.org/ - hostname was not match with the server certificate (https://api.rubygems.org/specs.4.8.gz)
The command "gem install bundler -v '< 2'" failed and exited with 2 during .
```

pirj:
There were some changes to `.travis.yml`, I've rebased this branch on `main` and force-pushed it.
However, chances of fixing the issue are quite low, since judging by the list of jobs (1.8.7, 1.9.2, ree, jruby-1.7) that the [failing build](https://travis-ci.org/github/rspec/rspec-mocks/builds/753288701) was using the [`.travis.yml` from `main`](https://github.com/rspec/rspec-mocks/blob/6ab343ca479c606e892c82ce0245e7410e3f0eac/.travis.yml#L17) anyway, not [the one from *this* branch](https://github.com/rspec/rspec-mocks/blob/85504dad67f34508d99032f566a619b29dd340e8/.travis.yml#L17) that also lists 1.9.3 and 2.0.0. I vaguely remember that was the case, can't remember why this is the case.

JonRowe:
Travis is broken and I'm replacing the builds as I can.

pirj:
@mame Thanks a lot for your help!

@bryanp Thanks a lot for dealing with this issue!

berkos:
Thanks for the fix! I can verify that it works in Ruby 3.0.0. When do you plan to release this in RubyGems. 🙏

eregon:
I've run into this weird issue with `respond_to?(:ruby2_keywords, true)` as well in #1464 and wondered for a while what the heck was going on. `bundle exec rspec -b spec ` in rspec-rails helps to see the first error, which helps. The following errors are all completely undecipherable.

For some reason it only happens on 2.5/2.6, I guess because 1) 2.7+ have the methods, so don't fall into the trap, and 2) only those Ruby versions test the Rails version with this weird behavior? (which still exists on [Rails main](https://github.com/rails/rails/blob/ad25dfad8797b5024d90592bdb6a5b2d725a4850/railties/lib/rails/railtie.rb#L216-L218))
What's pretty crazy here is that the Railtie class defines `respond_to_missing?` on the Railtie class, not for instances, and try to create an instance in `respond_to_missing?`, which is definitely not expected for a `respond_to_missing?` implementation.

eregon:
> Sorry, [the check in `ruby2_keywords` gem](https://github.com/ruby/ruby2_keywords/blob/23981c5296aec6c5dbe104b8adc7ca0e85cb4313/lib/ruby2_keywords.rb#L2) looks wrong. `Module.respond_to?(:ruby2_keywords)` does NOT check if `Module#ruby2_keywords` is available.

FWIW, that was fine and no bug, it finds Module#ruby2_keywords, since it's called on the Module class, which is a module (as all classes are).

eregon:
I've filed an issue to Rails about this, and found at least one spot where they unintentionally create a Railtie instance by a `respond_to?` check: https://github.com/rails/rails/issues/44761

PR #1394, opened 2021-01-06
Make `with` support Ruby 3 keywords

mame:
As you know, keyword arguments have been separated from positional ones in Ruby 3.
How about making `with` to match the semantics?
It would be good to allow rspec to check if positional or keyword arguments are passed appropriately.

For example, if a mock object has `with(a: 'a', b: 'b'), it should accept keywords, not a positional Hash.

```
class Foo
end

foo = Foo.new
allow(foo).to receive(:bar).with(a: 'a', b: 'b')
args = { a: 'a', b: 'b' }

foo.bar(args)   # fail
foo.bar(**args) # pass
```

Also, if a mock object has `with({ a: 'a', b: 'b' }), it should accept a positional Hash, not keywords.

```
foo = Foo.new
allow(foo).to receive(:bar).with({ a: 'a', b: 'b' })
args = { a: 'a', b: 'b' }

foo.bar(args)   # pass
foo.bar(**args) # fail
```

This PR is just a proof of concept and the change partially includes #1385. What do you think?

### DIFF
--- a/lib/rspec/mocks/argument_list_matcher.rb
+++ b/lib/rspec/mocks/argument_list_matcher.rb
@@ -46,16 +46,30 @@ module RSpec
         @expected_args = expected_args
         ensure_expected_args_valid!
       end
+      ruby2_keywords :initialize if Module.private_method_defined?(:ruby2_keywords)
 
       # @api public
-      # @param [Array] args
+      # @param [Array] actual_args
       #
       # Matches each element in the `expected_args` against the element in the same
       # position of the arguments passed to `new`.
       #
       # @see #initialize
-      def args_match?(*args)
-        Support::FuzzyMatcher.values_match?(resolve_expected_args_based_on(args), args)
+      def args_match?(*actual_args)
+        expected_args = resolve_expected_args_based_on(actual_args)
+
+        return false if expected_args.size != actual_args.size
+
+        if Hash.respond_to?(:ruby2_keywords_hash?, true)
+          # if both arguments end with Hashes, and if one is a keyword hash and the other is not, they don't match
+          if Hash === expected_args.last && Hash === actual_args.last
+            if Hash.ruby2_keywords_hash?(actual_args.last) != Hash.ruby2_keywords_hash?(expected_args.last)
+              return false
+            end
+          end
+        end
+
+        Support::FuzzyMatcher.values_match?(expected_args, actual_args)
       end
 
       # @private
--- a/lib/rspec/mocks/argument_list_matcher.rb
+++ b/lib/rspec/mocks/argument_list_matcher.rb
@@ -63,7 +63,7 @@ module RSpec
         if Hash.respond_to?(:ruby2_keywords_hash?, true)
           # if both arguments end with Hashes, and if one is a keyword hash and the other is not, they don't match
           if Hash === expected_args.last && Hash === actual_args.last
-            if Hash.ruby2_keywords_hash?(actual_args.last) != Hash.ruby2_keywords_hash?(expected_args.last)
+            if !Hash.ruby2_keywords_hash?(actual_args.last) && Hash.ruby2_keywords_hash?(expected_args.last)
               return false
             end
           end
--- a/lib/rspec/mocks/argument_list_matcher.rb
+++ b/lib/rspec/mocks/argument_list_matcher.rb
@@ -60,7 +60,7 @@ module RSpec
 
         return false if expected_args.size != actual_args.size
 
-        if Hash.respond_to?(:ruby2_keywords_hash?, true)
+        if RUBY_VERSION >= "3"
           # if both arguments end with Hashes, and if one is a keyword hash and the other is not, they don't match
           if Hash === expected_args.last && Hash === actual_args.last
             if !Hash.ruby2_keywords_hash?(actual_args.last) && Hash.ruby2_keywords_hash?(expected_args.last)
--- a/spec/rspec/mocks/argument_matchers_spec.rb
+++ b/spec/rspec/mocks/argument_matchers_spec.rb
@@ -387,8 +387,8 @@ module RSpec
           a_double.random_call(:a => "a", :b => "b")
         end
 
-        if RUBY_VERSION >= "2.7"
-          it "fails to matches against a hash submitted as a positional argument and received as keyword arguments in Ruby 2.7 or later", :reset => true do
+        if RUBY_VERSION >= "3"
+          it "fails to matches against a hash submitted as a positional argument and received as keyword arguments in Ruby 3.0 or later", :reset => true do
             opts = {:a => "a", :b => "b"}
             expect(a_double).to receive(:random_call).with(:a => "a", :b => "b")
             expect do
--- a/spec/rspec/mocks/argument_matchers_spec.rb
+++ b/spec/rspec/mocks/argument_matchers_spec.rb
@@ -381,20 +381,10 @@ module RSpec
           a_double.random_call(:a => "a", :b => "b")
         end
 
-        if RSpec::Support::RubyFeatures.required_kw_args_supported?
-          it "fails to match against a hash submitted by reference and received by value in Ruby 3", :reset => true do
-            opts = {:a => "a", :b => "b"}
-            expect(a_double).to receive(:random_call).with(opts)
-            expect do
-              a_double.random_call(:a => "a", :b => "b")
-            end.to fail_with(/expected: \(\{:a=>"a", :b=>"b"|:b=>"b", :a=>"a"\}\)/)
-          end
-        else
-          it "matches against a hash submitted by reference and received by value in Ruby 2" do
-            opts = {:a => "a", :b => "b"}
-            expect(a_double).to receive(:random_call).with(opts)
-            a_double.random_call(:a => "a", :b => "b")
-          end
+        it "matches against a hash submitted by reference and received by value (in both Ruby 2 and Ruby 3)" do
+          opts = {:a => "a", :b => "b"}
+          expect(a_double).to receive(:random_call).with(opts)
+          a_double.random_call(:a => "a", :b => "b")
         end
 
         if RUBY_VERSION >= "2.7"
--- a/spec/rspec/mocks/argument_matchers_spec.rb
+++ b/spec/rspec/mocks/argument_matchers_spec.rb
@@ -381,14 +381,14 @@ module RSpec
           a_double.random_call(:a => "a", :b => "b")
         end
 
-        it "matches against a hash submitted by reference and received by value (in both Ruby 2 and Ruby 3)" do
+        it "matches against a hash submitted as keyword arguments a and received as a positional argument (in both Ruby 2 and Ruby 3)" do
           opts = {:a => "a", :b => "b"}
           expect(a_double).to receive(:random_call).with(opts)
           a_double.random_call(:a => "a", :b => "b")
         end
 
         if RUBY_VERSION >= "2.7"
-          it "fails to match against a hash submitted by value and received by reference in Ruby 2.7 or later", :reset => true do
+          it "fails to matches against a hash submitted as a positional argument and received as keyword arguments in Ruby 2.7 or later", :reset => true do
             opts = {:a => "a", :b => "b"}
             expect(a_double).to receive(:random_call).with(:a => "a", :b => "b")
             expect do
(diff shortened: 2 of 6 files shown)


lib/rspec/mocks/argument_list_matcher.rb
```diff
@@ -54,8 +55,21 @@ def initialize(*expected_args)
...
-        Support::FuzzyMatcher.values_match?(resolve_expected_args_based_on(args), args)
+      def args_match?(*actual_args)
+        expected_args = resolve_expected_args_based_on(actual_args)
+
+        return false if expected_args.size != actual_args.size
+
+        if Hash.respond_to?(:ruby2_keywords_hash?, true)
+          # if both arguments end with Hashes, and if one is a keyword hash and the other is not, they don't match
+          if Hash === expected_args.last && Hash === actual_args.last
+            if Hash.ruby2_keywords_hash?(actual_args.last) != Hash.ruby2_keywords_hash?(expected_args.last)
```
pirj, line 66: Nice!

spec/rspec/mocks/argument_matchers_spec.rb
```diff
@@ -381,16 +381,36 @@ def ==(other)
           a_double.random_call(:a => "a", :b => "b")
         end
 
-        it "matches against a hash submitted by reference and received by value" do
-          opts = {:a => "a", :b => "b"}
-          expect(a_double).to receive(:random_call).with(opts)
-          a_double.random_call(:a => "a", :b => "b")
+        if RUBY_VERSION < '3'
+          it "matches against a hash submitted by reference and received by value in Ruby 2" do
```
eregon, line 385: It seems an issue of the original description, but I think "by value" and "by reference" are not really good terms to describe this (in general, everything is by passed reference in Ruby).
"positional hash" and "keyword arguments" seems far better (which are the terms used in https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/).

mame: I agree. If rspec developers also agree, I'll change the description like this.

> it "matches against a hash submitted as a positional argument and received as a keyword argument in Ruby 2"

pirj: I'm not aware of the internal implementation, and for me, as a Ruby user, this is more clear. :+1:

jeremyevans:
> Also, if a mock object has `with({ a: 'a', b: 'b' }), it should accept a positional Hash, not keywords.

I think in this case, it should accept a positional hash or keywords.  This is because Ruby 3 (and Ruby 2) will convert the keywords to a positional hash if the method does not accept keywords.

eregon:
@jeremyevans Good point, it's a scenario like:
```ruby
def foo(options)
end
# or
def foo(options = {})
end

# Both OK in all Ruby versions
foo({a: 'a'})
foo(a: 'a')
```

`with()` is a bit mind-bending because we have to guess what method parameters would correspond to it.

mame:
@jeremyevans @eregon Thanks. I've updated the PR to allow the conversion from keywords to a positional Hash.

I first thought that a `with` expectation checks actual arguments in the caller side (i.e., before the conversion) because it works even if there is no actual method definition. But, now I notice that the rspec DSL says in English `allow obj to recevie method_name with arguments`, which looks like this is a callee-side check.

mame:
I'll try fixing the CI error after #1385 is merged.

BTW, an assertion error message is less informative if this PR is accepted:

```
class Foo; end

RSpec.describe do
  it do
    foo = Foo.new
    allow(foo).to receive(:bar).with(a: 'a', b: 'b')
    foo.bar({ a: 'a', b: 'b' })
  end
end
```

```
       #<Foo:0x0000562a3c2c3ac8> received :bar with unexpected arguments
         expected: ({:a=>"a", :b=>"b"})
              got: ({:a=>"a", :b=>"b"})
```

The `expected` and `got` lines are completely the same, so it is a bit confusing. It would be good to change it to, for example,

```
       #<Foo:0x0000562a3c2c3ac8> received :bar with unexpected arguments
         expected: (a: "a", b: "b")
              got: ({:a=>"a", :b=>"b"})
```

or what style you like.

If you are okay, I have already created a patch of ruby-support to show the above message. If this PR is accepted, I'll send another PR to ruby-support. (I guess I need to update other rspec-* repositories because it will break many rspec's specs that checks an error message.)

lib/rspec/mocks/argument_list_matcher.rb
```diff
@@ -46,6 +46,7 @@ def initialize(*expected_args)
         @expected_args = expected_args
         ensure_expected_args_valid!
       end
+      ruby2_keywords :initialize if respond_to?(:ruby2_keywords, true)
 
       # @api public
       # @param [Array] args
```
pirj, line 52: I guess the build fails because the parameter was renamed in the method signature, but not here in the docs.

mame:
Sorry for too many retries, I think I managed to make it pass on Ruby 1.9.3 😅

@JonRowe @pirj Could you check this PR out? Thanks!

JonRowe:
@mame Yes it's on the list to review

benoittgt:
> The expected and got lines are completely the same, so it is a bit confusing. It would be good to change it to, for example,

I am wondering it this is something we should add to next major RSpec. Maybe it is a little bit abrupt and should deprecated first?

Thanks a lot @mame for your patch. 🙌🏻

spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -81,10 +81,15 @@ def call(name)
...
-        expect(object).to receive(:foobar).with(:key => "value").and_return(1)
-        expect(object.foobar(hash[:a])).to equal(1)
+      if RSpec::Support::RubyFeatures.required_kw_args_supported?
+        it "can accept an inner hash as a message argument" do
+          hash = {:a => {:key => "value"}}
+          expect(object).to receive(:foobar).with(:key => "value").and_return(1)
+          # Use eval to avoid syntax error on 1.8 and 1.9
+          eval("expect(object.foobar(**hash[:a])).to equal(1)")
+          # Avoid warning: assigned but unused variable - hash
+          hash = hash
```
JonRowe, line 91: If you wrap the whole spec contents in an eval don't you avoid this?

spec/rspec/mocks/should_syntax_spec.rb
```diff
@@ -466,36 +466,43 @@ def use_rspec_mocks
...
+    {:replacement => "the new `:expect` syntax or explicitly enable `:should`"}
+  }
 
-  it "it warns about should once, regardless of how many times it is called" do
-    expect(RSpec).to receive(:deprecate).with(*expected_arguments)
-    o = Object.new
-    o2 = Object.new
-    o.should_receive(:bees)
-    o2.should_receive(:bees)
+  if RSpec::Support::RubyFeatures.required_kw_args_supported?
```
JonRowe, line 475: Prehaps have an alternate for the older rubies? Not that fussed though as RSpec 4 won't support those.

JonRowe:
Can you rebase? Our CI is now fully on Github Actions so can run all the legacy rubies. If this goes green I'm inclined to merge it.

spec/rspec/mocks/partial_double_spec.rb
mame: Thanks! I've changed it as you said. https://github.com/rspec/rspec-mocks/pull/1394/commits/cdd9a0465a001f20479922ae3f6e45d1fa29f3a4

spec/rspec/mocks/should_syntax_spec.rb
mame: Okay, the old specs are restored as an alternative. https://github.com/rspec/rspec-mocks/pull/1394/commits/7c7500cb9e947a7ceb73d7cdf87924039f15031f

mame:
@JonRowe Thank you for your review! I've rebased.

mame:
The CI failed on Ruby 2.2, but I believe that this is not a fault of this PR.

https://github.com/rspec/rspec-mocks/pull/1394/checks?check_run_id=1836305443

spec/rspec/mocks/argument_matchers_spec.rb
```diff
@@ -381,16 +381,26 @@ def ==(other)
...
           expect(a_double).to receive(:random_call).with(opts)
           a_double.random_call(:a => "a", :b => "b")
         end
 
-        it "matches against a hash submitted by value and received by reference" do
-          opts = {:a => "a", :b => "b"}
-          expect(a_double).to receive(:random_call).with(:a => "a", :b => "b")
-          a_double.random_call(opts)
+        if RUBY_VERSION >= "2.7"
+          it "fails to matches against a hash submitted as a positional argument and received as keyword arguments in Ruby 2.7 or later", :reset => true do
```
JonRowe, line 391: My last concern is that this is a breaking change for 2.7, as 2.7 did accept hashes submitted in this fashion, it just issues a warning, WDYT about changing this to work for Ruby 3 only, or making it configurable for Ruby 2.7 for now, but we'll remove the config in RSpec 4 making this the permanent option?

I'm happy to tackle that in an after PR myself if you cannot.

lib/rspec/mocks/argument_list_matcher.rb
```diff
@@ -46,16 +46,30 @@ def initialize(*expected_args)
...
       #
       # @see #initialize
-      def args_match?(*args)
-        Support::FuzzyMatcher.values_match?(resolve_expected_args_based_on(args), args)
+      def args_match?(*actual_args)
+        expected_args = resolve_expected_args_based_on(actual_args)
+
+        return false if expected_args.size != actual_args.size
+
+        if Hash.respond_to?(:ruby2_keywords_hash?, true)
```
JonRowe, line 63: I think all we'd have to do is make this switch based on Ruby version (and/or optionally config)?

benoittgt:
> The CI failed on Ruby 2.2, but I believe that this is not a fault of this PR.

it's all green now. :)

ys:
Cannot wait to see this one landing! Thanks all for the work on this!

spec/rspec/mocks/argument_matchers_spec.rb
mame: Sorry for my slow reaction! I've made the change only since 3.0, and the CI has passed. I think that making it configurable for Ruby 2.7 is reasonable, but may I leave the implementation to you? Thanks!

spec/rspec/mocks/argument_matchers_spec.rb
```diff
@@ -381,16 +381,26 @@ def ==(other)
...
+        if RUBY_VERSION >= "3"
+          it "fails to matches against a hash submitted as a positional argument and received as keyword arguments in Ruby 3.0 or later", :reset => true do
+            opts = {:a => "a", :b => "b"}
+            expect(a_double).to receive(:random_call).with(:a => "a", :b => "b")
+            expect do
+              a_double.random_call(opts)
+            end.to fail_with(/expected: \(\{(:a=>\"a\", :b=>\"b\"|:b=>\"b\", :a=>\"a\")\}\)/)
+          end
+        else
+          it "matches against a hash submitted as a positional argument and received as keyword arguments in Ruby 2.6 or before" do
```
pirj, line 399: Should this say "Ruby 2.7"?

pirj (approved):
Thank you!

benoittgt (approved):
Thanks a lot @mame 🙇🏻‍♂️

spec/rspec/mocks/argument_matchers_spec.rb
mame: Fixed!

pirj: ❤️

JonRowe:
Thank you @mame!

casperisfine:
I submitted https://github.com/rspec/rspec-mocks/pull/1461 as some kind of proof of concept on how to better display the difference between keyword args and regular hashes when they are identical.

lib/rspec/mocks/argument_list_matcher.rb
```diff
@@ -46,16 +46,30 @@ def initialize(*expected_args)
...
       # @see #initialize
-      def args_match?(*args)
-        Support::FuzzyMatcher.values_match?(resolve_expected_args_based_on(args), args)
+      def args_match?(*actual_args)
+        expected_args = resolve_expected_args_based_on(actual_args)
+
+        return false if expected_args.size != actual_args.size
+
+        if RUBY_VERSION >= "3"
+          # if both arguments end with Hashes, and if one is a keyword hash and the other is not, they don't match
```
eregon, line 64: This caused quite a bit of confusion :sweat_smile: , will be fixed in https://github.com/rspec/rspec-mocks/pull/1514#discussion_r1060571780 hopefully

PR #1401, opened 2021-01-12
Be more permissive when detecting frozen objects

keegangroth:
Rather than rely on an object's #frozen? method, always attempt to
proxy a method when requested. Catch the errors resulting from failed
attempts to proxy and reraise (or log) them with helpful messages.

One notable use case for this is rails' activerecord's Model objects
becoming frozen after be deleted. In reality, only the underlying hash
of attributes is frozen and it should still be possible to write
proxies for these objects.

### DIFF
--- a/lib/rspec/mocks/method_double.rb
+++ b/lib/rspec/mocks/method_double.rb
@@ -77,7 +77,7 @@ module RSpec
         # TODO: drop in favor of FrozenError in ruby 2.5+
         #  RuntimeError (and FrozenError) for ruby 2.x
         #  TypeError for ruby 1.x
-        if FROZEN_ERROR_MSG === e.message
+        if (defined?(FrozenError) && e.is_a?(FrozenError)) || FROZEN_ERROR_MSG === e.message
           raise ArgumentError, "Cannot proxy frozen objects, rspec-mocks relies on proxies for method stubbing and expectations."
         end
         raise
@@ -105,7 +105,9 @@ module RSpec
         # TODO: drop in favor of FrozenError in ruby 2.5+
         #  RuntimeError (and FrozenError) for ruby 2.x
         #  TypeError for ruby 1.x
-        return show_frozen_warning if FROZEN_ERROR_MSG === e.message
+        if (defined?(FrozenError) && e.is_a?(FrozenError)) || FROZEN_ERROR_MSG === e.message
+          return show_frozen_warning
+        end
         raise
       end
 
--- a/lib/rspec/mocks/method_double.rb
+++ b/lib/rspec/mocks/method_double.rb
@@ -2,6 +2,9 @@ module RSpec
   module Mocks
     # @private
     class MethodDouble
+      # @private TODO: drop in favor of FrozenError in ruby 2.5+
+      FROZEN_ERROR_MSG = /can't modify frozen/
+
       # @private
       attr_reader :method_name, :object, :expectations, :stubs, :method_stasher
 
@@ -70,6 +73,14 @@ module RSpec
         end
 
         @method_is_proxied = true
+      rescue RuntimeError, TypeError => e
+        # TODO: drop in favor of FrozenError in ruby 2.5+
+        #  RuntimeError (and FrozenError) for ruby 2.x
+        #  TypeError for ruby 1.x
+        if FROZEN_ERROR_MSG === e.message
+          raise ArgumentError, "Cannot proxy frozen objects, rspec-mocks relies on proxies for method stubbing and expectations."
+        end
+        raise
       end
 
       # The implementation of the proxied method. Subclasses may override this
@@ -83,7 +94,6 @@ module RSpec
 
       # @private
       def restore_original_method
-        return show_frozen_warning if object_singleton_class.frozen?
         return unless @method_is_proxied
 
         remove_method_from_definition_target
@@ -91,6 +101,12 @@ module RSpec
         restore_original_visibility
 
         @method_is_proxied = false
+      rescue RuntimeError, TypeError => e
+        # TODO: drop in favor of FrozenError in ruby 2.5+
+        #  RuntimeError (and FrozenError) for ruby 2.x
+        #  TypeError for ruby 1.x
+        return show_frozen_warning if FROZEN_ERROR_MSG === e.message
+        raise
       end
 
       # @private
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -35,8 +35,7 @@ module RSpec
 
       # @private
       def ensure_can_be_proxied!(object)
-        return unless object.is_a?(Symbol) || object.frozen?
-        return if object.nil?
+        return unless object.is_a?(Symbol) || (object.is_a?(String) && object.frozen?)
 
         msg = "Cannot proxy frozen objects"
         if Symbol === object
--- a/lib/rspec/mocks/proxy.rb
+++ b/lib/rspec/mocks/proxy.rb
@@ -35,14 +35,9 @@ module RSpec
 
       # @private
       def ensure_can_be_proxied!(object)
-        return unless object.is_a?(Symbol) || (object.is_a?(String) && object.frozen?)
+        return unless object.is_a?(Symbol)
 
-        msg = "Cannot proxy frozen objects"
-        if Symbol === object
-          msg << ". Symbols such as #{object} cannot be mocked or stubbed."
-        else
-          msg << ", rspec-mocks relies on proxies for method stubbing and expectations."
-        end
+        msg = "Cannot proxy frozen objects. Symbols such as #{object} cannot be mocked or stubbed."
         raise ArgumentError, msg
       end
 
--- a/spec/rspec/mocks/matchers/receive_spec.rb
+++ b/spec/rspec/mocks/matchers/receive_spec.rb
@@ -285,8 +285,40 @@ module RSpec
           target.to receive(:foo).and_return(:baz)
           expect { reset object }.to change { object.foo }.from(:baz).to(:bar)
         end
+      end
 
-        context "on a frozen object" do
+      shared_examples "resets partial mocks of any instance cleanly" do
+        let(:klass)  { Struct.new(:foo) }
+        let(:object) { klass.new :bar }
+
+        it "removes the method double" do
+          target.to receive(:foo).and_return(:baz)
+          expect {
+            verify_all
+          }.to change { object.foo }.from(:baz).to(:bar)
+        end
+      end
+
+      shared_examples "handles frozen objects cleanly" do
+        let(:klass)  { Struct.new(:foo) }
+        let(:object) { klass.new :bar }
+
+        context "when adding the method double" do
+          it "raises clear error" do
+            object.freeze
+            expect {
+              target.to receive(:foo).and_return(:baz)
+            }.to raise_error(ArgumentError, /Cannot proxy frozen objects/)
+          end
+        end
+
+        context "when removing the method double" do
+          before do
+            if RUBY_VERSION < "2.2" && RUBY_VERSION >= "2.0" && RSpec::Support::Ruby.mri?
+              pending "Does not work on 2.0-2.1 because frozen structs can have methods restored"
+            end
+          end
+
           it "warns about being unable to remove the method double" do
             target.to receive(:foo).and_return(:baz)
             expect_warning_without_call_site(/rspec-mocks was unable to restore the original `foo` method on #{object.inspect}/)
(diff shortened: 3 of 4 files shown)


tannermares:
Hey there! Any reason this isn't being merged? I'm hitting this during a Rails update and could really use the fix.

lib/rspec/mocks/method_double.rb
```diff
@@ -68,6 +73,11 @@ def define_proxy_method
         end
 
         @method_is_proxied = true
+      rescue RuntimeError => e
```
pirj, line 76: I'd prefer this to catch `FrozenError`, but sadly, it was only introduced in Ruby 2.5
It is worth adding a comment, so can we change it once we drop support for older Rubies. In RSpec 4 we're dropping support for Ruby < 2.3.

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -236,6 +236,12 @@ def receiver.method_missing(*); end # a poor man's stub...
         end
 
         context "on a frozen object" do
+          before do
+            if RUBY_VERSION < "2.2" && RUBY_VERSION >= "2.0" && RSpec::Support::Ruby.mri?
```
pirj, line 240: Typically, we add [such checks to `rspec-support`](https://github.com/rspec/rspec-support/blob/main/lib/rspec/support/ruby_features.rb#L79), but I'm fine with keeping this check here as it's for specs only.

lib/rspec/mocks/proxy.rb
```diff
@@ -33,20 +31,6 @@ def initialize(object, order_group, options={})
...
-
-        msg = "Cannot proxy frozen objects"
-        if Symbol === object
-          msg << ". Symbols such as #{object} cannot be mocked or stubbed."
-        else
-          msg << ", rspec-mocks relies on proxies for method stubbing and expectations."
-        end
-        raise ArgumentError, msg
-      end
-
```
JonRowe, line 49: Restore this, if you want to relax the `object.frozen?` check do so but Symbols can't be proxied and was the original source of this code.

spec/rspec/mocks/space_spec.rb
```diff
@@ -235,17 +235,16 @@ def create_generations
     it 'raises ArgumentError with message if object is symbol' do
       space1 = Space.new
       object = :subject
-      expected_message = "Cannot proxy frozen objects. Symbols such as #{object} cannot be mocked or stubbed."
-
-      expect { space1.proxy_for(object) }.to raise_error(ArgumentError, expected_message)
+      expected_message = "Static objects such as #{object.class}:#{object} cannot be mocked or stubbed."
```
JonRowe, line 238:
```suggestion
      expected_message = "Frozen objects such as #{object.class}:#{object} cannot be mocked or stubbed."
```

spec/rspec/mocks/space_spec.rb
```diff
@@ -235,17 +235,16 @@ def create_generations
...
+      expect { space1.proxy_for(object).add_message_expectation(:foo) }.to raise_error(ArgumentError, expected_message)
     end
 
     it 'raises ArgumentError with message if object is frozen' do
       space1 = Space.new
-      object = "subject".freeze
+      object = {}.freeze
```
JonRowe, line 244: Why?

lib/rspec/mocks/method_double.rb
```diff
@@ -80,14 +90,16 @@ def proxy_method_invoked(_obj, *args, &block)
...
-        return show_frozen_warning if object_singleton_class.frozen?
         return unless @method_is_proxied
 
         remove_method_from_definition_target
         @method_stasher.restore if @method_stasher.method_is_stashed?
         restore_original_visibility
 
         @method_is_proxied = false
+      rescue RuntimeError => e
+        return show_frozen_warning if FROZEN_ERROR_MSG === e.message
```
JonRowe, line 101: Rather than checking the message which will change from time to time why not use the frozen check here? Or are you saying Rails has broken that as well?

```suggestion
        return show_frozen_warning if object_singleton_class.frozen?
```

JonRowe (changes requested):

JonRowe:
As the original source of this bug is Rails not implementing `frozen?` correctly, it might also be worth raising this with them. Only properly frozen objects should return true for `frozen?`

lib/rspec/mocks/method_double.rb
```diff
@@ -16,6 +19,8 @@ def initialize(object, method_name, proxy)
         @method_is_proxied = false
         @expectations = []
         @stubs = []
+      rescue TypeError
+        raise ArgumentError, "Static objects such as #{@object.class}:#{@object} cannot be mocked or stubbed."
```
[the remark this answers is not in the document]
keegangroth: This came from the symbol handling, I'll look to put all of that back: 

>   1) RSpec::Mocks::Space raises ArgumentError with message if object is symbol
>      Failure/Error: expect { space1.proxy_for(object).add_message_expectation(:foo) }.to raise_error(ArgumentError, expected_message)
> 
>        expected ArgumentError with "Static objects such as Symbol:subject cannot be mocked or stubbed.", got #<TypeError: can't define singleton>

lib/rspec/mocks/method_double.rb
keegangroth: In ruby 2.5 there's a specific exception class for frozen error, so the long term goal would be to use that rather than the (as noted) changeable message.

But yes, I thought it would be good if you could unstub "frozen" rails objects as well as stub them.

lib/rspec/mocks/proxy.rb
keegangroth: As noted in another comment, I changed this logic to catch the resulting TypeError, but I will revert to this way.

spec/rspec/mocks/space_spec.rb
keegangroth: It's because frozen string acts like symbol (throws a type error), I'll revert the test and add it to the symbol check

keegangroth:
Do I need to address the failures in the "legacy" tests? I've having some trouble getting rvm to install rubh-1.8.7...

keegangroth:
I think this is ready for another look. I manged to get 1.8.7 running and the tests passing there. Also rearranged and added more tests, the diff for those looks worse than it really is though.

keegangroth:
for real this time, set up the [actions on my fork](https://github.com/keegangroth/rspec-mocks/actions/runs/1872788461)  and the only failure was some ffi compilation issue on one windows run...

lib/rspec/mocks/proxy.rb
```diff
@@ -35,7 +35,7 @@ def initialize(object, order_group, options={})
 
       # @private
       def ensure_can_be_proxied!(object)
-        return unless object.is_a?(Symbol) || object.frozen?
+        return unless object.is_a?(Symbol) || (object.is_a?(String) && object.frozen?)
```
pirj, line 38: Cosmetic: This is getting hard to comprehend. Symbol and a frozen string can't be nil, can they (see a check below)?

pirj (approved):
Looks good, thank you!

lib/rspec/mocks/proxy.rb
keegangroth: Yeah agreed. These are really only here for the sake of maintaining the existing tests for these two classes. I had originally removed this method entirely because the handling in method double actually handles these same cases just fine. For example this extra test passes just fine:

```
context "when given a symbol" do
  let(:object) { :object }

  it "raises clear error" do
    expect {
      target.to receive(:foo).and_return(:baz)
    }.to raise_error(ArgumentError, /Cannot proxy frozen objects/)
  end
end
```

Just let me know what the preferred behavior for this method is and I'll do it

pirj: Understood. Let's keep the change to the minimum. I'd go with removing the redundant `nil?` check, and the rest is fine.

lib/rspec/mocks/method_double.rb
keegangroth: I thought of a way to improve this, added another commit for it. I'll also look at putting together a different PR to clean this up on the 4.x branch (I think moving to ruby 2.5+ was mentioned)

pirj:
> manged to get 1.8.7 running

Locally? Thanks for this effort, I know it is not easy.
Please accept my apologies for the delayed response and our CI build gate. I've triggered the build.

keegangroth:
> >   manged to get 1.8.7 running

> Locally? Thanks for this effort, I know it is not easy.

yeah, this took me a little while, but doing it with docker ended up not being too bad. Likely not perfect, but here's a [gist](https://gist.github.com/keegangroth/3372213e754e246eb670d68c7bdd4f1e) for it if anyone else needs it. Then you just have to find working docker images, here's what I used:
```
docker build --build-arg ruby=ruby:2.1.9 -t rspec-mocks .
docker build --build-arg ruby=okyada/ruby:1.8.7 -t rspec-mocks .
```

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -302,20 +334,21 @@ def receiver.method_missing(*); end # a poor man's stub...
...
+          it "allows the caller to configure how the subject responds" do
+            object.freeze
+            target.to receive(:foo).and_return(5)
+            expect(object.foo).to eq(5)
+            expect { reset object }.to change { object.foo }.from(5).to(:bar)
+          end
         end
       end
 
+
```
JonRowe, line 351: Remove excess whitespace.

lib/rspec/mocks/proxy.rb
```diff
@@ -35,8 +35,7 @@ def initialize(object, order_group, options={})
 
       # @private
       def ensure_can_be_proxied!(object)
-        return unless object.is_a?(Symbol) || object.frozen?
-        return if object.nil?
+        return unless object.is_a?(Symbol) || (object.is_a?(String) && object.frozen?)
```
JonRowe, line 38: This check is wrong, we still need to check other types of objects are frozen. If you want to improve this check to actually check the real object is frozen please do, (prehaps checking that the frozen method is kernels method) but as it is I don't want to only warn for frozen strings.

JonRowe (changes requested):

lib/rspec/mocks/proxy.rb
keegangroth: It's actually simply redundant, the new exception handling takes care of all objects regardless of this check. I had mentioned this previously in https://github.com/rspec/rspec-mocks/pull/1401#discussion_r810679191 and @pirj advised minimizing the changes, but I would be happy to remove this method entirely to avoid this confusion.

cknoxrun:
This one is biting me too with Rails - happy to help get it across the finish line if I can.

pirj:
@cknoxrun You're welcome to.

keegangroth:
I'm still watching this, was just waiting @JonRowe to say how he wanted it, but I guess I'll just take another shot

lib/rspec/mocks/proxy.rb
JonRowe: This still need to restore the original symbol check.

spec/rspec/mocks/syntax_agnostic_message_matchers_spec.rb
```diff
@@ -97,12 +91,6 @@ module Mocks
 
         verify subject
       end
-
-      context 'when target cannot be proxied' do
-        it 'raises ArgumentError with message' do
-          expect { ::RSpec::Mocks.expect_message(:subject, :foo) { :a } }.to raise_error(ArgumentError)
-        end
-      end
```
JonRowe, line 105: Restore this check.

spec/rspec/mocks/syntax_agnostic_message_matchers_spec.rb
```diff
@@ -42,12 +42,6 @@ module Mocks
 
         reset subject
       end
-
-      context 'when target cannot be proxied' do
-        it 'raises ArgumentError with message' do
-          expect { ::RSpec::Mocks.allow_message(:subject, :foo) { :a } }.to raise_error(ArgumentError)
-        end
-      end
```
JonRowe, line 50: Restore this check.

spec/rspec/mocks/space_spec.rb
```diff
@@ -232,22 +232,6 @@ def create_generations
...
-    end
-
-    it 'raises ArgumentError with message if object is frozen' do
-      space1 = Space.new
-      object = "subject".freeze
-      expected_message = "Cannot proxy frozen objects, rspec-mocks relies on proxies for method stubbing and expectations."
-
-      expect { space1.proxy_for(object) }.to raise_error(ArgumentError, expected_message)
-    end
-
```
JonRowe, line 250: Restore these checks.

JonRowe (changes requested):
@keegangroth Apologies, I thought I made it clear that you still needed to restore some existing behaviour for this to be mergable, as that hadn't happened I didn't re-review. I've highlighted the issues again.

lib/rspec/mocks/proxy.rb
keegangroth: Ok, so just so I'm clear. You want to maintain the check for Symbol in this precise location, even though it's covered by the exception handling in method_double? Are you ok with not checking `#frozen?` for any classes here? I'm confused because in https://github.com/rspec/rspec-mocks/pull/1401#discussion_r834210635 you said I was missing handling for non-Symbol/String frozen objects, which is the entire point of this PR.

keegangroth: OK, got it now.

simonxmh:
Any updates on this? We are still encountering this as a blocker

JonRowe:
Merged in [727ba92](https://github.com/rspec/rspec-mocks/commit/727ba922042df2da99933b44542952135dce37f0)

Issue #1396, opened 2021-01-07
Keyword argument error on Ruby 3.0 for #receive


swiknaba:
### Subject of the issue
When passing a block to`#receive`, it does not yield kwargs correctly but instead wraps kwargs in an array.

### Your environment
* Ruby version: 3.0.0
* rspec-mocks version: 3.10.1

### Steps to reproduce
```ruby
# Gemfile

source 'https://rubygems.org'
gem 'rspec'
ruby '3.0.0'
```

run `rspec --init`

```ruby
# spec/receive_spec.rb

class Maker
  def do_stuff(thing: nil)
    puts "thing from kwargs: #{thing}"
  end
end

class Caller
  def self.prepare(_arg, maker: nil)
    clerk = new(maker: maker)
    yield(clerk)
  end

  def initialize(maker: nil)
    @maker = maker
  end

  def call_maker(thing: nil)
    @maker.do_stuff(thing: thing)
  end
end

require_relative 'spec_helper' # just the boilerplate from rspec --init

describe Caller do
  let(:the_thing) { 'test' }
  let(:maker) { Maker.new }

  # fails in Ruby 3.0.0
  it 'hands the thing to the maker (kwargs)' do
    expect(maker).to receive(:do_stuff) do |**kwargs|
      expect(kwargs[:thing]).to eq(the_thing)
    end.exactly(1).times

    described_class.prepare('bar', maker: maker) do |clerk|
      clerk.call_maker(thing: the_thing)
    end
  end

  it 'hands the thing to the maker (args)' do
    expect(maker).to receive(:do_stuff) do |*args|
      expect(args.first[:thing]).to eq(the_thing)
    end.exactly(1).times

    described_class.prepare('bar', maker: maker) do |clerk|
      clerk.call_maker(thing: the_thing)
    end
  end
end

```

-> run `rspec`.
Ruby 2.7.2: all specs are passing ✅ 
Ruby 3.0.0: the first spec is failing ❌ 

```
  1) Caller hands the thing to the maker (kwargs)
     Failure/Error: expect(kwargs[:thing]).to eq(the_thing)
     
       expected: "test"
            got: nil
     
       (compared using ==)
     # ./spec/receive_spec.rb.rb:30:in `block (3 levels) in <top (required)>'
     # ./spec/receive_spec.rb.rb:18:in `call_maker'
     # ./spec/receive_spec.rb.rb:34:in `block (3 levels) in <top (required)>'
[...]

PR #1407, opened 2021-01-23
Add tests and fix for ruby 3 keyword arguments issues

jhottenstein:
I recently ran into a similar issue to #1396 when upgrading our codebase to ruby 3.

This PR adds tests for both scenarios (`receive` with a block with kwargs and `and_call_original` when mocking a method that uses keyword arguments).  The tests may be redundant since fixing recorder fixed both tests.

I noticed other spots where define_method is used to intercept method calls (e.g. https://github.com/rspec/rspec-mocks/blob/0a52e0a86b126b4bab94d277b2ad99a7492dc37d/lib/rspec/mocks/any_instance/chain.rb#L27 and  https://github.com/rspec/rspec-mocks/blob/0a52e0a86b126b4bab94d277b2ad99a7492dc37d/lib/rspec/mocks/matchers/receive.rb#L61) so this PR may not be sufficient to fix all keyword argument issues

### DIFF
--- a/spec/rspec/mocks/matchers/receive_spec.rb
+++ b/spec/rspec/mocks/matchers/receive_spec.rb
@@ -62,12 +62,44 @@ module RSpec
           expect(receiver.foo).to eq(4)
         end
 
-        it 'allows a `do...end` block implementation with keyword args to be provided' do
-          wrapped.to receive(:foo) do |**kwargs|
-            kwargs[:kw]
+        if RSpec::Support::RubyFeatures.kw_args_supported?
+          binding.eval(<<-RUBY, __FILE__, __LINE__)
+          it 'allows a `do...end` block implementation with keyword args to be provided' do
+            wrapped.to receive(:foo) do |**kwargs|
+              kwargs[:kw]
+            end
+
+            expect(receiver.foo(kw: :arg)).to eq(:arg)
           end
 
-          expect(receiver.foo(kw: :arg)).to eq(:arg)
+          it 'allows a `do...end` block implementation with optional keyword args to be provided' do
+            wrapped.to receive(:foo) do |kw: :arg|
+              kw
+            end
+
+            expect(receiver.foo(kw: 1)).to eq(1)
+          end
+
+          it 'allows a `do...end` block implementation with optional keyword args to be provided' do
+            wrapped.to receive(:foo) do |kw: :arg|
+              kw
+            end
+
+            expect(receiver.foo).to eq(:arg)
+          end
+          RUBY
+        end
+
+        if RSpec::Support::RubyFeatures.required_kw_args_supported?
+          binding.eval(<<-RUBY, __FILE__, __LINE__)
+          it 'allows a `do...end` block implementation with required keyword args' do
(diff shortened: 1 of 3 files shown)


JonRowe:
Thanks for tackling this, you need to wrap the specs in eval blocks so that they can safely be ignored by legacy Rubies. Theres examples around of this in #1385

jhottenstein:
Thanks for the review and the pointer.  I was unsure how rspec handled keyword args and legacy rubies.  

Do you want me to add more specs for keyword args in the other spots that I highlighted?  Our codebase didn't run into issues with those spots but I'm happy to attempt a reproduction or add more keyword argument tests in general.

JonRowe:
> Do you want me to add more specs for keyword args in the other spots that I highlighted? Our codebase didn't run into issues with those spots but I'm happy to attempt a reproduction or add more keyword argument tests in general.

More keyword specs in examples are welcome, we are lacking that aspect of coverage, although I'd prefer a seperate PR for general keyword examples, and focus this on your fix 👍

JonRowe:
This looks great apart from the JRuby failures, can you investigate?

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -62,6 +62,46 @@ module Mocks
...
+            end
+
+            expect(receiver.foo).to eq(:arg)
+          end
+          RUBY
+        end
+
+        if RSpec::Support::RubyFeatures.required_kw_args_supported?
+          binding.eval(<<-RUBY, __FILE__, __LINE__)
+          xit 'allows a `do...end` block implementation with required keyword args' do
```
jhottenstein, line 95: This test fails jruby 9.1 & 9.2 in a way that seems like an actual issue.  

I'm happy to remove it as I'm not even sure that anyone would use a mock method in this way.

I don't have jruby locally but I could try to get it set up to debug further if desired.

headius:
Was this fix really for JRuby? If we are differing on some language-level behavior compared to an equivalent CRuby, we definitely would like to see a bug report for that.

jhottenstein:
@JonRowe It was a bad test setup.  `wrapped` can be expect/allow or expect_any_instance_of/allow_any_instance_of.  A block implementation cannot be shared between those cases when `yield_receiver_to_any_instance_implementation_blocks` is true.

I set it to false for the test that pass block implementations.  I could move the test into the individual blocks (allow(...).to receive/expect(...).to receive/allow_any_instance_of.to receive/expect_any_instance_of.to receive

jhottenstein:
@headius, I can write something up.  Let me create a simpler repro.

JonRowe (approved):

JonRowe:
Thanks 🙏

dorner:
@JonRowe can we push a release for this? Currently upgrading to Ruby 3 breaks a whole lot of specs.

JonRowe:
A release is planned but I have to fix some CI issues first.

JonRowe:
Released as v3.10.2

dorner:
Awesome, thanks!

PR #1409, opened 2021-02-01
Verify partial doubles by default

pirj:
Depends on:
 - https://github.com/rspec/rspec-support/pull/494

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,8 @@ Breaking Changes:
 * Extract monkey-patching `should_receive`/`stub` syntax. (Phil Pirozhkov, #1365)
 * Remove the deprecated `RSpec::Mocks::CannotSupportArgMutationsError` constant.
   (Phil Pirozhkov, #1400)
+* Change the default setting for `RSpec::Mocks::Configuration#verify_partial_doubles`
+  to `true`. (Phil Pirozhkov, #1409)
 
 Bug Fixes:
 
--- a/features/verifying_doubles/partial_doubles.feature
+++ b/features/verifying_doubles/partial_doubles.feature
@@ -1,9 +1,8 @@
 Feature: Partial doubles
 
-  When the `verify_partial_doubles` configuration option is set, the same argument and
-  method existence checks that are performed for [`object_double`](./using-an-object-double) are also performed on
-  [partial doubles](../basics/partial-test-doubles). You should set this unless you have a good reason not to. It defaults to off
-  only for backwards compatibility.
+  The same argument and method existence checks that are performed for [`object_double`](./using-an-object-double) are also performed on
+  [partial doubles](../basics/partial-test-doubles). You should only turn off `verify_partial_doubles` (by setting it to false)
+  if you have a really good reason to.
 
   Scenario: doubling an existing object
     Given a file named "spec/user_spec.rb" with:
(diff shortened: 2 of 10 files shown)


spec/spec_helper.rb
```diff
@@ -77,7 +77,10 @@ def self.fake_matcher_description
 require 'rspec/support/spec'
 
 RSpec.configure do |config|
-  config.mock_with :rspec
+  config.mock_with :rspec do |mocks|
+    mocks.verify_partial_doubles = false
```
pirj, line 81: Sadly enough, half of our own specs fail if this is on.

features/verifying_doubles/partial_doubles.feature
```diff
@@ -1,9 +1,7 @@
 Feature: Partial doubles
 
-  When the `verify_partial_doubles` configuration option is set, the same argument and
-  method existence checks that are performed for [`object_double`](./using-an-object-double) are also performed on
-  [partial doubles](../basics/partial-test-doubles). You should set this unless you have a good reason not to. It defaults to off
-  only for backwards compatibility.
+  The same argument and method existence checks that are performed for [`object_double`](./using-an-object-double) are also performed on
+  [partial doubles](../basics/partial-test-doubles). Do not unset this `verify_partial_doubles` unless you have a really good reason to.
```
JonRowe, line 4:
```suggestion
  [partial doubles](../basics/partial-test-doubles). You should only turn off `verify_partial_doubles` (by setting it to false)
  if you have a really good reason to.
```

spec/spec_helper.rb
JonRowe: Do you intend to change that? You seem to have started above?

features/verifying_doubles/partial_doubles.feature
```diff
@@ -16,12 +14,6 @@ Feature: Partial doubles
         "saved!" if user.save
       end
 
-      RSpec.configure do |config|
-        config.mock_with :rspec do |mocks|
-          mocks.verify_partial_doubles = true
-        end
-      end
```
JonRowe, line 23: Do we want an example with this turned off?

Changelog.md
```diff
@@ -7,6 +7,7 @@ Breaking Changes:
 * Extract monkey-patching `should_receive`/`stub` syntax. (Phil Pirozhkov, #1365)
 * Remove the deprecated `RSpec::Mocks::CannotSupportArgMutationsError` constant.
   (Phil Pirozhkov, #1400)
+* Default `verify_partial_doubles` to `true`. (Phil Pirozhkov, #1409)
```
JonRowe, line 10:
```suggestion
* Change the default setting for `RSpec::Mocks::Configuration#verify_partial_doubles`
  to `true`. (Phil Pirozhkov, #1409)
```

spec/spec_helper.rb
pirj: I've only fixed those examples that were also failing with `mocks.verify_partial_doubles = false` in `spec_helper` due to the use of isolated configuration using the default `true`.

Making all our `expect(Object.new).to receive(:foo)` verifiable is a huge task.

features/verifying_doubles/partial_doubles.feature
pirj: Added 👍

JonRowe (approved):

PR #1411, opened 2021-02-12
Add `and_invoke` for sequential mixed (return/raise) responses.

askreet:
In tests I occasionally encounter things like this:

```ruby
called = false
allow(mock_api).to receive(:put_something) do
  # raise first call, second call succeeds
  unless called
    called = true
    raise ApiError, "Some Failure!"
  end
  :some_useful_value
end
```

It's commonly suggested to do similar things on [StackOverflow](https://stackoverflow.com/questions/14206477/rspec-retry-throw-exception-and-then-return-value), for example. It seemed a common enough problem to try to generalize a solution that is a bit more readable. This patch allows the following alternative:

```ruby
allow(mock_api).to receive(:put_something).and_invoke(-> { raise ApiError, "Some Failure!" },
                                                      -> { :some_useful_value })
```

`and_invoke` behaves like `and_return` with regards to sequential calls, matching expected call counts, and repeating the last call indefinitely in cases where `allow` is used vs. `expect` with `exactly(n).times`.

I alternatively considered some more magical approaches like:

```ruby
allow(mock_api).to receive(:put_something).and_return_or_raise(ApiError.new("Some Failure!"), true)
```

```ruby
allow(mock_api).to receive(:put_something).chain_response
                                          .and_raise(RuntimeError, "Some Failure!")
                                          .and_return(true)
```

Ultimately, I thought Procs were more "native", if slightly ugly.

### DIFF
--- a/features/configuring_responses/mixed_responses.feature
+++ b/features/configuring_responses/mixed_responses.feature
@@ -10,7 +10,7 @@ Feature: Mixed responses
       RSpec.describe "when the method is called multiple times" do
         it "raises and then later returns a value" do
           dbl = double
-          allow(dbl).to receive(:foo).and_invoke(-> { raise "failure" }, -> { true })
+          allow(dbl).to receive(:foo).and_invoke(lambda { raise "failure" }, lambda { true })
 
           expect { dbl.foo }.to raise_error("failure")
           expect(dbl.foo).to eq(true)
--- a/features/configuring_responses/mixed_responses.feature
+++ b/features/configuring_responses/mixed_responses.feature
@@ -1,0 +1,21 @@
+Feature: Mixed responses
+
+  Use `and_invoke` to invoke a Proc when a message is received. Pass `and_invoke` multiple
+  Procs to have different behavior for consecutive calls. The final Proc will continue to be
+  called if the message is received additional times.
+
+  Scenario: Mixed responses
+    Given a file named "raises_and_then_returns.rb" with:
+      """ruby
+      RSpec.describe "when the method is called multiple times" do
+        it "raises and then later returns a value" do
+          dbl = double
+          allow(dbl).to receive(:foo).and_invoke(-> { raise "failure" }, -> { true })
+
+          expect { dbl.foo }.to raise_error("failure")
+          expect(dbl.foo).to eq(true)
+        end
+      end
+      """
+     When I run `rspec raises_and_then_returns.rb`
+     Then the examples should all pass
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -732,12 +732,14 @@ module RSpec
         @procs_to_invoke = procs_to_invoke
       end
 
-      def call(*_args_to_ignore, &_block)
-        if @procs_to_invoke.size > 1
-          @procs_to_invoke.shift.call
-        else
-          @procs_to_invoke.first.call
-        end
+      def call(*args, &block)
+        proc = if @procs_to_invoke.size > 1
+                 @procs_to_invoke.shift
+               else
+                 @procs_to_invoke.first
+               end
+
+        proc.call(*args, &block)
       end
     end
 
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -117,6 +117,10 @@ module RSpec
         end
 
         procs.unshift(first_proc)
+        if procs.any? { |p| !p.respond_to?(:call) }
+          raise ArgumentError, "Arguments to `and_invoke` must be callable."
+        end
+
         @expected_received_count = [@expected_received_count, procs.size].max unless ignoring_args? || (@expected_received_count == 0 && @at_least)
         self.terminal_implementation_action = AndInvokeImplementation.new(procs)
 
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -53,7 +53,7 @@ module RSpec
       # etc.
       #
       # If the message is received more times than there are values, the last
-      # value is received for every subsequent call.
+      # value is returned for every subsequent call.
       #
       # @return [nil] No further chaining is supported after this.
       # @example
@@ -81,6 +81,44 @@ module RSpec
         values.unshift(first_value)
         @expected_received_count = [@expected_received_count, values.size].max unless ignoring_args? || (@expected_received_count == 0 && @at_least)
         self.terminal_implementation_action = AndReturnImplementation.new(values)
+
+        nil
+      end
+
+      # Tells the object to invoke a Proc when it receives the message. Given
+      # more than one value, the result of the first Proc is returned the first
+      # time the message is received, the result of the second Proc is returned
+      # the next time, etc, etc.
+      #
+      # If the message is received more times than there are Procs, the result of
+      # the last Proc is returned for every subsequent call.
+      #
+      # @return [nil] No further chaining is supported after this.
+      # @example
+      #   allow(api).to receive(:get_foo).and_return(-> { raise ApiTimeout })
+      #   api.get_foo # => raises ApiTimeout
+      #   api.get_foo # => raises ApiTimeout
+      #
+      #   allow(api).to receive(:get_foo).and_return(-> { raise ApiTimeout }, -> { raise ApiTimeout }, -> { :a_foo })
+      #   api.get_foo # => raises ApiTimeout
+      #   api.get_foo # => rasies ApiTimeout
+      #   api.get_foo # => :a_foo
+      #   api.get_foo # => :a_foo
+      #   api.get_foo # => :a_foo
+      #   # etc
+      def and_invoke(first_proc, *procs)
+        raise_already_invoked_error_if_necessary(__method__)
+        if negative?
+          raise "`and_invoke` is not supported with negative message expectations"
+        end
+
+        if block_given?
+          raise ArgumentError, "Implementation blocks aren't supported with `and_invoke`"
+        end
+
+        procs.unshift(first_proc)
+        @expected_received_count = [@expected_received_count, procs.size].max unless ignoring_args? || (@expected_received_count == 0 && @at_least)
+        self.terminal_implementation_action = AndInvokeImplementation.new(procs)
 
         nil
       end
@@ -679,6 +717,22 @@ module RSpec
           @values_to_return.shift
         else
           @values_to_return.first
+        end
+      end
+    end
+
+    # Handles the implementation of an `and_invoke` implementation.
+    # @private
+    class AndInvokeImplementation
+      def initialize(procs_to_invoke)
+        @procs_to_invoke = procs_to_invoke
+      end
+
+      def call(*_args_to_ignore, &_block)
+        if @procs_to_invoke.size > 1
+          @procs_to_invoke.shift.call
+        else
+          @procs_to_invoke.first.call
         end
       end
     end
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -95,11 +95,11 @@ module RSpec
       #
       # @return [nil] No further chaining is supported after this.
       # @example
-      #   allow(api).to receive(:get_foo).and_return(-> { raise ApiTimeout })
+      #   allow(api).to receive(:get_foo).and_invoke(-> { raise ApiTimeout })
       #   api.get_foo # => raises ApiTimeout
       #   api.get_foo # => raises ApiTimeout
       #
-      #   allow(api).to receive(:get_foo).and_return(-> { raise ApiTimeout }, -> { raise ApiTimeout }, -> { :a_foo })
+      #   allow(api).to receive(:get_foo).and_invoke(-> { raise ApiTimeout }, -> { raise ApiTimeout }, -> { :a_foo })
       #   api.get_foo # => raises ApiTimeout
       #   api.get_foo # => rasies ApiTimeout
       #   api.get_foo # => :a_foo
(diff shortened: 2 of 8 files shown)


askreet:
Ah, did not realize `rspec-mocks` kept compatibility with Ruby 1.8.7 -- impressive! Will fix the syntax to be compatible. (I assume this is just a "stabby lambda" issue.)

pirj:
How do you feel about accumulating configured responses?
Now, e.g. `and_return` accepts several arguments, but the following:
```ruby
allow(foo)
  .to receive(:bar)
  .and_return(1)
  .and_return(2)
```
won't work.

In the same way, it could be possible to use `and_raise`.

It's very similar to what you suggest with `chain_response`.

Side note: it's kind of confusing that it is possible to configure several responses at the moment, but only one, the last defined, actually comes into play.

askreet:
I'm surprised this works, too, @pirj. I'm a bit worried about changing it because in all liklihood there are people who have done something like:

```ruby
def default_fixture
  instance_double(Foo).tap |dbl|
    expect(dbl).to receive(:color).and_return(:blue)
  end
end

it "bar" do
  expect(baz(default_fixture)).to eq(4)
end

it "bars slightly differently under certain circumstances" do
  fixture = default_fixture.and_return(:aqua)

  expect(baz(fixture)).to eq(5)
end
```

However, your example raises a neat idea that would require a bit more careful architecture change:

```ruby
allow(foo)
  .to receive(:bar)
  .and_return(1) # <== sets terminal action to AndReturnImplementation.new([1])
  .and_then_return(2) # <== replaces terminal action with ChainedTerminalAction.new([ existing_action, AndReturnImplementation([2]) ])
  .and_then_raise(RuntimeError, "failure!") # <== calls ChainedTerminalAction.append(AndRaiseImplementation.new(...))
```

The trick here would be the earlier chained implementations need to be extends to support _not_ repeating the last element in their list, or track if they have been 'consumed' or not so that the chained action can move onto the next item. Likewise, if we want to keep existing behavior, we'd have to repeat the last action of the last chain element. Alternatively we could say that once you opt into this chaining construct, you must build an explicit chain of response elements.

Overall, I quite like this approach, but I'll hold off on making the change until I hear from more maintainers about the direction.

Note that my original `#chained_responses` implementation is actually easier to pull off because it can set the terminal implementation to a new ChainedResponses instance and then return that, allowing all further calls to go to the new object (although that would mean you couldn't use more RSpec DSL at that point).

pirj:
I believe that chaining configured response qualifiers would imply those responses being called in order.

It needs a second thought on how countable qualifiers should work though.
```ruby
receive(:foo)
  .and_return(1).twice
  .and_return(2).once # return 1, 1, then 2 
```
vs

```ruby
receive(:foo)
  .twice
  .and_return(1)
  .and_return(2) # ok
```

```ruby
receive(:foo)
  .thrice
  .and_return(1)
  .and_return(2) # the number of configured responses doesn't add up. Should we return `nil` on the third call?
```

```ruby
receive(:foo)
  .twice
  .and_return(1)
  .and_return(2)
  .and_return(3) # Error! The number of configured responses exceeds the expected number of calls
```

Semi-related https://github.com/rspec/rspec-expectations/issues/1252

askreet:
I don't know, are you saying that in the case where we've qualified the counts with _expectations_ that the responses should switch from being overridden to appending in a chained fashion? The inconsistency with the behavior when you call `and_return` twice (override) is pretty surprising. The nice thing about introducing a new method (`and_then...` or `chained_responses`) is that it's clear you're altering the default behavior.

Even in qualified cases where expectation counts have been specified, people may still be relying on the override behavior (like my example from above). It seems backwards compatibility is a reasonable top priority for this gem.

lib/rspec/mocks/message_expectation.rb
```diff
@@ -85,6 +85,44 @@ def and_return(first_value, *values)
...
+      # more than one value, the result of the first Proc is returned the first
+      # time the message is received, the result of the second Proc is returned
+      # the next time, etc, etc.
+      #
+      # If the message is received more times than there are Procs, the result of
+      # the last Proc is returned for every subsequent call.
+      #
+      # @return [nil] No further chaining is supported after this.
+      # @example
+      #   allow(api).to receive(:get_foo).and_return(-> { raise ApiTimeout })
```
JonRowe, line 98:
```suggestion
      #   allow(api).to receive(:get_foo).and_invoke(-> { raise ApiTimeout })
```

lib/rspec/mocks/message_expectation.rb
```diff
@@ -85,6 +85,44 @@ def and_return(first_value, *values)
...
+      #   api.get_foo # => raises ApiTimeout
+      #   api.get_foo # => raises ApiTimeout
+      #
+      #   allow(api).to receive(:get_foo).and_return(-> { raise ApiTimeout }, -> { raise ApiTimeout }, -> { :a_foo })
```
JonRowe, line 102:
```suggestion
      #   allow(api).to receive(:get_foo).and_invoke(-> { raise ApiTimeout }, -> { raise ApiTimeout }, -> { :a_foo })
```

lib/rspec/mocks/message_expectation.rb
```diff
@@ -683,6 +721,22 @@ def call(*_args_to_ignore, &_block)
...
+      def initialize(procs_to_invoke)
+        @procs_to_invoke = procs_to_invoke
+      end
+
+      def call(*_args_to_ignore, &_block)
+        if @procs_to_invoke.size > 1
+          @procs_to_invoke.shift.call
+        else
+          @procs_to_invoke.first.call
+        end
```
JonRowe, line 736: This has a bug if there are nils or an empty argument splat is used.

lib/rspec/mocks/message_expectation.rb
```diff
@@ -683,6 +721,22 @@ def call(*_args_to_ignore, &_block)
...
     end
 
+    # Handles the implementation of an `and_invoke` implementation.
+    # @private
+    class AndInvokeImplementation
+      def initialize(procs_to_invoke)
+        @procs_to_invoke = procs_to_invoke
+      end
+
+      def call(*_args_to_ignore, &_block)
```
JonRowe, line 731: The arguments should probably be passed to the procs similar to how blocks are called...

spec/rspec/mocks/and_invoke_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+        it 'raises ArgumentError' do
+          expect {
+            allow(obj).to receive(:foo).and_invoke('bar') { 'baz' }
+          }.to raise_error(ArgumentError, /implementation block/i)
+        end
+      end
+
+      context 'when no argument is passed' do
+        it 'raises ArgumentError' do
+          expect { allow(obj).to receive(:foo).and_return }.to raise_error(ArgumentError)
```
JonRowe, line 16:
```suggestion
          expect { allow(obj).to receive(:foo).and_invoke }.to raise_error(ArgumentError)
```

JonRowe:
Initially I was going to suggest that as `and_return` already allows multiple static responses[1], it could be re-written to allow this scenario too, as having another name for configuring return responses seems sub-ideal, but I think I'm convinced this specific scenario needs its own name, to distinguish between a list of callables and others.

[1](https://relishapp.com/rspec/rspec-mocks/v/3-10/docs/configuring-responses/returning-a-value#specify-different-return-values-for-multiple-calls)

lib/rspec/mocks/message_expectation.rb
askreet: The original intent was to chain returns and rescues, but I don't see why this can't share the same block semantics, will fix.

lib/rspec/mocks/message_expectation.rb
askreet: Good catch -- I'll validate the inputs immediately and raise some error (will try to find prior art).

lib/rspec/mocks/message_expectation.rb
JonRowe: Chain returns is easy with `and_return(1,2,3,4)` etc, so this is a custom implementation so should respect the same rules, e.g. `and_return { |args| expect(args).to eq [] }` is valid and used to validate arguments, so its surprising if this doesn't also have access to args, as its essentially the "chain" form of `and_return { ... }`

askreet: Agreed, implemented in latest commit. :+1:

lib/rspec/mocks/message_expectation.rb
askreet: Should be fixed in the last couple commits. :+1:

benoittgt:
Thanks for this proposal. I love the idea. Last version of code looks good to me. Thanks for the clarity of the PR.

askreet:
@benoittgt Great, glad to hear it. I'm good with where this is at whenever we have quorum to merge. Thanks!

features/configuring_responses/mixed_responses.feature
```diff
@@ -0,0 +1,21 @@
...
+  Scenario: Mixed responses
+    Given a file named "raises_and_then_returns.rb" with:
+      """ruby
+      RSpec.describe "when the method is called multiple times" do
+        it "raises and then later returns a value" do
+          dbl = double
+          allow(dbl).to receive(:foo).and_invoke(lambda { raise "failure" }, lambda { true })
+
+          expect { dbl.foo }.to raise_error("failure")
+          expect(dbl.foo).to eq(true)
```
pirj, line 16: The feature description says "if the message is received additional times", maybe add an additional line repeating this one to prove/indicate that?

lib/rspec/mocks/message_expectation.rb
```diff
@@ -85,6 +85,48 @@ def and_return(first_value, *values)
...
+        raise_already_invoked_error_if_necessary(__method__)
+        if negative?
+          raise "`and_invoke` is not supported with negative message expectations"
+        end
+
+        if block_given?
+          raise ArgumentError, "Implementation blocks aren't supported with `and_invoke`"
+        end
+
+        procs.unshift(first_proc)
```
pirj, line 119: 👍 I clearly understand that you enforce passing at least a single argument and rely on Ruby.

The mutating `unshift` slightly concerned me, in the following scenario:
```ruby
procs = [-> { 1 }, -> { raise }]
allow(foo).to receive(:bar).and_invoke(*procs)
allow(foo).to receive(:baz).and_invoke(*procs)
```
but on a synthetic test, it seems that it has no effect on this user-defined `procs`, so all good.

lib/rspec/mocks/message_expectation.rb
```diff
@@ -683,6 +725,24 @@ def call(*_args_to_ignore, &_block)
...
 
+    # Handles the implementation of an `and_invoke` implementation.
+    # @private
+    class AndInvokeImplementation
+      def initialize(procs_to_invoke)
+        @procs_to_invoke = procs_to_invoke
+      end
+
+      def call(*args, &block)
+        proc = if @procs_to_invoke.size > 1
```
pirj, line 736: We could check for `@procs_to_invoke.one?`, it's [there since 1.8.7](https://ruby-doc.org/core-1.8.7/Enumerable.html#method-i-one-3F).

lib/rspec/mocks/message_expectation.rb
```diff
@@ -85,6 +85,48 @@ def and_return(first_value, *values)
...
       end
 
+      # Tells the object to invoke a Proc when it receives the message. Given
+      # more than one value, the result of the first Proc is returned the first
+      # time the message is received, the result of the second Proc is returned
+      # the next time, etc, etc.
+      #
+      # If the message is received more times than there are Procs, the result of
+      # the last Proc is returned for every subsequent call.
+      #
```
pirj, line 95: We seem to pass message args to that proc, does it make sense to reflect this in the docs?

JonRowe:
Thanks!

spec/rspec/mocks/and_invoke_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          expect { allow(obj).to receive(:foo).and_invoke }.to raise_error(ArgumentError)
+        end
+      end
+
+      context 'when a non-callable are passed in any position' do
+        let(:non_callable) { nil }
+        let(:callable) { lambda { nil } }
+
+        it 'raises ArgumentError' do
+          error = [ArgumentError, "Arguments to `and_invoke` must be callable."]
```
pirj, line 25: Personally, I would prefer not to DRY this.

spec/rspec/mocks/and_invoke_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          expect { allow(obj).to receive(:foo).and_invoke(non_callable) }.to raise_error(*error)
+          expect { allow(obj).to receive(:foo).and_invoke(callable, non_callable) }.to raise_error(*error)
+        end
+      end
+
+      context 'when calling passed callables' do
+        let(:dbl) { double }
+
+        it 'passes the arguments into the callable' do
+          expect(dbl).to receive(:square_then_cube).and_invoke(lambda { |i| i ** 2 },
```
pirj, line 36: Probably `allow`, not `expect`?

spec/rspec/mocks/and_invoke_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+                                                               lambda { |i| i ** 3 })
+
+          expect(dbl.square_then_cube(2)).to eq 4
+          expect(dbl.square_then_cube(2)).to eq 8
+        end
+      end
```
pirj, line 42: Do you mind adding an example showing that the block is also passed to that proc for completeness?

pirj (approved):
Oh, I'm late for the party 🐌 

It looks amazing, thanks for the contribution!

lib/rspec/mocks/message_expectation.rb
JonRowe: Its not a directly passed in array, as its essentially the same as `and_invoke(-> { 1 }, -> { raise })` its also identical to our `and_return` implementation (I squashed a different quibble I had due to that).

PR #1415, opened 2021-02-19
Get rid of redefinition of `__send__`

nobu:
It may not be guaranteed in future versions.


JonRowe:
Migrated to the monorepo as rspec/rspec#115

Issue #1444, opened 2021-12-03
"singleton on non-persistent Java type" warning on JRuby when mocking an instance method on a Java object proxy


ikaronen-relex:
### Subject of the issue

As described at https://github.com/jruby/jruby/wiki/Persistence, JRuby issues a warning when accessing the singleton class of a Java object proxy unless the object's class has been explicitly marked as persistent. Such a warning is triggered by the code in RSpec::Mocks::InstanceMethodStasher#initialize when the `object` argument is a Java object proxy.

### Your environment
* Ruby version: JRuby 9.3.2.0
* rspec-mocks version: 3.10.0

### Steps to reproduce

Run the following spec under JRuby:

```ruby
describe RSpec::Mocks::InstanceMethodStasher do
  it 'does not warn when mocking a method on a Java proxy object' do
    list = java.util.ArrayList.new
    allow(list).to receive(:foo).and_return(42)
    expect(list.foo).to eq(42)
  end
end
```

### Expected behavior

The spec passes with no warnings printed to standard output or error streams.

### Actual behavior

The spec passes, but the following warning is printed to stderr:

```
<path_to_gems>/rspec-mocks-3.10.0/lib/rspec/mocks/instance_method_stasher.rb:8: warning: singleton on non-persistent Java type Java::JavaUtil::ArrayList (https://github.com/jruby/jruby/wiki/Persistence)
```

pirj:
> We recommend using other mechanisms to store per-Java-object data

We don't really store just yet, we're getting the singleton class:

```ruby
@klass = (class << object; self; end)
```

I recall there's a difference between this and `object.singleton_class`, like in certain cases one approach is forcing the singleton to be initialized, and the other does not, but it's hard to recall the details.

`#singleton_class` to my best memory was introduced in Ruby 1.9.x, and we still support 1.8.6 in RSpec 3.

I'd suggest finding discussions here around `singleton_class`, and see why we [decided to stick to this notation in RSpec 4](https://github.com/rspec/rspec-mocks/pull/1349).

Would the use of `@klass = obeject.singleton_class` fix the warning for you?

ikaronen-relex:
> Would the use of `@klass = obeject.singleton_class` fix the warning for you?

@pirj No, that gives the same warning. AIUI, simply accessing the singleton class (which causes JRuby to generate one) is enough to trigger the warning.

[...]

PR #1445, opened 2021-12-03
Fix persistence warning on JRuby when mocking a Java object wrapper

ikaronen-relex:
This is enough to fix https://github.com/rspec/rspec-mocks/issues/1444

### DIFF
--- a/lib/rspec/mocks/instance_method_stasher.rb
+++ b/lib/rspec/mocks/instance_method_stasher.rb
@@ -5,6 +5,7 @@ module RSpec
       def initialize(object, method)
         @object = object
         @method = method
+        object.class.__persistent__ = true if object.respond_to?(:java_class)  # https://github.com/jruby/jruby/wiki/Persistence
         @klass = (class << object; self; end)
 
         @original_method = nil


lib/rspec/mocks/instance_method_stasher.rb
```diff
@@ -5,6 +5,7 @@ class InstanceMethodStasher
       def initialize(object, method)
         @object = object
         @method = method
+        object.class.__persistent__ = true if object.respond_to?(:java_class)  # https://github.com/jruby/jruby/wiki/Persistence
```
pirj, line 8: I'd prefer to conditionally define `def initialize` depending on the platform.
Still not convinced that the warning is correctly triggered - we're not adding anything to the singleton class just yet. Sounds like it is worth bringing to the attention of JRuby team.

ikaronen-relex: @pirj Looks like my first attempt doesn't work quite right even on JRuby. ☹️ I'll see if I can fix it and then you can decide if you want to do some conditional method definition hacking for performance.

Anyway, FWIW, just getting the singleton class and storing it absolutely *does* require enabling the instance proxy cache, which is what JRuby is warning about. If it wasn't enabled at that point, what could happen is that the proxied Java object could be passed to some Java code that later returns it back to Ruby, at which point JRuby would normally construct a *new* instance proxy for it, and the new proxy wouldn't know that there's an old singleton class for an old proxy for the same Java object still floating around somewhere.

pirj: This is fine, doubles expect certain messages, and calling `respond_to?` on it may break expectations.

I'd suggest to check using `RSpec::Support::Ruby.jruby?`.

pirj:
Closing as the ticket this PR was purposed to fix has a workaround.

PR #1450, opened 2022-01-11
Allow Cucumber to run across Rubies.

petergoldstein:
Makes the version of Cucumber to be more free-floating
Detect the current version of Cucumber (specifically for old JRubies) and use an appropriate tag syntax

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -81,6 +81,10 @@ else
   gem 'json', '> 2.3.0'
 end
 
+if RUBY_VERSION < '2.0.0'
+  gem 'cucumber', "<= 1.3.22"
+end
+
 platforms :jruby do
   if RUBY_VERSION < '1.9.0'
     # Pin jruby-openssl on older J Ruby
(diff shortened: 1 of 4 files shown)


pirj:
There is a difference for 1.9.3, it seems to have stopped running cucumber. Didn't check other jobs.
https://github.com/rspec/rspec-mocks/runs/4779954634?check_suite_focus=true
https://github.com/rspec/rspec-mocks/runs/4772787816?check_suite_focus=true

petergoldstein:
@pirj I didn't see those because they didn't show failures.  I'll take a look, thanks.

petergoldstein:
@pirj Ok, that's fixed.  MRI Ruby 1.8.7, 1.9.2, and 1.9.3 were failing silently on bad tags.  I loosened the condition for tilde tags to include them, and now they are running cucumber and it's green.

In my view this is ready to be merged.

JonRowe:
Thanks!

PR #1458, opened 2022-01-30
Support more recent Rubygems and bundler for recent Rubies

petergoldstein:
This PR fixes the CI failure for ruby-head by allowing Ruby versions 3.1.x and later to use a more recent rubygems / bundler.


JonRowe:
Thanks, I've pulled this into our dev repo that manages these files.

PR #1461, opened 2022-02-18
Display keyword hashes in in expectation error messages

casperisfine:
Ref: https://github.com/vcr/vcr/pull/925
Ref: https://github.com/rspec/rspec-mocks/pull/1394

I spent quite a lot of time figuring this error:

```
  2) VCR.turned_on passes options through to .turn_off!
     Failure/Error: turn_off!(options)

       VCR received :turn_off! with unexpected arguments
         expected: ({:ignore_cassettes=>true})
              got: ({:ignore_cassettes=>true})
     # ./lib/vcr.rb:317:in `turned_on'
     # ./spec/lib/vcr_spec.rb:367:in `block (3 levels) in <top (required)>'
```

I quickly suspected it was a keyword argument issue, but it's far from
obvious to everyone, and even when you are familair with the issue
it doesn't tell you what was expected and what was received.

I doubt the way I implemented this is ok, but I think it's worth
opening the discussion

```
  2) VCR.turned_on passes options through to .turn_off!
     Failure/Error: turn_off!(options)

       VCR received :turn_off! with unexpected arguments
         expected: ({:ignore_cassettes=>true}) (keyword arguments)
              got: ({:ignore_cassettes=>true}) (options hash)
     # ./lib/vcr.rb:317:in `turned_on'
     # ./spec/lib/vcr_spec.rb:367:in `block (3 levels) in <top (required)>'
```

### DIFF
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -268,6 +268,17 @@ module RSpec
       def error_message(expectation, args_for_multiple_calls)
         expected_args = format_args(expectation.expected_args)
         actual_args = format_received_args(args_for_multiple_calls)
+
+        if RSpec::Support::RubyFeatures.distincts_kw_args_from_positional_hash? && expected_args == actual_args
+          expected_hash = expectation.expected_args.last
+          actual_hash = args_for_multiple_calls.last.last
+          if Hash === expected_hash && Hash === actual_hash &&
+            (Hash.ruby2_keywords_hash?(expected_hash) != Hash.ruby2_keywords_hash?(actual_hash))
+            actual_args += Hash.ruby2_keywords_hash?(actual_hash) ? " (keyword arguments)" : " (options hash)"
+            expected_args += Hash.ruby2_keywords_hash?(expected_hash) ? " (keyword arguments)" : " (options hash)"
+          end
+        end
+
         message = default_error_message(expectation, expected_args, actual_args)
 
         if args_for_multiple_calls.one?
--- a/spec/rspec/mocks/diffing_spec.rb
+++ b/spec/rspec/mocks/diffing_spec.rb
@@ -83,6 +83,39 @@ RSpec.describe "Diffs printed when arguments don't match" do
       end
     end
 
+    if RSpec::Support::RubyFeatures.distincts_kw_args_from_positional_hash?
+      eval <<-'RUBY', nil, __FILE__, __LINE__ + 1
+        it "print a diff when keyword argument were expected but got an option hash (using splat)" do
+          with_unfulfilled_double do |d|
+            expect(d).to receive(:foo).with(**expected_hash)
+            expect {
+              d.foo(expected_hash)
+            }.to fail_with(
+              "#<Double \"double\"> received :foo with unexpected arguments\n" \
+              "  expected: ({:baz=>:quz, :foo=>:bar}) (keyword arguments)\n" \
+              "       got: ({:baz=>:quz, :foo=>:bar}) (options hash)"
+            )
+          end
+        end
+      RUBY
+
+      eval <<-'RUBY', nil, __FILE__, __LINE__ + 1
+        it "print a diff when keyword argument were expected but got an option hash (literal)" do
+          with_unfulfilled_double do |d|
+            expect(d).to receive(:foo).with(:positional, keyword: 1)
+            expect {
+              options = { keyword: 1 }
+              d.foo(:positional, options)
+            }.to fail_with(
+              "#<Double \"double\"> received :foo with unexpected arguments\n" \
+              "  expected: (:positional, {:keyword=>1}) (keyword arguments)\n" \
+              "       got: (:positional, {:keyword=>1}) (options hash)"
+            )
+          end
+        end
+      RUBY
+    end
+
     if RUBY_VERSION.to_f < 1.9
       # Ruby 1.8 hashes are not ordered, but `#inspect` on a particular unchanged
       # hash instance should return consistent output. However, on Travis that does


pirj:
Do you think it's achievable with Ruby 2.x?

eregon:
In 2.7 yes, but before I believe it's impossible to know in the callee if kwargs or a positional Hash was passed.

lib/rspec/mocks/error_generator.rb
```diff
@@ -268,6 +268,18 @@ def unexpected_arguments_message(expected_args_string, actual_args_string)
       def error_message(expectation, args_for_multiple_calls)
         expected_args = format_args(expectation.expected_args)
         actual_args = format_received_args(args_for_multiple_calls)
+
+        if RSpec::Support::RubyFeatures.kw_args_supported? && expected_args == actual_args
+          expected_hash = expectation.expected_args.last
+          actual_hash = args_for_multiple_calls.last.last
+          if Hash === expected_hash && Hash === actual_hash
+            if Hash.ruby2_keywords_hash?(expected_hash) ^ Hash.ruby2_keywords_hash?(actual_hash)
+              actual_args += Hash.ruby2_keywords_hash?(actual_hash) ? " (keyword arguments)" : " (options hash)"
```
eregon, line 277: Note: `Hash.ruby2_keywords_hash?` will only recognize hashes from a  `ruby2_keywords` method/block, or a hash explcitly marked with `Hash.ruby2_keywords_hash`, but e.g., not `kw` in `def m(*args, **kw); kw; end`.
I guess `ruby2_keywords` is used somewhere before and so it just works here (is that in RSpec code?), but it's a bit subtle.

If `rspec-mocks` would store positional and kwargs arguments in separate fields (only on Ruby 3+, before can't separate them reliably) then I think it would be nicer and more reliable to check.

pirj:
In addition to what Benoit said, there are [certain problems](https://github.com/rspec/rspec-support/issues/522#issuecomment-991585928) with comparing method signatures, might be related.

lib/rspec/mocks/error_generator.rb
casperisfine: @eregon what you are saying is right, but I'm not trying to display whether a hash was keyword args 100% of the times, I only want to display it when two hashes are identical except for the `ruby2_keywords_hash?`.

#1394 add this code: 

```ruby
        if RUBY_VERSION >= "3"
          # if both arguments end with Hashes, and if one is a keyword hash and the other is not, they don't match
          if Hash === expected_args.last && Hash === actual_args.last
            if !Hash.ruby2_keywords_hash?(actual_args.last) && Hash.ruby2_keywords_hash?(expected_args.last)
              return false
            end
          end
        end
```

And it's this case I want to display. So yeah, my PR only apply to `3+` like the previous PR.

spec/rspec/mocks/diffing_spec.rb
```diff
@@ -83,6 +83,21 @@
       end
     end
 
+    if RUBY_VERSION >= "3" && RSpec::Support::RubyFeatures.kw_args_supported?
```
casperisfine, line 86: I added the `RUBY_VERSION >= "3"` which should solve the CI issues.

lib/rspec/mocks/error_generator.rb
```diff
@@ -268,6 +268,17 @@ def unexpected_arguments_message(expected_args_string, actual_args_string)
       def error_message(expectation, args_for_multiple_calls)
         expected_args = format_args(expectation.expected_args)
         actual_args = format_received_args(args_for_multiple_calls)
+
+        if RUBY_VERSION >= "3" && RSpec::Support::RubyFeatures.kw_args_supported? && expected_args == actual_args
+          expected_hash = expectation.expected_args.last
+          actual_hash = args_for_multiple_calls.last.last
+          if Hash === expected_hash && Hash === actual_hash &&
+            (Hash.ruby2_keywords_hash?(expected_hash) ^ Hash.ruby2_keywords_hash?(actual_hash))
```
pirj, line 276: Is `^` same as `!=`? I find `^` a bit cryptic.

casperisfine: Ah good point 😓

casperisfine: Fixed.

spec/rspec/mocks/diffing_spec.rb
pirj: We typically add such things to `rspec-support` and use a self-explanatory method on `RubyFeatures`. Do you think it's possible to come up with a name that would explain why a special case is needed?

casperisfine: Hum, `RubyFeatures.has_ruby2_compatibility_keywords`?

lib/rspec/mocks/error_generator.rb
```diff
@@ -268,6 +268,17 @@ def unexpected_arguments_message(expected_args_string, actual_args_string)
       def error_message(expectation, args_for_multiple_calls)
         expected_args = format_args(expectation.expected_args)
         actual_args = format_received_args(args_for_multiple_calls)
+
+        if RUBY_VERSION >= "3" && RSpec::Support::RubyFeatures.kw_args_supported? && expected_args == actual_args
```
pirj, line 272: Cosmetic: with `RUBY_VERSION >= "3"`, I believe `RSpec::Support::RubyFeatures.kw_args_supported?` will always evaluate to `true`, so it's redundant.

casperisfine: Well, `ruby2_keywords` might go away one day (not any time soon though).

spec/rspec/mocks/diffing_spec.rb
casperisfine: Sure. Should I make a PR to `rspec-support`?

lib/rspec/mocks/error_generator.rb
eregon: `Hash.ruby2_keywords_hash?` still can't tell if a Hash was used as keyword arguments, except for the case we are sure the keyword arguments were passed to a `ruby2_keywords` method, and that Hash then arrives it, not through another call (which would remove the flag IIRC).
Is that guaranteed here that `expected_hash` and `actual_hash` come from `ruby2_keywords` methods?

In other words, I think the current patch might not be able to show the helpful kwargs/non-kwargs suffix for things like:
```ruby
def foo(*args, **kwargs)
  ...
end

foo({a: 1})
vs
foo(a: 1)
```
but given this is rspec-mocks maybe it could because it might always control the definition of `foo` and always use `ruby2_keywords` there?

casperisfine: > Is that guaranteed here that expected_hash and actual_hash come from ruby2_keywords methods?

if `expected_hash == actual_hash` then AFAIK yes that's the only way. Otherwise we wouldn't be printing an error.

Of course there might be some weird edge cases if you monkey patch `Hash#==` but...

spec/rspec/mocks/diffing_spec.rb
pirj: Yes, please :+1:

lib/rspec/mocks/error_generator.rb
pirj: [`kw_args_supported?`](https://github.com/rspec/rspec-support/blob/623608b4f49c063fca2616cb3e35ae62ea5fe5cf/lib/rspec/support/ruby_features.rb#L130) has no awareness of that.
Currently, this code expands to:
```ruby
if RUBY_VERSION >= "3" && RUBY_VERSION >= '2.0.0'
```

lib/rspec/mocks/error_generator.rb
eregon: I see, there is similar logic in https://github.com/mame/rspec-mocks/blob/865ea3a795e0e56309b823e76e3f65eff5cfa448/lib/rspec/mocks/argument_list_matcher.rb#L63-L70
I think a `# See #args_match?` comment would be helpful.

Also the check doesn't necessarily seem correct, it's only in one direction, because passing kwargs to a method not taking kwargs is fine: https://github.com/mame/rspec-mocks/blob/865ea3a795e0e56309b823e76e3f65eff5cfa448/lib/rspec/mocks/argument_list_matcher.rb#L66
(the comment is unfortunately outdated/not matching the code)

I'd suggest to extract the check to a separate method, then it's clear it's the same condition.

spec/rspec/mocks/diffing_spec.rb
```diff
@@ -83,6 +83,21 @@
...
+      it "display keyword hashes" do
+        with_unfulfilled_double do |d|
+          expect(d).to receive(:foo).with(**expected_hash)
```
pirj, line 89:
```
spec/rspec/mocks/diffing_spec.rb:89: syntax error, unexpected tPOW, expecting ')'
            expect(d).to receive(:foo).with(**expected_hash)
```
is not something that is hard to work around, see an example https://github.com/rspec/rspec-support/blob/d04c8aef72d030fc354902f1e1b486b5c21e0732/spec/rspec/support/method_signature_verifier_spec.rb#L276

casperisfine: Oh wow, I didn't realize rspec was still compatible with 1.8. 

I'll fix this.

casperisfine: Done.

pirj (approved):
Nice! Thank you for the contribution.
Does it make sense to add an example the is the other way around (`receive(:foo).with(expected_hash)` and `d.foo(**expected_hash)`)?

pirj:
Another example probably worth adding/checking is reported in https://github.com/rspec/rspec-expectations/issues/1359, https://github.com/ruby-grape/grape/runs/5347064188?check_suite_focus=true:
```ruby
expect(subject).to receive(:namespace_inheritable).with(:version_options, using: :path)
...
options = options.reverse_merge(using: :path)
namespace_inheritable(:version_options, options)
```

casperisfine:
> Does it make sense to add an example the is the other way around

I don't think it's possible, because passing keyword arguments to a method that expects an option hash automatically fallback to an option hash. So unless I'm missing something I don't think it can happen.

casperisfine:
> Another example probably worth adding/checking is reported in

I added a second test case, the first one used splats `**`, the additional one use literal keyword arguments.

spec/rspec/mocks/diffing_spec.rb
```diff
@@ -83,6 +83,39 @@
...
+      eval <<-'RUBY', nil, __FILE__, __LINE__ + 1
+        it "print a diff when keyword argument were expected but got an option hash (using splat)" do
+          with_unfulfilled_double do |d|
+            expect(d).to receive(:foo).with(**expected_hash)
+            expect {
+              d.foo(expected_hash)
+            }.to fail_with(
+              "#<Double \"double\"> received :foo with unexpected arguments\n" \
+              "  expected: ({:baz=>:quz, :foo=>:bar}) (keyword arguments)\n" \
+              "       got: ({:baz=>:quz, :foo=>:bar}) (options hash)"
```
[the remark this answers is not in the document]
casperisfine: No as I said above, I don't think it's possible. If you know a way it could happen, let me know.

pirj:
I don't mind merging as is.
Just wondering if the following will be sufficient:

```diff
 if Hash === expected_hash && Hash === actual_hash &&
-  (Hash.ruby2_keywords_hash?(expected_hash) != Hash.ruby2_keywords_hash?(actual_hash))
+  Hash.ruby2_keywords_hash?(expected_hash) && !Hash.ruby2_keywords_hash?(actual_hash)
-  actual_args += Hash.ruby2_keywords_hash?(actual_hash) ? " (keyword arguments)" : " (options hash)"
+  actual_args += " (options hash)"
-  expected_args += Hash.ruby2_keywords_hash?(expected_hash) ? " (keyword arguments)" : " (options hash)"
+  expected_args += " (keyword arguments)"
 end
```
because it's enough to make the specs pass.

casperisfine:
> Just wondering if the following will be sufficient:

Assuming I'm right on the fact that it's not possible to pass keyword args when option hashes are expected then yes.

That being said I may be wrong, in which case making this change would mean that we don't handle it.

So it's up to you.

pirj:
@eregon I guess you're working on the bleeding edge and don't need a release to benefit from this improvement?

eregon:
For https://github.com/rspec/rspec-metagem/issues/68 indeed I don't need a release.
However, I think several people already ran into this (e.g., see links to this PR), so it would be very beneficial to get this in a release (avoids a significant amount of confusion).

BTW, for https://github.com/rspec/rspec-mocks/pull/1464 a release will be needed, I'll explain there.

pirj:
@casperisfine Thank you for the contribution!

PR #1464, opened 2022-03-21
Add missing ruby2_keywords calls in rspec-mocks

eregon:
* And standardize checks to `if respond_to?(:ruby2_keywords, true)`.

cc @bjfish

This is progress towards https://github.com/rspec/rspec-metagem/issues/68.
Before: 25 failures, After: 2 failures left (which seem unrelated):
```
Failures:

  1) #any_instance when stubbing aliased methods tracks aliased method calls
     Failure/Error: parent_aliased_method
     
     NameError:
       undefined local variable or method `parent_aliased_method' for #<AnyInstanceSpec::ParentClass:0x2bb88>
     # ./spec/rspec/mocks/any_instance_spec.rb:20:in `caller_of_parent_aliased_method'
     # ./spec/rspec/mocks/any_instance_spec.rb:210:in `block (4 levels) in <module:Mocks>'

  2) Marshal extensions #dump when rspec-mocks has been fully initialized applying and unapplying patch is idempotent
     Failure/Error: expect { Marshal.dump(obj) }.to raise_error(TypeError)
       expected TypeError but nothing was raised
     # ./spec/rspec/mocks/marshal_extension_spec.rb:46:in `block (4 levels) in <top (required)>'
```

BTW, the reason the code worked before these `ruby2_keywords` seems accidental and due to a bug/inconsistency in CRuby: https://bugs.ruby-lang.org/issues/18625

### DIFF
--- a/lib/rspec/mocks/argument_list_matcher.rb
+++ b/lib/rspec/mocks/argument_list_matcher.rb
@@ -46,7 +46,7 @@ module RSpec
         @expected_args = expected_args
         ensure_expected_args_valid!
       end
-      ruby2_keywords :initialize if Module.private_method_defined?(:ruby2_keywords)
+      ruby2_keywords :initialize if respond_to?(:ruby2_keywords, true)
 
       # @api public
       # @param [Array] actual_args
@@ -71,6 +71,7 @@ module RSpec
 
         Support::FuzzyMatcher.values_match?(expected_args, actual_args)
       end
+      ruby2_keywords :args_match? if respond_to?(:ruby2_keywords, true)
 
       # @private
       # Resolves abstract arg placeholders like `no_args` and `any_args` into
--- a/lib/rspec/mocks/matchers/receive.rb
+++ b/lib/rspec/mocks/matchers/receive.rb
@@ -62,7 +62,7 @@ module RSpec
             @recorded_customizations << ExpectationCustomization.new(method, args, block)
             self
           end
-          ruby2_keywords(method) if Module.private_method_defined?(:ruby2_keywords)
+          ruby2_keywords(method) if respond_to?(:ruby2_keywords, true)
         end
 
       private
--- a/lib/rspec/mocks/message_expectation.rb
+++ b/lib/rspec/mocks/message_expectation.rb
@@ -139,9 +139,12 @@ module RSpec
       #   counter.increment
       #   expect(counter.count).to eq(original_count + 1)
       def and_call_original
-        wrap_original(__method__) do |original, *args, &block|
-          original.call(*args, &block)
+        block = lambda do |original, *args, &b|
+          original.call(*args, &b)
         end
+        block = block.ruby2_keywords if block.respond_to?(:ruby2_keywords)
+
+        wrap_original(__method__, &block)
       end
 
       # Decorates the stubbed method with the supplied block. The original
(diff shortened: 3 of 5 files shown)


eregon:
This change is also needed for Ruby 3.2 if the fix for https://bugs.ruby-lang.org/issues/18625 gets in.
So what I did is build [that change](https://github.com/ruby/ruby/pull/5645) locally and run rspec-mocks specs without & with this patch.
Without this PR, there are [23 failures](https://gist.github.com/eregon/5f1c8e46b69eec9e4f13b44a0c3fb15d). With this PR, no failures.

Since 3.2 needs that fix, it seems we will need a rspec release, so Bundler can get this fix, and Bundler is tested in ruby/ruby's CI. Bundler's specs can't just point to a github branch due to rspec inter-gem version restrictions: https://github.com/ruby/ruby/pull/5684/checks

pirj:
Re-running jobs to see if [a weird failure](https://github.com/rspec/rspec-mocks/runs/5643874497?check_suite_focus=true) was temporary.

JonRowe:
I'm fine with this if we can get a passing build, although if there is an easy way to get a spec (or specs) that would fail without these changes on one of the existing MRI builds that'd be a nice to have.

eregon:
I'm really confused how this PR can cause 2.5/2.6 to fail rspec-rails specs, on Linux but not Windows, when `ruby2_keywords` doesn't exist in Ruby < 2.7 (and even if it's defined there it would noop).
I'll schedule a few runs on my fork to try to debug it.

eregon:
CI shall be green now (here is [my run](https://github.com/eregon/rspec-mocks/actions/runs/2034082804)), except 2.3 Windows which is already failing on main.

> although if there is an easy way to get a spec (or specs) that would fail without these changes on one of the existing MRI builds that'd be a nice to have.

This is not possible, the code before this PR only works "thanks" to a CRuby bug of not resetting the `ruby2_keywords` flag correctly (https://bugs.ruby-lang.org/issues/18625).
So this PR has no effect on CRuby <= 3.1, but it does fix TruffleRuby (which doesn't have that bug) and CRuby 3.2 (if the bug is fixed in CRuby 3.2, let's hope so).
I verified this change is required to pass existing specs on a local build of CRuby 3.2-dev with the fix (https://github.com/rspec/rspec-mocks/pull/1464#issuecomment-1074998528).

pirj:
I recalled that there was a reason to switch to `Module.private_method_defined?` in the first place. There's some discussion here https://github.com/rspec/rspec-mocks/pull/1385#issuecomment-755829430

Just to have some records of what the fix entailed, what is the difference between your original implementation of this PR and the final result, @eregon?
I can spot a call of `ruby2_keywords` on an instance of a `Proc` (TIL) 🤯

eregon:
> Just to have some records of what the fix entailed, what is the difference between your original implementation of this PR and the final result, @eregon?

You can see it with the `Compare` button above, it links to [this diff](https://github.com/rspec/rspec-mocks/compare/9703d86c2a6ccf3b4b2ca8aff5ea39d181f472ad..114e6ba75512204a4a5873ee2ab405e09bafd559). So it's just restoring that check and adding comment why it's done that way (I otherwise used the well-known `if respond_to?(:ruby2_keywords, true)` check).

eregon:
Could I ask for a release of rspec-mocks including this fix?
I tried various things, but I can't get the Bundler tests in ruby/ruby CI to use rspec-mocks master (https://github.com/ruby/ruby/pull/5684)

JonRowe:
Released as 3.11.1

PR #1469, opened 2022-04-19
Update README.md

ro31337:
Fix the typo `die` -> `dice`


JonRowe:
Thanks, but `die` is the singular form of `dice` and is correct in this instance.

PR #1470, opened 2022-04-21
Use a better and more reliable check for whether a method is the same as Class#new

eregon:
* See https://bugs.ruby-lang.org/issues/18729#note-5

The added test would fail without the change.
This also helps for TruffleRuby support (https://github.com/rspec/rspec-metagem/issues/68), as TruffleRuby does not plan to replicate that CRuby inconsistency/bug (https://bugs.ruby-lang.org/issues/18729).

### DIFF
--- a/spec/rspec/mocks/partial_double_spec.rb
+++ b/spec/rspec/mocks/partial_double_spec.rb
@@ -622,6 +622,24 @@ module RSpec
           end
         end
 
+        context "on a class with a twice-aliased `new`" do
+          it 'uses the method signature from `#initialize` for arg verification' do
+            if RUBY_VERSION.to_i < 3
+              pending "Failing due to Ruby 2's Method#== being too strict"
+            end
+
+            subclass = Class.new(klass) do
+              class << self
+                alias_method :_new, :new
+                alias_method :new, :_new
+              end
+            end
+
+            prevents(/arguments/) { allow(subclass).to receive(:new).with(1) }
+            allow(subclass).to receive(:new).with(1, 2)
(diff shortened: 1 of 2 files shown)


spec/rspec/mocks/partial_double_spec.rb
```diff
@@ -622,6 +622,24 @@ class << self
           end
         end
 
+        context "on a class with a twice-aliased `new`" do
+          it 'uses the method signature from `#initialize` for arg verification' do
+            if RUBY_VERSION.to_i < 3
+              pending "Failing due to Ruby 2's Method#== being too strict"
+            end
```
JonRowe, line 629:
```suggestion
          it 'uses the method signature from `#initialize` for arg verification', :if => RUBY_VERSION.to_i > = 3 do
```
Typically we use meta data rather than pending in spec, and flipping the conditional here should also fix the 1.8.7 error (which apparently supports this :joy:)

eregon: Right, thanks, I applied it and fixed the `> =`.

eregon:
Unfortunately this doesn't seem to work on Ruby master, even though it does work on 3.0 and 3.1.
I've filed https://bugs.ruby-lang.org/issues/18751, it seems a CRuby regression.

eregon:
I fixed that CRuby issue, and with a rerun of `main` on my fork ruby-head now passes :tada: 
https://github.com/eregon/rspec-mocks/runs/7931839710?check_suite_focus=true

PR #1473, opened 2022-05-25
Fix `with` for instance doubles didn't verify Ruby 3 keyword arguments

nashbridges:
This is a follow-up to https://github.com/rspec/rspec-mocks/pull/1394.

The change introduced in that PR handle a pure double, but not an instance double. 

The reason is that `VerifyingMessageExpectation` is a subclass of `MessageExpectation` with `with` method being overridden. That makes `ruby2_keywords` definition declared in the superclass to be "dropped" (and therefore to confuse the [`Hash.ruby2_keywords_hash?`](https://github.com/nashbridges/rspec-mocks/blob/7a0e363f3c081f0a49a3340e9eb0f41eba76af82/lib/rspec/mocks/argument_list_matcher.rb#L66-L68) check). 

I'm not fully sure if this is an expected Ruby behavior, but the current fix is to repeat the definition in the subclass.

### DIFF
--- a/lib/rspec/mocks/verifying_message_expectation.rb
+++ b/lib/rspec/mocks/verifying_message_expectation.rb
@@ -31,6 +31,7 @@ module RSpec
           end
         end
       end
+      ruby2_keywords(:with) if respond_to?(:ruby2_keywords, true)
 
     private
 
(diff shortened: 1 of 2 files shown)


pirj (approved):
Thank you!

nashbridges:
@pirj thanks for reviewing this!

eregon:
> I'm not fully sure if this is an expected Ruby behavior, but the current fix is to repeat the definition in the subclass.

It is expected, there is "inheritance" of `ruby2_keywords` or anything like that, each method which needs to delegate kwargs through `*args` needs `ruby2_keywords`.

PR #1477, opened 2022-08-10
Document receive(:msg).with(satisfy{|arg| ... })

tjallingvanderwal:
This is the combination I've been looking for, but that has eluded me for years.

The [readme](https://github.com/rspec/rspec-mocks#argument-matchers) and the table in the  [relish docs](https://relishapp.com/rspec/rspec-mocks/v/3-11/docs/setting-constraints/matching-arguments) both mention that any matcher can be used, but it never occurred to me to combine `with` with `satisfy`.

I've added 2 example usages to the readme and added a Cucumber story to document this combination.

The [documentation](https://www.rubydoc.info/gems/rspec-expectations/3.11.0/RSpec/Matchers#satisfy-instance_method) for `satisfy` suggest to write a custom matcher, but that feels over the top for many one-off situations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ expect(double).to receive(:msg).with(hash_including(:a => 5)) # first arg is a h
 expect(double).to receive(:msg).with(array_including(5)) # first arg is an array with 5 as one of the key-values
 expect(double).to receive(:msg).with(hash_excluding(:a => 5)) # first arg is a hash without a: 5 as one of the key-values
 expect(double).to receive(:msg).with(start_with('a')) # any matcher, custom or from rspec-expectations
-expect(double).to receive(:msg).with(satisfy{|data| data.dig(:a, :b, :c) == 5 }) # any expectation
+expect(double).to receive(:msg).with(satisfy { |data| data.dig(:a, :b, :c) == 5 }) # assert anything you want
 ```
 
 ## Receive Counts
--- a/README.md
+++ b/README.md
@@ -278,8 +278,8 @@ expect(double).to receive(:msg).with(1, duck_type(:abs, :div), "b") #2nd argumen
 expect(double).to receive(:msg).with(hash_including(:a => 5)) # first arg is a hash with a: 5 as one of the key-values
 expect(double).to receive(:msg).with(array_including(5)) # first arg is an array with 5 as one of the key-values
 expect(double).to receive(:msg).with(hash_excluding(:a => 5)) # first arg is a hash without a: 5 as one of the key-values
-expect(double).to receive(:msg).with(start_with('a')) # any matcher from rspec-expectations
-expect(double).to receive(:msg).with(satisfy{|arg| arg[:a] == 5 }) # any expectation
+expect(double).to receive(:msg).with(start_with('a')) # any matcher, custom or from rspec-expectations
+expect(double).to receive(:msg).with(satisfy{|data| data.dig(:a, :b, :c) == 5 }) # any expectation
 ```
 
 ## Receive Counts
--- a/README.md
+++ b/README.md
@@ -278,6 +278,8 @@ expect(double).to receive(:msg).with(1, duck_type(:abs, :div), "b") #2nd argumen
 expect(double).to receive(:msg).with(hash_including(:a => 5)) # first arg is a hash with a: 5 as one of the key-values
 expect(double).to receive(:msg).with(array_including(5)) # first arg is an array with 5 as one of the key-values
 expect(double).to receive(:msg).with(hash_excluding(:a => 5)) # first arg is a hash without a: 5 as one of the key-values
+expect(double).to receive(:msg).with(start_with('a')) # any matcher from rspec-expectations
+expect(double).to receive(:msg).with(satisfy{|arg| arg[:a] == 5 }) # any expectation
 ```
 
 ## Receive Counts
--- a/features/setting_constraints/matching_arguments.feature
+++ b/features/setting_constraints/matching_arguments.feature
@@ -111,7 +111,7 @@ Feature: Matching arguments
   Scenario: Using a RSpec matcher
     Given a file named "rspec_matcher_spec.rb" with:
       """ruby
-      RSpec.describe "Using a RSpec matcher" do
+      RSpec.describe "Using any matcher from rspec-expectations" do
         let(:dbl) { double }
         before { expect(dbl).to receive(:foo).with(a_collection_containing_exactly(1, 2)) }
 
@@ -132,6 +132,36 @@ Feature: Matching arguments
       | #<Double (anonymous)> received :foo with unexpected arguments |
       | expected: (a collection containing exactly 1 and 2)           |
       | got: ([1, 3])                                                 |
+
+  Scenario: Using satisfy for any expectation
+    Given a file named "rspec_satisfy_spec.rb" with:
+      """ruby
+      RSpec.describe "Using satisfy for any expectation" do
+        let(:dbl) { double }
+        before do
+          expectation = satisfy do |arg|
+            arg[:a] == 5
+          end
+          expect(dbl).to receive(:foo).with(expectation)
+        end
+
+        it "passes when the expectation is true" do
+          dbl.foo(a: 5)
+        end
+
+        it "fails when the expectation is false" do
+          dbl.foo(a: 3)
+        end
+      end
+      """
+    When I run `rspec rspec_satisfy_spec.rb`
+    Then it should fail with the following output:
+      | 2 examples, 1 failure                                         |
+      |                                                               |
+      | Failure/Error: dbl.foo(a: 3)                                  |
+      | #<Double (anonymous)> received :foo with unexpected arguments |
+      | expected: (satisfy expression `arg[:a] == 5`)                 |
+      |      got: ({:a=>3})                                           |
 
   Scenario: Using a custom matcher
     Given a file named "custom_matcher_spec.rb" with:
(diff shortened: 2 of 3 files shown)


features/setting_constraints/matching_arguments.feature
```diff
@@ -47,10 +47,10 @@ Feature: Matching arguments
       |     expected: (1, anything, /bar/)                              |
       |          got: (1, nil, "other")                                 |
 
-  Scenario: Using a RSpec matcher
+  Scenario: Using any matcher from rspec-expectations
     Given a file named "rspec_matcher_spec.rb" with:
       """ruby
-      RSpec.describe "Using a RSpec matcher" do
+      RSpec.describe "Using any matcher from rspec-expectations" do
```
JonRowe, line 53: The original is correct, mocks matchers are also available here, as is any matcher from any library

features/setting_constraints/matching_arguments.feature
```diff
@@ -72,6 +72,36 @@ Feature: Matching arguments
       | expected: (a collection containing exactly 1 and 2)           |
       | got: ([1, 3])                                                 |
 
+  Scenario: Using satisfy for any expectation
```
JonRowe, line 75: How about, `Using satisfy for complex custom expecations`

features/setting_constraints/matching_arguments.feature
```diff
@@ -72,6 +72,36 @@ Feature: Matching arguments
...
+    Given a file named "rspec_satisfy_spec.rb" with:
+      """ruby
+      RSpec.describe "Using satisfy for any expectation" do
```
JonRowe, line 78: As above if changed.

features/setting_constraints/matching_arguments.feature
```diff
@@ -72,6 +72,36 @@ Feature: Matching arguments
...
+        let(:dbl) { double }
+        before do
+          expectation = satisfy do |arg|
+            arg[:a] == 5
```
JonRowe, line 82: Prehaps make this more complex to demonstrate why `satisfy` is needed

JonRowe (commented):
We're happy to improve documentation here, I've a bit of feedback for the scenario.

features/setting_constraints/matching_arguments.feature
```diff
@@ -47,10 +47,10 @@ Feature: Matching arguments
       |     expected: (1, anything, /bar/)                              |
       |          got: (1, nil, "other")                                 |
 
-  Scenario: Using a RSpec matcher
+  Scenario: Using any matcher from rspec-expectations
```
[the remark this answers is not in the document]
tjallingvanderwal: I've reverted this change.

features/setting_constraints/matching_arguments.feature
tjallingvanderwal: This shows my own use case: asserting something about a value deep inside a blob of data.

Nested usage of `hash_including` would work for this use case, but becomes unwieldy fast with each level added.
 
```ruby
expectation = satisfy do |data|
  data.dig(:a, :b, :c) == 5
end
```

JonRowe (approved):

features/setting_constraints/matching_arguments.feature
JonRowe: Great, perfect.

README.md
```diff
@@ -278,6 +278,8 @@ expect(double).to receive(:msg).with(1, duck_type(:abs, :div), "b") #2nd argumen
 expect(double).to receive(:msg).with(hash_including(:a => 5)) # first arg is a hash with a: 5 as one of the key-values
 expect(double).to receive(:msg).with(array_including(5)) # first arg is an array with 5 as one of the key-values
 expect(double).to receive(:msg).with(hash_excluding(:a => 5)) # first arg is a hash without a: 5 as one of the key-values
+expect(double).to receive(:msg).with(start_with('a')) # any matcher, custom or from rspec-expectations
+expect(double).to receive(:msg).with(satisfy{|data| data.dig(:a, :b, :c) == 5 }) # any expectation
```
pirj, line 282: I don't quite understand why the comment here says "expectation", while this is also an example of using a matcher as an argument matcher for `receive`. Is `satisfy` somehow different from anything above?

pirj, line 282:
```suggestion
expect(double).to receive(:msg).with(satisfy { |data| data.dig(:a, :b, :c) == 5 }) # any expectation
```

pirj, line 282:
```suggestion
expect(double).to receive(:msg).with(satisfy { |data| data.dig(:a, :b, :c) == 5 })
```

features/setting_constraints/matching_arguments.feature
```diff
@@ -72,6 +72,36 @@ Feature: Matching arguments
...
+          expect(dbl).to receive(:foo).with(expectation)
+        end
+
+        it "passes when the expectation is true" do
+          dbl.foo(a: { b: { c: 5 } })
+        end
+
+        it "fails when the expectation is false" do
+          dbl.foo(a: { b: { c: 3 } })
+        end
```
pirj, line 93: It is very subjective, but I would argue about this example's style. Firstly, the expectation is made in a hook.
The term `expectation` from my perspective is used incorrectly, the `satisfy` matcher is used as an argument matcher here.

How do you feel about this instead:
```ruby
      RSpec.describe "Using satisfy for complex custom expecations" do
        let(:dbl) { double }

        def a_b_c_equals_5
          satisfy { |data| data[:a][:b][:c] == 5 }
        end

        it "passes when the expectation is true" do
          expect(dbl).to receive(:foo).with(a_b_c_equals_5)
          dbl.foo(a: { b: { c: 5 } })
        end

        it "fails when the expectation is false" do
          expect(dbl).to receive(:foo).with(a_b_c_equals_5)
          dbl.foo(a: { b: { c: 3 } })
        end
```

pirj (changes requested):
Thank you!

README.md
tjallingvanderwal: > Is satisfy somehow different from anything above?

I think that technically, from a library perspective, you're right that there isn't much difference between `satisfy` and the other examples.

But for a RSpec user, I think it is nice that the documentation points out that `satisfy` allows you to assert anything you want.

features/setting_constraints/matching_arguments.feature
tjallingvanderwal: > It is very subjective, but I would argue about this example's style. Firstly, the expectation is made in a hook. 
> ...
> How do you feel about this instead:

I followed the style of other stories in this file, but I like yours better, so I adopted your suggestion.


> The term `expectation` from my perspective is used incorrectly, the `satisfy` matcher is used as an argument matcher here.
 
As I understand it, the goal is to verify an expectation. The mechanism is to use a matcher. If I've mixed things up, or if you want to emphasise one or the other in the documentation, let me know what changes you'd like to see.

pirj (approved):
Thank you!

features/setting_constraints/matching_arguments.feature
```diff
@@ -72,6 +72,36 @@ Feature: Matching arguments
...
+      end
+      """
+    When I run `rspec rspec_satisfy_spec.rb`
+    Then it should fail with the following output:
+      | 2 examples, 1 failure                                         |
+      |                                                               |
+      | Failure/Error: dbl.foo({ :a => { :b=> { :c => 3 } } })        |
+      | #<Double (anonymous)> received :foo with unexpected arguments |
+      | expected: (satisfy expression `data[:a][:b][:c] == 5`)        |
```
pirj, line 102: Wondering if some option would make cucumber's error message more verbose on CI to avoid running 1.8.7 locally.
```
 expected ".F\n\nFailures:\n\n  1) Using satisfy for complex custom expecations fails when the expectation is f...y_spec.rb:13 # Using satisfy for complex custom expecations fails when the expectation is false\n\n" to include "expected: (satisfy expression `data[:a][:b][:c] == 5`)" (RSpec::Expectations::ExpectationNotMetError)
./features/step_definitions/additional_cli_steps.rb:20:in `/^it should fail with the following output:$/'
features/setting_constraints/matching_arguments.feature:97:in `Then it should fail with the following output:'
```

If not, [this](https://github.com/rbenv/homebrew-tap) helped me last time.

pirj:
https://github.com/cucumber/cucumber-expressions/pull/272 https://github.com/cucumber/cucumber-expressions/pull/273 for ruby-head

Something weird is wrong with 3.2, IDK

For 1.8.7 the failure diff is
```diff
+         expected: (satisfy block)
-expected: (satisfy expression `data[:a][:b][:c] == 5`)
```
which is due to the [lack of Ripper](https://github.com/rspec/rspec-expectations/blob/d0bb212c371fb6381cae2f9b86c4a025881f7ead/spec/rspec/matchers/built_in/satisfy_spec.rb#L18) in [Ruby < 1.9.2 ](https://github.com/rspec/rspec-support/blob/22e567e38c5d7f2e1a6d9302d067875b2774f50c/lib/rspec/support/ruby_features.rb#L129).
I intend to clamp this feature on 1.8.7 cucumber runs.

pirj:
The Ruby 3.2 rspec-core related failure reproduced in a https://github.com/rspec/rspec-core/pull/3046 build, so I intend to fix it there.

@tjallingvanderwal Thank you for this nice doc addition! Please accept my apologies for the long processing.

tjallingvanderwal:
> @tjallingvanderwal Thank you for this nice doc addition! Please accept my apologies for the long processing.

No apologies needed. Thank you for finishing my pull request.

PR #1485, opened 2022-09-09
Don't convert keywords to hash in #send for verifying doubles on Ruby 3

honigc:
The current re-implementation of `send` on verifying doubles uses `*args` by itself. Everything works fine on `main` for Ruby 2 or lower, but on Ruby 3 this converts keyword arguments to a Hash element inside the `args` array, which then gets passed to `__send__` and then `super` and (correctly) causes a failure when the method gets a Hash instead of keywords.

Without this patch, the new test passes on Ruby 2.7.6 and fails on Ruby 3.1.2:
```
Failures:

  1) Expected argument verification (when `#with` is called) when doubling a loaded class for a method with keyword args when using `send` matches against keyword arguments
     Failure/Error: dbl.send(:m, :k => 1)
     
       #<InstanceDouble(#<Class:0x00007fbd16cf1ba8>) (anonymous)> received :m with unexpected arguments
         expected: ({:k=>1}) (keyword arguments)
              got: ({:k=>1}) (options hash)
       Diff:
```

With this patch, all tests pass on all Ruby versions.

### DIFF
--- a/lib/rspec/mocks/verifying_double.rb
+++ b/lib/rspec/mocks/verifying_double.rb
@@ -42,11 +42,13 @@ module RSpec
       ensure
         @__sending_message = nil
       end
+      ruby2_keywords :__send__ if respond_to?(:ruby2_keywords, true)
       $VERBOSE = old
 
       def send(name, *args, &block)
         __send__(name, *args, &block)
       end
+      ruby2_keywords :send if respond_to?(:ruby2_keywords, true)
 
       def initialize(doubled_module, *args)
         @doubled_module = doubled_module
--- a/spec/rspec/mocks/verifying_doubles/expected_arg_verification_spec.rb
+++ b/spec/rspec/mocks/verifying_doubles/expected_arg_verification_spec.rb
@@ -143,6 +143,15 @@ module RSpec
                 dbl.kw_args_method(1, {:required_arg => 2, :optional_arg => 3})
               end
             end
+
+            context "when using `send`" do
+              let(:dbl) { instance_double(Class.new { eval "def m(k:); end" }) }
+
+              it "matches against keyword arguments" do
+                expect(dbl).to receive(:m).with(:k => 1)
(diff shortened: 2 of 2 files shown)


JonRowe:
Instead please use `ruby2_keywords :send` as it parses on all Rubies, the `**` won't parse on older Rubies

JonRowe (dismissed):
Instead please use `ruby2_keywords :__send__` as it parses on all Rubies, the `**` won't parse on older Rubies and changes the semantics

honigc:
@JonRowe Done

honigc:
Updated again to use `eval` to set up the kwarg method in tests, to avoid syntax errors in old Rubies

JacobEvelyn:
Hi @JonRowe! We're eager to see this merged—anything I can do to help move it forward?

pirj (approved):
Looks good, thank you.

@JonRowe The build isn't triggered, I can't see why.

JonRowe:
Can this get rebased? That might trigger the build again

pirj:
Yesterday I rebased, added a separate changelog commit and force-pushed, no build :thinking: 
On the Actions tab, it says "Workflow runs completed with no jobs".

JonRowe:
Theres no changelog or seperate commit here, so no build... You must have pushed it to a different branch?

pirj:
> You must have pushed it to a different branch?

Apparently, [yes](https://github.com/rspec/rspec-mocks/compare/ruby3-verifying-double-send) :facepalm:

pirj:
> ERROR: Permission to panorama-ed/rspec-mocks.git denied to pirj

Might be "allow edits for maintainers" is turned off for this PR?

@honigc Do you mind ticking that checkbox please, or rebase and push again for the build to trigger?

JonRowe:
Thanks for the work on this @honigc sorry it took so long to merge.

PR #1487, opened 2022-10-05
replacing lambdas with procs

C-Miles:
We tried using `.and_invoke` and found the lambdas were being called with an argument.


JonRowe:
If you pass an argument to your method calls, then lambdas or procs passed to `and_invoke` are called with those arguments, so this example is correct, as there are no arguments passed.

C-Miles:
@JonRowe right on! that makes sense, thank you!

Issue #1495, opened 2022-11-03
Confused over instance doubles, keyword args, with, and Ruby 3.2 previews


philnash:
### Subject of the issue

I ran a bunch of specs on my project through GitHub actions and a bunch failed recently against Ruby head. I'm not sure whether it's the fault of Ruby or RSpec or my own misunderstanding, so I thought I would report it here first.

The issue came down to expectations for instance doubles for objects that take keyword arguments and how to test them. Please see the steps to reproduce below for details.

### Your environment
* Ruby version: 3.2.0-preview2
* rspec-mocks version: 3.12.0

### Steps to reproduce

```ruby
# frozen_string_literal: true

begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rspec", "3.12.0"
end

puts "Ruby version is: #{RUBY_VERSION}"
require 'rspec/autorun'

class TestObject
  def kw_args_method(a:, b:); end
end

RSpec.describe TestObject do
  it "expects to receive keyword args" do
    dbl = instance_double(TestObject)
    expect(dbl).to receive(:kw_args_method).with(a: 1, b: 2)
    dbl.kw_args_method(a: 1, b: 2)
  end

  it "expects to receive keyword args with a hash" do
    dbl = instance_double(TestObject)
    expect(dbl).to receive(:kw_args_method).with({a: 1, b: 2})
    dbl.kw_args_method(a: 1, b: 2)
  end
end
```

### Expected behavior
I expected the first test to pass, since the expectation is that the method is called with two keyword arguments and then the method is called with two keyword arguments.

I'd expect the second test to fail because the expectation is that the method is called with a hash, but the method is called with keyword arguments.

### Actual behavior
The first test fails and the second passes.

```
Failures:

  1) TestObject expects to receive keyword args
     Failure/Error: dbl.kw_args_method(a: 1, b: 2)
     
       #<InstanceDouble(TestObject) (anonymous)> received :kw_args_method with unexpected arguments
         expected: ({:a=>1, :b=>2}) (keyword arguments)
              got: ({:a=>1, :b=>2}) (options hash)
       Diff:
       
     # test_spec.rb:28:in `block (2 levels) in <main>'
```

However, just to throw some extra information in:

* On Ruby 3.1 with RSpec 3.12.0 both tests pass.
[...]

PR #1497, opened 2022-11-05
Reproduction case for Ruby 3.2 and kwargs detection

pirj:
Related to #1495

### DIFF
--- a/spec/rspec/mocks/matchers/receive_spec.rb
+++ b/spec/rspec/mocks/matchers/receive_spec.rb
@@ -31,6 +31,16 @@ module RSpec
         end
       end
 
+      # FIXME: this is defined here to prevent
+      # "warning: method redefined; discarding old kw_args_method"
+      # because shared examples are evaluated several times.
+      # When we flatten those shared examples in RSpec 4 because
+      # of no "should" syntax, it will become possible to put this
+      # class definition closer to examples that use it.
+      class TestObject
+        def kw_args_method(a:, b:); end
+      end
+
       shared_examples "a receive matcher" do |*options|
         it 'allows the caller to configure how the subject responds' do
           wrapped.to receive(:foo).and_return(5)
@@ -108,10 +118,6 @@ module RSpec
               end
 
               expect(receiver.foo(kw: :arg)).to eq(:arg)
-            end
-
-            class TestObject
-              def kw_args_method(a:, b:); end
             end
 
             it "expects to receive keyword args" do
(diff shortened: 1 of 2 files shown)


pirj:
@eregon Do you happen to know what might have caused the failures on [3.2.0-preview2](https://github.com/rspec/rspec-mocks/actions/runs/3400302848/jobs/5654635455) and [ruby-head](https://github.com/rspec/rspec-mocks/actions/runs/3400302848/jobs/5654635939) while passed [just fine with 3.1](https://github.com/rspec/rspec-mocks/actions/runs/3400302848/jobs/5654635494)?

Is this still correct for Ruby 3.2?
```ruby

        if RSpec::Support::RubyFeatures.distincts_kw_args_from_positional_hash? && expected_args == actual_args
          expected_hash = expectation.expected_args.last
          actual_hash = args_for_multiple_calls.last.last
            (Hash.ruby2_keywords_hash?(expected_hash) != Hash.ruby2_keywords_hash?(actual_hash))
            actual_args += Hash.ruby2_keywords_hash?(actual_hash) ? " (keyword arguments)" : " (options hash)"
            expected_args += Hash.ruby2_keywords_hash?(expected_hash) ? " (keyword arguments)" : " (options hash)"
          end
        end
```


2.2 failure is unrelated, to be fixed by #1498

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -109,6 +123,18 @@ module Mocks
...
+            it "expects to receive keyword args" do
+              dbl = instance_double(TestObject)
+              expect(dbl).to receive(:kw_args_method).with(a: 1, b: 2)
+              dbl.kw_args_method(a: 1, b: 2)
+            end
+
+            it "expects to receive keyword args with a hash" do
+              dbl = instance_double(TestObject)
+              expect(dbl).to receive(:kw_args_method).with({a: 1, b: 2})
+              dbl.kw_args_method(a: 1, b: 2)
```
eregon, line 136: Isn't this expected to fail? It expects a positional Hash but is given kwargs.
But I guess `expect(dbl).to receive` actually creates a method with no kwargs so then it should pass?

eregon:
That code looks fine. I suspect it's a missing `ruby2_keywords` on the chain of calls, and so the flag is lost somewhere.
You can debug by printing `Hash.ruby2_keywords_hash?(h)` before the exception happens.

eregon:
And 3.2 fixes a bug where some missing `ruby2_keywords` accidentally kept the flag: https://github.com/ruby/ruby/blob/master/NEWS.md

pirj:
Thank you so much for the explanation, @eregon !

spec/rspec/mocks/matchers/receive_spec.rb
pirj: This indeed should have failed on earlier Ruby versions, too.

pirj:
Superseded by #1514

PR #1498, opened 2022-11-05
Pin childprocess on Ruby 2.2

pirj:
Borrowed from https://github.com/rspec/rspec-support/commit/a816ca10e87ccf68023752af6a8824496d336f85

Failure example https://github.com/rspec/rspec-mocks/actions/runs/3400302848/jobs/5654635850


pirj:
@JonRowe Thanks for restarting the build for JRuby 1.7 in Ruby 1.8 mode. It appears to be related to that legacy Bundler API issue under load messing up with dependency constraints.

PR #1502, opened 2022-11-18
Fixup argument forwarding on ruby-3.2

### LINKED ISSUE #1495: Confused over instance doubles, keyword args, with, and Ruby 3.2 previews

ojab:
Fixes #1495
close #1497

Depends on https://github.com/rspec/rspec-support/pull/553

### DIFF
--- a/lib/rspec/mocks/verifying_proxy.rb
+++ b/lib/rspec/mocks/verifying_proxy.rb
@@ -160,6 +160,7 @@ module RSpec
         validate_arguments!(args)
         super
       end
+      ruby2_keywords :proxy_method_invoked if respond_to?(:ruby2_keywords, true)
 
       def validate_arguments!(actual_args)
         @method_reference.with_signature do |signature|


pirj (approved):

pirj:
Would you throw in some examples, @ojab? You could probably reuse ones from #1497

JonRowe:
Some specs would be good 👍

PR #1506, opened 2022-11-28
Improve error generator

### LINKED ISSUE #1504: Unexpected arguments error (keywords vs hash) has misleading message with another argument
### Subject of the issue

This is a continuation of rspec/rspec-mocks#1505.  The fix https://github.com/rspec/rspec-mocks/pull/1461 seems incomplete.

When my test contains a wrong `with` expectation, where the actual arguments are keywords and I use a hash (or the other way around), RSpec reports an error. And in most cases the message correctly points out the problem, per 1461.

But in my case I got a misleading message:

```
Failures:

  1) MyClass#outer calls #inner with a tag and a hash
     Failure/Error: inner(Tag.new("a tag"), a_hash)
     
       #<MyClass:0x00007fca6cf24638> received :inner with unexpected arguments
         expected: (#<IdTag:0x00007fca69849668 @id="a tag">, {:one=>1})
              got: (#<IdTag:0x00007fca69831298 @id="a tag">, {:one=>1})
       Diff:
       @@ -1 +1 @@
       -[#<IdTag:0x00007fca69849668 @id="a tag">, {:one=>1}]
       +[#<IdTag:0x00007fca69831298 @id="a tag">, {:one=>1}]
       
     # ./spec/bug_spec.rb:22:in `outer'
     # ./spec/bug_spec.rb:30:in `block (3 levels) in <top (required)>'
```

I've traced it down to `IdTag` instances comparing as equal yet having different `inspect` representations.

### Your environment
* Ruby version: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-gnu]
* rspec-expectations version: 3.12.0

### Steps to reproduce

spec/bug_spec.rb:

```rb
class IdTag
  attr_accessor :id

  def initialize(id)
    @id = id
  end

  def ==(other)
    other.class == self.class && other.id == id
  end
end

Tag = IdTag
# these would make the bug go away
# Tag = String
# Tag = Struct.new(:id)

class MyClass
  def inner(tag, hash)
    puts "Inner: #{tag.inspect}, #{hash.inspect}"
  end

  def outer
    a_hash = { one: 1 }
    inner(Tag.new("a tag"), a_hash)
  end
end

describe MyClass do
  describe "#outer" do
    it "calls #inner with a tag and a hash" do
      # This expectation is wrong: it actually expects keyword arguments
      expect(subject).to receive(:inner).with(Tag.new("a tag"), one: 1)
      subject.outer
    end
  end
end
```

Gemfile:

```
source "https://rubygems.org"

gem "rspec-mocks", "= 3.12.0"
gem "rspec", "= 3.12.0"
gem "rspec-core", "= 3.12.0"
gem "rspec-expectations", "= 3.12.0"
gem "rspec-support", "= 3.12.0"

```

### Expected behavior

Even in this case it should point out the difference between keyword and hash arguments

JonRowe:
The previous implementation of our keyword vs hash highlight only added the difference when arguments matched, this leads to confusion because the argument matcher check this in reverse (e.g. if the keywords / vs hash doesn't match it fast fails) meaning that RSpec can be ambiguous about whats failing to be more consitent we can show the extra diff information all the time if they mis match.

As an aside this PR also fixes an inconsitency with our empty diff "trim" logic where colourisation would produce an empty:

```
Diff:
```

As well as this the PR improves the specs by adding partial double checks as well (as given the keyword semantics these can exhibit different behaviour).

Fixes #1504

### DIFF
--- a/lib/rspec/mocks/error_generator.rb
+++ b/lib/rspec/mocks/error_generator.rb
@@ -269,13 +269,19 @@ module RSpec
         expected_args = format_args(expectation.expected_args)
         actual_args = format_received_args(args_for_multiple_calls)
 
-        if RSpec::Support::RubyFeatures.distincts_kw_args_from_positional_hash? && expected_args == actual_args
+        if RSpec::Support::RubyFeatures.distincts_kw_args_from_positional_hash?
           expected_hash = expectation.expected_args.last
           actual_hash = args_for_multiple_calls.last.last
           if Hash === expected_hash && Hash === actual_hash &&
             (Hash.ruby2_keywords_hash?(expected_hash) != Hash.ruby2_keywords_hash?(actual_hash))
-            actual_args += Hash.ruby2_keywords_hash?(actual_hash) ? " (keyword arguments)" : " (options hash)"
-            expected_args += Hash.ruby2_keywords_hash?(expected_hash) ? " (keyword arguments)" : " (options hash)"
+
+            actual_description = Hash.ruby2_keywords_hash?(actual_hash) ? " (keyword arguments)" : " (options hash)"
+            expected_description = Hash.ruby2_keywords_hash?(expected_hash) ? " (keyword arguments)" : " (options hash)"
+
+            if actual_description != expected_description
+              actual_args += actual_description
+              expected_args += expected_description
+            end
           end
         end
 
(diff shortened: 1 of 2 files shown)


benoittgt (approved):
Love this addition. I am wondering if we should add in another PR a tips when it is only a diff in object's id? Like in you example between `expected_input` VS `actual_input`?

PR #1508, opened 2022-12-27
Fix mocking keyword args via block implementation in Ruby 3.2

adsteel:
👀 Check out first commit and switch Ruby versions 3.1.1 -> 3.20 to see passing -> failing spec, then check out second commit to see specs pass in both.

Default flag `ruby2_keywords` was declared a Ruby 3.0 bug and removed
in Ruby 3.2. All methods that take `*args` that we also want to use to
support keyword args must be explicitly instrumented via the
`ruby2_keywords` method.

https://github.com/ruby/ruby/blob/07c19cf551e58e1b82af5efeb0047f10588fff9f/NEWS.md

### DIFF
--- a/spec/rspec/mocks/matchers/have_received_spec.rb
+++ b/spec/rspec/mocks/matchers/have_received_spec.rb
@@ -107,6 +107,20 @@ module RSpec
           }.to_not raise_error
         end
 
+        if RSpec::Support::RubyFeatures.required_kw_args_supported?
+          it "passes if expectations against yielded keyword arguments pass" do
+            binding.eval(<<-RUBY, __FILE__, __LINE__)
+              dbl = double(:foo => nil)
+              yielded = []
+              dbl.foo(a: 1)
+              expect(dbl).to have_received(:foo) do |a:|
+                yielded << a
+              end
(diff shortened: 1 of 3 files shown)


spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -107,6 +107,16 @@ module Mocks
           }.to_not raise_error
         end
 
+        it "passes if expectations against yielded keyword arguments pass" do
```
pirj, line 110: May I suggest wrapping this in a `required_kw_args_supported?` check and `eval` like we do e.g. [here](https://github.com/rspec/rspec-mocks/blob/e03dd5e1ed216369223d43b181d40668be779763/spec/rspec/mocks/verifying_doubles/expected_arg_verification_spec.rb#L78) not to break earlier Ruby's parser?

pirj (approved):
Thank you!

spec/rspec/mocks/matchers/have_received_spec.rb
adsteel: Thank you!

JonRowe:
Released in 3.12.2

PR #1510, opened 2022-12-29
Document multiple calls with `allow_any_instance_of`

nycdotnet:
Hello,

I have attempted to improve the documentation for the multiple calls feature in combination with everyone's friend `allow_any_instance_of`.  It does have some confusing behavior as identified in the docs.  I hope that documenting the expected behavior here should at least mitigate confusion when used in combination with multiple call mocks.

Please let me know if you have any questions.  I hope this PR is acceptable without having opened an issue first.  I believe that this file corresponds with this document: https://relishapp.com/rspec/rspec-mocks/docs/configuring-responses/returning-a-value#specify-different-return-values-for-multiple-calls

Thank you.

### DIFF
--- a/features/configuring_responses/returning_a_value.feature
+++ b/features/configuring_responses/returning_a_value.feature
@@ -4,6 +4,12 @@ Feature: Returning a value
   different return values for consecutive calls. The final value will continue to be returned if
   the message is received additional times.
 
+  Note - using the multiple calls feature with `allow_any_instance_of` can result in confusing
+  behavior.  The rspec-mocks [team discourages](../../working_with_legacy_code/any_instance.feature)
+  using `allow_any_instance_of`, but its interaction with `and_return` is documented in
+  the [Working with Legacy Code](../../working_with_legacy_code/any_instance.feature#specify-different-return-values-for-multiple-calls-in-combination-with-`allow-any-instance-of`)
+  section.
+
   Scenario: Nil is returned by default
     Given a file named "returns_nil_spec.rb" with:
       """ruby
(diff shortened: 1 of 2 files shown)


JonRowe (changes requested):
Hi, can you move the extra scenario to `features/working_with_legacy_code/any_instance.feature` thats where they live as we try to semi discourage usage of it (its a huge maintenance burden for us and its considered best for legacy tests).

nycdotnet:
Sure.  Is it ok if I link to it from here?  Something like “See here for considerations when working with legacy code”.  Thanks.

JonRowe:
Its already linked to elsewhere? Do you find it difficult to find the page?

nycdotnet:
Well since this example didn't exist, I do think it was difficult to find! :-)

I fully understand and respect the desire to keep this functionality compartmentalized since it's specifically for legacy that isn't easy to test otherwise.  I think that's a worthy goal because it encourages people to write things in a better way.

My thought here is that this behavior feels more contextually specific to the behavior of `and_return`, so at least having a link to the legacy feature page with another discouragement would help to clarify how it's intended to work but would reinforce the "you should try to restructure your code if possible to make this not necessary" aka "don't do this" message.  I'll put it together that way and see what you think.  It's ok to delete the cross reference if you think it's not aligned with your goals.

Thank you!

nycdotnet:
I am investigating the build failures.

nycdotnet:
Running `script/run_build` passes locally now - sorry for the trouble.

I am curious if you have a thought on the links to `.feature` - should I remove that extension in the markdown or is it resolved in the generation?  The link on the real site would not have `.feature` on it.

Thank you.

nycdotnet:
@JonRowe sorry to trouble you, but is this PR now satisfactory?  Thank you.

JonRowe:
Thanks for the contribution, apologies it has taken so long, I'm doing some minor rewording and will then get this up on the docs.

nycdotnet:
@JonRowe excellent - many thanks for the review.

PR #1514, opened 2023-01-03
Fix argument forwarding for "receive" with Ruby 3.2.0

### LINKED ISSUE #1495: Confused over instance doubles, keyword args, with, and Ruby 3.2 previews

### LINKED ISSUE #1512: `.and_call_original` breaks on Ruby 3.2 for methods with keyword arguments
### Subject of the issue

With Ruby 3.2 rspec-mocks seems to not pass keyword arguments to the original implementation of a method when using `.and_call_original`

### Your environment
* Ruby version: 3.2.0p0
* rspec-mocks version: 3.12.1

### Steps to reproduce
I was able to create this minimal test that reproduces the issue:

```ruby
      it "handles keyword arguments correctly" do
        A = Class.new do
          def initialize(kw:)
          end
        end

        A.new(kw: 42)

        allow(A).to receive(:new).and_call_original

        A.new(kw: 42)
      end
```

### Expected behavior
The test should pass

### Actual behavior
The test fails on the second `A` instantiation with the error:
`ArgumentError: wrong number of arguments (given 1, expected 0; required keyword: kw)`

### LINKED ISSUE #1513: Ruby 3.2.0 and keyword arguments vs options hash
### Subject of the issue

I'm currently in the process of upgrading some Rails apps from Ruby 3.1.3 to 3.2.0 and am now getting 'expected keyword arguments, received options hash' failures in a bunch of specs that pass in 3.1.3.

Wasn't sure if it is related to https://github.com/rspec/rspec-expectations/issues/1370 or https://github.com/rspec/rspec-expectations/issues/1350.

### Your environment

* Ruby version: 3.2.0
* rspec-expectations version: 3.12.1

### Steps to reproduce

Rough basic example pulled from one of the Rails apps where the spec is now failing after the Ruby upgrade.

```ruby
# app/models/app.rb

class App
  def initialize(app_name, do_something: false)
    # ...
  end

  def things
    # ...
  end
end

# app/controllers/things_controller.rb

class ThingsController < ApplicationController
  def index
    @app = App.new(app_name, do_something: true)
    @things = @app.things

    render partial: "index"
  end

  private

  def app_name
    @app_name ||= params[:app_name].to_sym
  end
end

# spec/requests/things_spec.rb

RSpec.describe "/things" do
  describe "GET /" do
    before do
      app = instance_double(App)
      allow(App).to receive(:new).with(:jimmy, do_something: true).and_return(app)
      allow(app).to receive(:things).and_return([])
      
      get "/things", params: { app_name: "jimmy" }
    end

    it "returns status 200 OK" do
      expect(response).to have_http_status(:ok)
    end
  end
end
```

### Expected behavior

The spec passes with the expectations met.

### Actual behavior

```sh
Failures:

  1) /things GET / returns status 200 OK
     Failure/Error: @app = App.new(app_name, do_something: true)
     
       #<App (class)> received :new with unexpected arguments
         expected: (:jimmy, {:do_something=>true}) (keyword arguments)
              got: (:jimmy, {:do_something=>true}) (options hash)
        Please stub a default value first if message might be received with other args as well.
```

benoittgt:
It seems there is place where arguments forwarding is not ok.

I continued the work started by @pirj on https://github.com/rspec/rspec-mocks/pull/1497

fixes #1497 
fixes #1502 
fixes #1495 
fixes #1513
fixes #1512

Related:
  - https://github.com/rspec/rspec-mocks/issues/1513#issuecomment-1368473138
  - https://github.com/rspec/rspec-mocks/issues/1495

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
     strategy:
       matrix:
         ruby:
+          - '3.2'
           - '3.1'
           - '3.0'
           - 2.7
--- a/spec/rspec/mocks/matchers/receive_spec.rb
+++ b/spec/rspec/mocks/matchers/receive_spec.rb
@@ -31,6 +31,20 @@ module RSpec
         end
       end
 
+      # FIXME: this is defined here to prevent
+      # "warning: method redefined; discarding old kw_args_method"
+      # because shared examples are evaluated several times.
+      # When we flatten those shared examples in RSpec 4 because
+      # of no "should" syntax, it will become possible to put this
+      # class definition closer to examples that use it.
+      if RSpec::Support::RubyFeatures.required_kw_args_supported?
+        binding.eval(<<-RUBY, __FILE__, __LINE__)
+        class TestObject
+          def kw_args_method(a:, b:); end
+        end
+        RUBY
+      end
+
       shared_examples "a receive matcher" do |*options|
         it 'allows the caller to configure how the subject responds' do
           wrapped.to receive(:foo).and_return(5)
@@ -108,6 +122,45 @@ module RSpec
               end
 
               expect(receiver.foo(kw: :arg)).to eq(:arg)
+            end
+
+            it "expects to receive keyword args" do
+              dbl = instance_double(TestObject)
+              expect(dbl).to receive(:kw_args_method).with(a: 1, b: 2)
+              dbl.kw_args_method(a: 1, b: 2)
+            end
+
+            if RUBY_VERSION >= '3.0'
+              it "fails to expect to receive hash with keyword args" do
+                expect {
+                  dbl = instance_double(TestObject)
+                  expect(dbl).to receive(:kw_args_method).with(a: 1, b: 2)
+                  dbl.kw_args_method({a: 1, b: 2})
+                }.to fail_with do |failure|
+                  reset_all
+                  expect(failure.message)
+                    .to include('expected: ({:a=>1, :b=>2}) (keyword arguments)')
+                    .and include('got: ({:a=>1, :b=>2}) (options hash)')
+                end
+              end
+            else
+              it "expects to receive hash with keyword args" do
+                dbl = instance_double(TestObject)
+                expect(dbl).to receive(:kw_args_method).with(a: 1, b: 2)
+                dbl.kw_args_method({a: 1, b: 2})
+              end
+            end
+
+            it "expects to receive hash with a hash" do
+              dbl = instance_double(TestObject)
+              expect(dbl).to receive(:kw_args_method).with({a: 1, b: 2})
+              dbl.kw_args_method({a: 1, b: 2})
+            end
+
+            it "expects to receive keyword args with a hash" do
+              dbl = instance_double(TestObject)
+              expect(dbl).to receive(:kw_args_method).with({a: 1, b: 2})
+              dbl.kw_args_method(a: 1, b: 2)
             end
             RUBY
           end
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,8 @@ Bug Fixes:
 * Prevent a misleading error message when using `allow(...).not_to` with
   unsupported matchers. (Phil Pirozhkov, #1503)
 * Fix mocking keyword args for `have_received` with a block. (Adam Steel, #1508)
+* Fix mocking keyword args for `received` with Ruby 3.2.0. (Slava Kardakov,
+  Benoit Tigeot, Phil Pirozhkov, Benoit Daloze, #1514)
 
 ### 3.12.0 / 2022-10-26
 [Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.11.2...v3.12.0)
--- a/lib/rspec/mocks/argument_list_matcher.rb
+++ b/lib/rspec/mocks/argument_list_matcher.rb
@@ -61,7 +61,9 @@ module RSpec
         return false if expected_args.size != actual_args.size
 
         if RUBY_VERSION >= "3"
-          # if both arguments end with Hashes, and if one is a keyword hash and the other is not, they don't match
+          # If the expectation was set with keywords, while the actual method was called with a positional hash argument, they don't match.
+          # If the expectation was set without keywords, e.g., with({a: 1}), then it fine to call it with either foo(a: 1) or foo({a: 1}).
+          # This corresponds to Ruby semantics, as if the method was def foo(options).
           if Hash === expected_args.last && Hash === actual_args.last
             if !Hash.ruby2_keywords_hash?(actual_args.last) && Hash.ruby2_keywords_hash?(expected_args.last)
               return false
--- a/lib/rspec/mocks/verifying_proxy.rb
+++ b/lib/rspec/mocks/verifying_proxy.rb
@@ -160,6 +160,7 @@ module RSpec
         validate_arguments!(args)
         super
       end
+      ruby2_keywords :proxy_method_invoked if respond_to?(:ruby2_keywords, true)
 
       def validate_arguments!(actual_args)
         @method_reference.with_signature do |signature|
--- a/spec/rspec/mocks/argument_matchers_spec.rb
+++ b/spec/rspec/mocks/argument_matchers_spec.rb
@@ -388,7 +388,7 @@ module RSpec
         end
 
         if RUBY_VERSION >= "3"
-          it "fails to matches against a hash submitted as a positional argument and received as keyword arguments in Ruby 3.0 or later", :reset => true do
+          it "fails to match against a hash submitted as a positional argument and received as keyword arguments in Ruby 3.0 or later", :reset => true do
             opts = {:a => "a", :b => "b"}
             expect(a_double).to receive(:random_call).with(:a => "a", :b => "b")
             expect do
--- a/spec/rspec/mocks/stub_implementation_spec.rb
+++ b/spec/rspec/mocks/stub_implementation_spec.rb
@@ -92,6 +92,20 @@ module RSpec
         include_context "with isolated configuration"
         before { RSpec::Mocks.configuration.verify_partial_doubles = true }
         include_examples "stubbing `new` on class objects"
+
+        if RSpec::Support::RubyFeatures.required_kw_args_supported?
+          binding.eval(<<-RUBY, __FILE__, __LINE__)
+          it "handles keyword arguments correctly" do
+            klass = Class.new do
+              def initialize(kw:)
+              end
+            end
+
+            allow(klass).to receive(:new).and_call_original
+            klass.new(kw: 42)
+          end
+          RUBY
+        end
       end
     end
   end


benoittgt:
CI failures seems to be unrelated. I will look at them separately on `rspec-support`.

spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -109,6 +123,18 @@ module Mocks
...
+            it "expects to receive keyword args" do
+              dbl = instance_double(TestObject)
+              expect(dbl).to receive(:kw_args_method).with(a: 1, b: 2)
+              dbl.kw_args_method(a: 1, b: 2)
+            end
+
+            it "expects to receive keyword args with a hash" do
+              dbl = instance_double(TestObject)
+              expect(dbl).to receive(:kw_args_method).with({a: 1, b: 2})
+              dbl.kw_args_method(a: 1, b: 2)
```
pirj, line 136: This should fail, shouldn't it? We expect an options hash, and pass kwargs.

paddor:
This branches fixes my test suite on Ruby 3.2. 👍

spec/rspec/mocks/matchers/receive_spec.rb
benoittgt: Good point. I am wondering if it is a side effect of changes in 3.2. Like here. @eregon any idea?

Screen Shot 2023-01-03 at 10 09 41

pirj: We seem to be in the opposite situation when an option hash passed as a positional argument to an expectation suddenly matches with a kwargs hash.

I suppose that the problem is here:
```ruby
            if Hash === expected_args.last && Hash === actual_args.last
              if !Hash.ruby2_keywords_hash?(actual_args.last) && Hash.ruby2_keywords_hash?(expected_args.last)
                return false
```
In our scenario, `Hash.ruby2_keywords_hash?(actual_args.last)` is `true`, while the other one is `false`.
They don't match, but we don't return `false` like we do e.g. here:
```ruby
            (Hash.ruby2_keywords_hash?(expected_hash) != Hash.ruby2_keywords_hash?(actual_hash))
```

pirj: I suggest changing this:
```diff
-              if !Hash.ruby2_keywords_hash?(actual_args.last) && Hash.ruby2_keywords_hash?(expected_args.last)
+              if Hash.ruby2_keywords_hash?(actual_args.last) != Hash.ruby2_keywords_hash?(expected_args.last)
```

This is causing two more specs to fail, at least on Ruby 3.2:
```ruby
            it "matches against a hash submitted as keyword arguments and received as positional argument (in both Ruby 2 and Ruby 3)" do
              expect(dbl).to receive(:kw_args_method).with(1, {:required_arg => 2, :optional_arg => 3})
              dbl.kw_args_method(1, :required_arg => 2, :optional_arg => 3)
            end
```
```ruby
        it "matches against a hash submitted as keyword arguments a and received as a positional argument (in both Ruby 2 and Ruby 3)" do
          opts = {:a => "a", :b => "b"}
          expect(a_double).to receive(:random_call).with(opts)
          a_double.random_call(:a => "a", :b => "b")
        end
```

But if we'd make this comparison strict, I agree that it should fail.

Do you guys agree to make such a change, and see if it affects other Rubies?

pirj: Now this is where I got stuck:
```ruby
               opts = {:a => "a", :b => "b"}
               expect(a_double).to receive(:random_call).with(opts)
               a_double.random_call(:a => "a", :b => "b")
```
fails with
```
     Failure/Error: a_double.random_call(:a => "a", :b => "b")
       #<Double (anonymous)> received unexpected message :random_call with ({:a=>"a", :b=>"b"})
```
but when I want to expect that failure:
```ruby
             expect do
               opts = {:a => "a", :b => "b"}
               expect(a_double).to receive(:random_call).with(opts)
               a_double.random_call(:a => "a", :b => "b")
               # RSpec::Mocks.space.verify_all
             end.to fail_with(/received unexpected message :random_call with \({:a=>"a", :b=>"b"}\)/)
```
it:
```
     Failure/Error: expect(a_double).to receive(:random_call).with(opts)

       (Double (anonymous)).random_call({:a=>"a", :b=>"b"})
           expected: 1 time with arguments: ({:a=>"a", :b=>"b"})
           received: 0 times
```
and adding `verify_all`/`RSpec::Mocks.verify` doesn't help.

What's even more peculiar is that if I expect a failure message that doesn't match:
```ruby
             expect do
               opts = {:a => "a", :b => "b"}
               expect(a_double).to receive(:random_call).with(opts)
               a_double.random_call(:a => "a", :b => "b")
               RSpec::Mocks.verify
             end.to fail_with(/foo/)
```
it seems to make sense:
```ruby
     Failure/Error:
       expect do
         opts = {:a => "a", :b => "b"}
         expect(a_double).to receive(:random_call).with(opts)
         a_double.random_call(:a => "a", :b => "b")
       end.to fail_with(/foo/)

       expected RSpec::Mocks::MockExpectationError with message matching /foo/, got #<RSpec::Mocks::MockExpectationError: #<Double (anonymous)> received unexpected message :random_call with ({:a=>"a", :b=>"b"})> with backtrace:
```
@JonRowe I ended up with:
```ruby
             expect do
               opts = {:a => "a", :b => "b"}
               expect(a_double).to receive(:random_call).with(opts)
               a_double.random_call(:a => "a", :b => "b")
             end.to fail_with do |failure|
               expect(failure.message).to match(/received unexpected message :random_call with \(\{:a=>"a", :b=>"b"\}\)/)
               reset_all
             end
```
and double-checked that this example fails if I pass a different failure message to `match`. Do you have a better suggestion?

pirj:
I dared pushing a commit that makes the kwargs vs positional options hash check more strict.

TODO:
 - [ ] `spec/rspec/mocks/matchers/receive_spec.rb` fails for some weird reason similar to the one described above in `spec/rspec/mocks/argument_matchers_spec.rb`, but I couldn't quickly find a workaround the same way I did for `spec/rspec/mocks/argument_matchers_spec.rb`
 - [ ] failure message is suboptimal. We need some refactoring so that for a `MethodDouble` the Proxy called `raise_unexpected_message_args_error` instead of `raise_unexpected_message_error` the same way it does in other cases when there's a message expectation, but arguments don't match

pirj:
@paddor May I kindly ask you to repeat your test after the last commit?

pirj:
https://github.com/rspec/rspec-core/pull/2993 is needed to make the CI green(er) for `rspec-core` sub-build.

paddor:
@pirj It mostly passes. Some `#receive_message_chain`-lines with kwargs fail:
```
#<Double (anonymous)> received unexpected message :store_event with ({:message=>"core.system_events.force_variables", :object=>{:note=>nil, :release_at=>nil, :variable=>{:code=>nil, :desc_i18n=>nil, :label=>nil, :label_i18n=>nil}}, :username=>nil})
```

Note that the call-site correctly specified these arguments as keywords, not as a Hash.

If I refactor them to use `#receive` with intermediate spy-objects, the whole testsuite passes.

eregon:
@pirj I think it would be safer to revert that strict check (i.e., stricter than Ruby itself is), see https://github.com/rspec/rspec-mocks/pull/1514#discussion_r1060563272. Otherwise it sounds like a huge backward-incompatible change for something which doesn't really gain anything.

spec/rspec/mocks/matchers/receive_spec.rb
eregon: The problem is the comment above:
> # if both arguments end with Hashes, and if one is a keyword hash and the other is not, they don't match

is super confusing. That should be updated to something like:

```ruby
# if both arguments end with Hashes, and if #with was called with keywords but the method was called without keywords, they don't match
# if #with is called without keywords, e.g., with({a: 1}), then it's fine to call it with either foo(a: 1) or foo({a: 1}) because it's as if the method was: def foo(options).
```

pirj: But this is only true for methods that accept positional parameters, right? E.g. for:
```ruby
class A
  def foo(bar:)
  end
end

a = A.new
expect(a).to receive(:foo).with({bar: 1})
a.foo({bar: 1})
```

If we follow
> if `#with` is called without keywords, e.g., `with({a: 1})`, then it's fine to call it with either `foo(a: 1)` or `foo({a: 1})` because it's as if the method was: `def foo(options)`.

The spec will pass, but the real-life code:
```ruby
a = A.new
a.foo({bar: 1})
```
would blow up with:
```
ArgumentError: wrong number of arguments (given 1, expected 0; required keyword: bar)
```

that means we would be concealing the error.

Is my understanding correct, @eregon ?

pirj:
@eregon The check is certainly stricter now for `def foo(options = {})` or `def foo(options)`.

But it's not stricter than Ruby's for `def foo(**options)` and `def foo(bar:)`.

To make a distinction between those two cases, we would have to make `ArgumentListMatcher` aware of the method signature.

All those cases are different `instance.method(:foo).parameters`:
```ruby
Class.new { def foo(options); end }.new.method(:foo).parameters # => [[:req, :options]]
Class.new { def foo(options = {}); end }.new.method(:foo).parameters # => [[:opt, :options]]
Class.new { def foo(**options); end }.new.method(:foo).parameters # => [[:keyrest, :options]]
Class.new { def foo(bar:); end }.new.method(:foo).parameters # => [[:keyreq, :bar]]
```

What is making matter worse is that Ruby was doing weird things with extracting some arguments from the options hash, see e.g. https://github.com/rspec/rspec-support/pull/537

spec/rspec/mocks/matchers/receive_spec.rb
eregon: It's a mock, so the method might not exist, right?
From `with({a: 1})` that doesn't indicate the method takes keyword arguments, hence it can be called with whatever (passing `a: 1` or `{a: 1}` shouldn't make any difference).

If the method does exist like in that case, I think maybe RSpec could warn that you call `with` without kwargs, but the method accepts kwargs.
Or, you could actually check whether the existing method accepts kwargs or not for the kwargs-vs-positional check, but then what happens if the method does not exist?

Wouldn't this specific example fail anyway because it actually calls the original method?
And if it doesn't call the original method, isn't it wrong to look at the original method?

eregon: Also this example wouldn't be caught by your change, right? (no ruby2_keyword flag on both sides)
So it's a bug of writing `.with({bar: 1})` which means "expect no kwargs" when the real method only accepts kwargs. Hence maybe that's worth a warning.

rickychilcott:
Running this branch fixes all of my mocking issues.  Thank you!

This is more of an rspec-rails issue, but I'm noticing a `have_enqueued_job` matcher is no longer finding the job.

```ruby
expect {
    get :create, params: {onboarding_signup_id: onboarding_signup_id}
  }.to have_enqueued_job(Sendgrid::ListSignupJob).with(
    email: user.email, list: OnboardingSignupsController::CURRENT_ONBOARDING_LIST_NAME
  )

  expected to enqueue exactly 1 jobs, with [{:email=>"clair_mcdermott@example.net", :list=>"PROD Onboarding v1"}], but enqueued 0
  Queued jobs:
    Sendgrid::ListSignupJob job with [{:email=>"clair_mcdermott@example.net", :list=>"PROD Onboarding v1"}], on queue default
```

^^^ passes in Ruby 3.1

Mange:
I've tested this branch but expectations on any keyword argument seem to not work.

```ruby
require "spec_helper"

class Ex
  def self.foo(a:)
    puts "Called foo(a: #{a.inspect})"
    true
  end
end

RSpec.describe "example" do
  it "finds no call on single call" do
    allow(Ex).to receive(:foo).and_call_original

    Ex.foo(a: 1)

    #  Failure/Error: expect(Ex).to have_received(:foo).with(a: 1)
    #    (Ex (class)).foo({:a=>1})
    #        expected: 1 time with arguments: ({:a=>1})
    #        received: 0 times
    expect(Ex).to have_received(:foo).with(a: 1)
  end

  it "finds other call on two calls, but still fails" do
    allow(Ex).to receive(:foo).and_call_original

    Ex.foo(a: 1)
    Ex.foo(a: 2)

    # #<Ex (class)> received :foo with unexpected arguments
    #   expected: ({:a=>1}) (options hash)
    #        got: ({:a=>2}) (keyword arguments)
    #   Diff:
    #   @@ -1 +1 @@
    #   -[{:a=>1}]
    #   +[{:a=>2}]
    expect(Ex).to have_received(:foo).with(a: 1)
    expect(Ex).to have_received(:foo).with(a: 2)

    # Note that just expecting the second call will still fail, but with an
    # inverted error message; it will show the first call.
  end
end
```

<details>
<summary>Output</summary>

```
Randomized with seed 60000

example
Called foo(a: 1)
  finds no call on single call (FAILED - 1)
Called foo(a: 1)
Called foo(a: 2)
  finds other call on two calls, but still fails (FAILED - 2)

Failures:

  1) example finds no call on single call
     Failure/Error: expect(Ex).to have_received(:foo).with(a: 1)

       (Ex (class)).foo({:a=>1})
           expected: 1 time with arguments: ({:a=>1})
           received: 0 times
     # ./spec/broken_spec.rb:20:in `block (2 levels) in <top (required)>'

  2) example finds other call on two calls, but still fails
     Failure/Error: expect(Ex).to have_received(:foo).with(a: 1)

       #<Ex (class)> received :foo with unexpected arguments
         expected: ({:a=>1}) (options hash)
              got: ({:a=>2}) (keyword arguments)
       Diff:
       @@ -1 +1 @@
       -[{:a=>1}]
       +[{:a=>2}]

     # ./spec/broken_spec.rb:36:in `block (2 levels) in <top (required)>'

Finished in 0.01183 seconds (files took 0.26323 seconds to load)
2 examples, 2 failures

Failed examples:

rspec ./spec/broken_spec.rb:11 # example finds no call on single call
rspec ./spec/broken_spec.rb:23 # example finds other call on two calls, but still fails
```
</details>

<details>
<summary>Gem versions (commits)</summary>

```
GIT
  remote: https://github.com/rspec/rspec-core.git
  revision: 522b7727d02d9648c090b56fa68bbdc18a21c04d
  branch: main
  specs:
    rspec-core (3.13.0.pre)
      rspec-support (= 3.13.0.pre)

GIT
  remote: https://github.com/rspec/rspec-expectations.git
  revision: b1fa2e620b03ed05c737db10b2727a0706eca7d3
  branch: main
  specs:
    rspec-expectations (3.13.0.pre)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (= 3.13.0.pre)

GIT
  remote: https://github.com/rspec/rspec-mocks.git
  revision: d8a213c26c31bc940f5060145b85131cbcf43715
  branch: ruby-3.2
  specs:
    rspec-mocks (3.13.0.pre)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (= 3.13.0.pre)

GIT
  remote: https://github.com/rspec/rspec-rails.git
  revision: c60ff7907559653cd9d1ec1a6113bf86c9359fab
  branch: main
  specs:
    rspec-rails (6.1.0.pre)
      actionpack (>= 6.1)
      activesupport (>= 6.1)
      railties (>= 6.1)
      rspec-core (= 3.13.0.pre)
      rspec-expectations (= 3.13.0.pre)
      rspec-mocks (= 3.13.0.pre)
      rspec-support (= 3.13.0.pre)
```

</details>

spec/rspec/mocks/matchers/receive_spec.rb
pirj: >  this example wouldn't be caught by your change

Fair enough. The rest makes total sense, too.

paddor:
Now it works with and without the refactorings in my testsuite. 👍

.github/workflows/ci.yml
```diff
@@ -38,8 +38,9 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - '3.1'
-          - '3.0'
+          - 3.2
+          - 3.1
+          - 3.0
```
eregon, line 43: This will be parsed as `3` and will become `3.2`: https://github.com/ruby/setup-ruby#matrix-of-ruby-versions

eregon (approved):
Looks good to me

pirj:
@paddor Just to make sure, do your `receive_message_chain` work, too?

pirj (approved):
Looks good!

pirj:
@JonRowe I intend to:
 - merge this
 - fix the [ReentrantMutex issue](https://github.com/rspec/rspec-support/pull/553) to fix Ruby 3.2 sub-builds
 - handle a coordinated update of buildfiles from `rspec-dev` to add Ruby 3.2 to ci

JonRowe:
Thanks for tackling this, would have been nice to wait for my review but as this ended up as a simple one line lib change it's ok, good work on the updated comment too!

JonRowe:
Released in 3.12.2

JonRowe:
I added the 3.2 builds myself coordinating with 3-12-maintenance, with a temporary suppression of the broken spec, I still think it can be fixed rather than skipped like this.

Also I've now disabled rebase merges on our repos as they've been used accidentally again and it makes cherry picking to maintenance branches a nightmare, we want single commits for PRs (merge or squash only).

PR #1520, opened 2023-01-10
Don't convert keywords to hash in #send for verifying doubles on Ruby 3

pirj:
The current re-implementation of `send` on verifying doubles uses `*args` by itself. Everything works fine on `main` for Ruby 2 or lower, but on Ruby 3 this converts keyword arguments to a Hash element inside the `args` array, which then gets passed to `__send__` and then `super` and (correctly) causes a failure when the method gets a Hash instead of keywords.

Without this patch, the new test passes on Ruby 2.7.6 and fails on Ruby 3.1.2:
```
Failures:

  1) Expected argument verification (when `#with` is called) when doubling a loaded class for a method with keyword args when using `send` matches against keyword arguments
     Failure/Error: dbl.send(:m, :k => 1)
     
       #<InstanceDouble(#<Class:0x00007fbd16cf1ba8>) (anonymous)> received :m with unexpected arguments
         expected: ({:k=>1}) (keyword arguments)
              got: ({:k=>1}) (options hash)
       Diff:
```

With this patch, all tests pass on all Ruby versions.

Fixes #1485

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,11 @@
 ### Development
 [Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.12.1...main)
 
+Bug Fixes:
+
+* Fix keyword delegation in `send` for verifying doubles on Ruby 3.
+  (Charlie Honig, #1485)
+
 ### 3.12.2 / 2023-01-07
 [Full Changelog](http://github.com/rspec/rspec-mocks/compare/v3.12.1...v3.12.2)
 
--- a/lib/rspec/mocks/verifying_double.rb
+++ b/lib/rspec/mocks/verifying_double.rb
@@ -42,11 +42,13 @@ module RSpec
       ensure
         @__sending_message = nil
       end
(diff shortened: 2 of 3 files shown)


Laykou:
Do you think this could be released as new version? 

Currently I had to load from source:

```ruby
gem 'rspec-rails', '6.1.0.pre', github: 'rspec/rspec-rails', branch: 'main'
gem 'rspec', '3.13.0.pre', github: 'rspec/rspec', branch: 'main'
gem 'rspec-core', '3.13.0.pre', github: 'rspec/rspec-core', branch: 'main'
gem 'rspec-mocks', github: 'rspec/rspec-mocks', branch: 'main'
gem 'rspec-expectations', github: 'rspec/rspec-expectations', branch: 'main'
gem 'rspec-support', github: 'rspec/rspec-support', branch: 'main'
```

to fix the issue with mocking of method with keyword arguments

JonRowe:
Released as `3.12.3`

Laykou:
Thank you! This resolved the "keyword" argument mock issue in Ruby 3.2 for me. Yay! 🥳

Issue #1527, opened 2023-03-01
array_excluding matcher?


sirwolfgang:
Is there any technical reason there is not an `array_excluding` argument matcher? There's hash excluding right below it.

https://github.com/rspec/rspec-mocks/blob/main/lib/rspec/mocks/argument_matchers.rb#L74-L93

sirwolfgang:
This seems to work fine; Just weird to me that this has never come up before?
```
module RSpec
  module Mocks
    module ArgumentMatchers
      # Matches an array that includes the specified items at least once.
      # Ignores duplicates and additional values
      #
      # @example
      #   expect(object).to receive(:message).with(array_excluding(1,2,3))
      #   expect(object).to receive(:message).with(array_excluding([1,2,3]))
      def array_excluding(*args)
        actually_an_array = Array === args.first && args.count == 1 ? args.first : args
        ArrayExcludingMatcher.new(actually_an_array)
      end

      # @private
      class ArrayExcludingMatcher
        def initialize(expected)
          @expected = expected
        end

        def ===(actual)
          actual = actual.uniq
          @expected.uniq.none? do |expected_element|
            actual.any? do |actual_element|
              RSpec::Support::FuzzyMatcher.values_match?(expected_element, actual_element)
            end
          end
        rescue NoMethodError
          false
        end

        def description
          "array_excluding(#{formatted_expected_values})"
        end

      private

        def formatted_expected_values
          @expected.map do |x|
            RSpec::Support.rspec_description_for_object(x)
          end.join(", ")
        end
      end
    end
  end
end
```

JonRowe:
Closing to continue discussion in #1528

PR #1528, opened 2023-03-01
Add array_excluding matcher

sirwolfgang:
Closes https://github.com/rspec/rspec-mocks/issues/1527

- Add array_excluding matcher
- Reorder hash_excluding to match the order of the specs and classes

### DIFF
--- a/lib/rspec/mocks/argument_matchers.rb
+++ b/lib/rspec/mocks/argument_matchers.rb
@@ -71,6 +71,16 @@ module RSpec
         HashIncludingMatcher.new(ArgumentMatchers.anythingize_lonely_keys(*args))
       end
 
+      # Matches a hash that doesn't include the specified key(s) or key/value.
+      #
+      # @example
+      #   expect(object).to receive(:message).with(hash_excluding(:key => val))
+      #   expect(object).to receive(:message).with(hash_excluding(:key))
+      #   expect(object).to receive(:message).with(hash_excluding(:key, :key2 => :val2))
+      def hash_excluding(*args)
+        HashExcludingMatcher.new(ArgumentMatchers.anythingize_lonely_keys(*args))
+      end
+
       # Matches an array that includes the specified items at least once.
       # Ignores duplicates and additional values
       #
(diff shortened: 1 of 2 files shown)


lib/rspec/mocks/argument_matchers.rb
```diff
@@ -258,6 +268,36 @@ def formatted_expected_values
...
+          @unexpected = unexpected
+        end
+
+        def ===(actual)
+          actual = actual.uniq
+          @unexpected.uniq.none? do |unexpected_element|
+            actual.any? do |actual_element|
+              RSpec::Support::FuzzyMatcher.values_match?(unexpected_element, actual_element)
+            end
+          end
```
JonRowe, line 283: This implementation will be slow to pass, as it has to check the actual array repeadedly for each unexpected value, given that I'd expect the unexpected values to be shorter (although you could actually gate this check on the size and flip flop) I'd invert this to loop over actual once and check each value against all unexpected...

I'd also consider shortcutting with a plain intersection check, if that fails you can fail the example quickly.

JonRowe (changes requested):

JonRowe:
:wave: Thanks for adding this, I don't see why this shouldn't be present except for some concerns about performance of the implementation.

spec/rspec/mocks/argument_matchers_spec.rb
```diff
@@ -295,6 +300,34 @@ module Mocks
...
+      describe "array_excluding" do
+        it "matches array with array_excluding different array" do
+          expect(a_double).to receive(:random_call).with(array_excluding(3, 4))
+          a_double.random_call([1, 2])
+        end
+
+        it "fails array_excluding when args aren't array", :reset => true do
+          expect(a_double).to receive(:msg).with(array_excluding(1, 2, 3))
+          expect {
+            a_double.msg(1, 2, 3)
```
pirj, line 312: It's not exactly clear why this expectation would fail. We expect the array to exclude 1,2,3, but this is what we call it with.
To avoid confusion, do you mind making those 3,4 and 1,2 like in the example above?

spec/rspec/mocks/argument_matchers_spec.rb
```diff
@@ -295,6 +300,34 @@ module Mocks
...
+          }.to fail_including "expected: (array_excluding(1, 2, 3))"
+        end
+
+        it "fails array_excluding when arg doesn't contain all elements", :reset => true do
```
pirj, line 316: This sounds more like "when the arg contains some elements", doesn't it?

spec/rspec/mocks/argument_matchers_spec.rb
```diff
@@ -295,6 +300,34 @@ module Mocks
...
+          expect(a_double).to receive(:msg).with(array_excluding(1, 2, 3))
+          expect {
+            a_double.msg([1, 2])
+          }.to fail_including "expected: (array_excluding(1, 2, 3))"
+        end
+
+        it "fails array_excluding when using the fuzzy matcher", :reset => true  do
```
pirj, line 323: I believe this needs a balancing example where it would match, e.g. `array_excluding(a_value > 2)`.

spec/rspec/mocks/argument_matchers_spec.rb
sirwolfgang: I changed the message of this to be more clear in what was being tested

JonRowe:
Released in 3.13.0 apologies it has taken so long.

Issue #1530, opened 2023-03-06
Rails 7.1 breaks receive().with


palkan:
### Your environment
* Ruby version: 3.2
* rspec-mocks version: 3.12.2
* Rails version: [main](https://github.com/rails/rails/commit/20d87500a4568b175f1e06796572154a5a8d283d).

### Steps to reproduce

In a Rails application (i.e., with Rails loaded), define a method expectation:

```ruby
mock = instance_double("Hash")
allow(mock).to receive(:key?).with(:x) { 1 }

expect(mock.key?(:x)).to eq 1
```

### Expected behavior

Pass.

### Actual behavior

```
ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # ../rails/activesupport/lib/active_support/core_ext/object/with.rb:24:in `with'
```

### What happened?

Rails monkey-patched `Object#with` https://github.com/rails/rails/pull/46681 /cc @byroot 

RSpec `receive` matcher defines chain methods only for undefined methods: https://github.com/rspec/rspec-mocks/blob/4a1032bf93fe8b47c9e3416413e2b05d604d109f/lib/rspec/mocks/matchers/receive.rb#L58-L59

Since `#with` is defined on `Object`, the `with` _modifier_ is no longer available.

nobu:
Why does `MessageExpectation` need to b e a subclass of `Object`, not `BasicObject`?

byroot:
Even then, why would it not redefine existing methods? I'll git blame to see if I can find the reason for this.

byroot:
Ah, I think the intent is to not redefine methods directly defined in `Receive`, since we're copying all the matchers. `method_defined?(name, false)` would fix it I think.

byroot:
I think I have a fix for it, but GitHub seems to be having issues, I get an error message when trying to create the PR....

https://github.com/rspec/rspec-mocks/compare/rspec:main...casperisfine:receiver-with?expand=1

pirj:
@byroot I’ve opened one, does it work for you?

byroot:
Sure, thank you.

PR #1531, opened 2023-03-07
Receive: only avoid overriding own methods

pirj:
Fix: https://github.com/rspec/rspec-mocks/issues/1530

My understanding of this `next if method_defined?` check that there since this class was added in 08ec2e80325a5959ac13b113a0175d28aca352a8 is that it's to prevent accidentally overriding Matcher or own methods.

If my understanding is correct, then we should ignore methods inherited from `Object`.

### DIFF
--- a/lib/rspec/mocks/matchers/receive.rb
+++ b/lib/rspec/mocks/matchers/receive.rb
@@ -55,8 +55,9 @@ module RSpec
           setup_any_instance_method_substitute(subject, :stub, block)
         end
 
+        own_methods = (instance_methods - superclass.instance_methods)
         MessageExpectation.public_instance_methods(false).each do |method|
-          next if method_defined?(method)
+          next if own_methods.include?(method)
 
           define_method(method) do |*args, &block|
             @recorded_customizations << ExpectationCustomization.new(method, args, block)


byroot:
The JRuby failure seem unrelated.

lib/rspec/mocks/matchers/receive.rb
```diff
@@ -55,8 +55,9 @@ def setup_any_instance_allowance(subject, &block)
           setup_any_instance_method_substitute(subject, :stub, block)
         end
 
+        own_methods = (instance_methods - superclass.instance_methods)
```
byroot, line 58: Note, I used `superclass.instance_methods` here, but depending on what the exact intent was, `Object.instance_methods` may make more sense.

lib/rspec/mocks/matchers/receive.rb
```diff
@@ -55,8 +55,9 @@ def setup_any_instance_allowance(subject, &block)
...
         MessageExpectation.public_instance_methods(false).each do |method|
-          next if method_defined?(method)
+          next if own_methods.include?(method)
```
pirj, line 60: Wondering if this check is even needed.
The two methods it skipped are `to_s` and `inspe﻿ct` which are defined in `MessageExpectation` and we want to avoid pulling that into `Receive`.
What if we just explicitly state that?
```suggestion
          next if method == :to_s || method == :inspect
```
```

**Review by pirj — COMMENTED** [review]
Do you think it would be possible to write a test this case when some of the method that `Receive` pulls from `MessageExpectation` are still defined if someone monkey patches `Object` with the same ones?

We have this `in_sub_process_if_possible` helper that would run in a fork and would not affect other specs, e.g.:

```ruby
      in_sub_process_if_possible do
        require 'set'

        set = Set.new([1])
        expect {
          set << 2
        }.to change { set }.from([1].to_set).to([2, 1].to_set)
      end
```

byroot:
> Do you think it would be possible to write a test this case when some of the method that Receive pulls from MessageExpectation are still defined if someone monkey patches Object with the same ones?

Hum, not sure how we could do this, that code is only executed once, so it's very hard to unit test it, and even if we refactored it out we'd have to patch object temporarily to simulate the issue.

I get the want for a test case, but it seems very impractical here.

lib/rspec/mocks/matchers/receive.rb
byroot: Up to you, I don't know rspec-mock much and I rarely use it. 

But yes, that would make sense to me, it would be a much more predictable behavior.

JonRowe: The intent is to be future proof.

byroot:
Any blocker on this? I don't use RSpec myself, but I assume it my be quite blocking for RSpec users who track Rails edge.

JonRowe:
It needs a test, I've added one in #1534 but some of the builds aren't passing yet

byroot:
Thanks, I didn't notice that other PR.

JonRowe:
It _was_ intended to be quick addition then merge but the build gods had other ideas

JonRowe:
Merged in [7c10871](https://github.com/rspec/rspec-mocks/commit/7c10871f60a3a5e402b4852da616dae8290e338a)

JonRowe:
Released in 3.12.4 thank you!

PR #1534, opened 2023-03-09
Receive: only avoid overriding own methods (with added spec)

JonRowe:
This is #1531 with an integration test

### DIFF
--- a/lib/rspec/mocks/matchers/receive.rb
+++ b/lib/rspec/mocks/matchers/receive.rb
@@ -55,8 +55,9 @@ module RSpec
           setup_any_instance_method_substitute(subject, :stub, block)
         end
 
+        own_methods = (instance_methods - superclass.instance_methods)
         MessageExpectation.public_instance_methods(false).each do |method|
-          next if method_defined?(method)
+          next if own_methods.include?(method)
 
           define_method(method) do |*args, &block|
             @recorded_customizations << ExpectationCustomization.new(method, args, block)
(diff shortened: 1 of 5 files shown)


lib/rspec/mocks/matchers/receive.rb
```diff
@@ -55,8 +55,9 @@ def setup_any_instance_allowance(subject, &block)
           setup_any_instance_method_substitute(subject, :stub, block)
         end
 
+        own_methods = (instance_methods - superclass.instance_methods)
```
pirj, line 58:
```suggestion
```

lib/rspec/mocks/matchers/receive.rb
```diff
@@ -55,8 +55,9 @@ def setup_any_instance_allowance(subject, &block)
...
         MessageExpectation.public_instance_methods(false).each do |method|
-          next if method_defined?(method)
+          next if own_methods.include?(method)
```
pirj, line 60:
```suggestion
          next if method == :to_s || method == :inspect
```
Just because those are the only ones that overlap between `Receive` and `MessageExpectation`. We want to pull the rest with no regards if it's already defined in `Receive`.

JonRowe: The intent is to be future proof, and not overwrite conflicting methods.

JonRowe:
Released in 3.12.4

PR #1540, opened 2023-03-28
fix stubbing with allow_any_instance_of when #with is monkey patched by Rails

lsylvester:
When stubbing with `allow_any_instance_of` and using the Rails monkey patch `with`, the `with` method is define on `FluentInterfaceProxy` so it is not being correctly delegated to the target.

By using BasicObject, FluentInterfaceProxy doesn't inherit any monkey patches that Rails or other libraries make on Object and so is better able to delegate the method calls.

### DIFF
--- a/lib/rspec/mocks/any_instance/proxy.rb
+++ b/lib/rspec/mocks/any_instance/proxy.rb
@@ -83,6 +83,14 @@ module RSpec
         end
       end
 
+      unless defined?(BasicObject)
+        class BasicObject
+          (instance_methods.map(&:to_sym) - [:__send__, :!, :instance_eval, :==, :instance_exec, :!=, :equal?, :__id__, :__binding__, :object_id]).each do |method|
+            undef_method method
+          end
+        end
+      end
+
       # @private
       # Delegates messages to each of the given targets in order to
       # provide the fluent interface that is available off of message
@@ -91,13 +99,7 @@ module RSpec
       # `targets` will typically contain 1 of the `AnyInstance::Recorder`
       # return values and N `MessageExpectation` instances (one per instance
       # of the `any_instance` klass).
-      class FluentInterfaceProxy < defined?(BasicObject) ? BasicObject : Object
-        if superclass == ::Object
-          (instance_methods.map(&:to_sym) - [:__send__, :!, :instance_eval, :==, :instance_exec, :!=, :equal?, :__id__, :__binding__, :object_id]).each do |method|
-            undef_method method
-          end
-        end
-
+      class FluentInterfaceProxy < BasicObject
         def initialize(targets)
           @targets = targets
         end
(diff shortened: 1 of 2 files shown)


lib/rspec/mocks/any_instance/proxy.rb
```diff
@@ -91,12 +91,12 @@ def perform_proxying(method_name, args, block, &target_proxy_block)
...
       # return values and N `MessageExpectation` instances (one per instance
       # of the `any_instance` klass).
-      class FluentInterfaceProxy
+      class FluentInterfaceProxy < BasicObject
         def initialize(targets)
           @targets = targets
         end
 
-        if RUBY_VERSION.to_f > 1.8
+        if ::RUBY_VERSION.to_f > 1.8
```
JonRowe, line 99: I'm on the fence about this, its not ambiguous so it just creates churn

spec/integration/rails_support_spec.rb
```diff
@@ -33,4 +33,31 @@ def with
     expect(last_command_started).to have_output(/0 failures/)
     expect(last_command_started).to have_exit_status(0)
   end
+
+  it "works mocking any instance when Rails has monkey patched #with" do
```
JonRowe, line 37:
```suggestion
  it "works mocking any instance when Rails has monkey patched #with", :if => RUBY_VERSION.to_f > 1.8 do
```

JonRowe (changes requested):
:wave: Thanks for this, needs a change to make CI pass, I've suggested one way (just ignoring Rubies lacking BasicObject) but an alternative would be to do something similar to #1534 to find methods to remove, or manually proxy

lib/rspec/mocks/any_instance/proxy.rb
lsylvester: If we have subclassed from BasicObject, then we do not have access to any constants (as root level constants are defined on object). 

This is not a style change. Without this change I get error:

```
NameError:
  uninitialized constant RSpec::Mocks::AnyInstance::FluentInterfaceProxy::RUBY_VERSION
```

lib/rspec/mocks/any_instance/proxy.rb
```diff
@@ -91,12 +91,18 @@ def perform_proxying(method_name, args, block, &target_proxy_block)
       # `targets` will typically contain 1 of the `AnyInstance::Recorder`
       # return values and N `MessageExpectation` instances (one per instance
       # of the `any_instance` klass).
-      class FluentInterfaceProxy
+      class FluentInterfaceProxy < defined?(BasicObject) ? BasicObject : Object
+        if superclass == ::Object
+          (instance_methods.map(&:to_sym) - [:__send__, :!, :instance_eval, :==, :instance_exec, :!=, :equal?, :__id__, :__binding__, :object_id]).each do |method|
```
lsylvester, line 96: These methods came from running `BasicObject.instance_methods` in modern ruby and adding in `object_id` (as it being missing caused test failures).

JonRowe: Can you add a comment explain the source of the methods please.

JonRowe (approved):

JonRowe:
Released in 3.12.5

PR #1552, opened 2023-07-05
Fix and_call_original for Ruby 3.2

igor-drozdov:
When the original method isn't present, a proc that sends method_missing is returned. This block must support ruby2_keywords in order to work correctly with Ruby 3.2.

Otherwise, the following error is raised

```ruby
     Failure/Error:
       def existing_method_with_args(arg, kwarg:)
         [arg, kwarg]
       end

     ArgumentError:
       wrong number of arguments (given 2, expected 1; required keyword: kwarg)
```

### DIFF
--- a/spec/rspec/mocks/and_call_original_spec.rb
+++ b/spec/rspec/mocks/and_call_original_spec.rb
@@ -243,6 +243,40 @@ RSpec.describe "and_call_original" do
       CODE
     end
 
+    if RSpec::Support::RubyFeatures.required_kw_args_supported?
+      binding.eval(<<-RUBY, __FILE__, __LINE__)
+      context 'on an object with a method propagated by method_missing' do
+        before do
+          klass.class_exec do
+            private
+
+            def call_method_with_kwarg(arg, kwarg:)
+              [arg, kwarg]
+            end
+
+            def method_missing(name, *args, **kwargs)
+              if name.to_s == "method_with_kwarg"
+                call_method_with_kwarg(*args, **kwargs)
+              else
+                super
+              end
+            end
+          end
+        end
+
+        it 'works for the method propagated by method missing' do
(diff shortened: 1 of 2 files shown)


spec/rspec/mocks/and_call_original_spec.rb
```diff
@@ -248,9 +248,15 @@ def instance.foo(bar: nil); bar; end
         klass.class_exec do
           private
 
-          def method_missing(name, *args)
+          def existing_method_with_args(arg, kwarg:)
+            [arg, kwarg]
+          end
+
+          def method_missing(name, *args, **kwargs)
```
JonRowe, line 255: You'll need to eval these from strings conditionally to pass our build

spec/rspec/mocks/and_call_original_spec.rb
```diff
@@ -248,9 +248,19 @@ def instance.foo(bar: nil); bar; end
...
+              [arg, kwarg]
+            end
+            RUBY
+          end
+
+          def method_missing(name, *args, **kwargs)
             if name.to_s == "greet_jack"
               "Hello, jack"
+            elsif name.to_s == "method_with_kwarg"
+              call_method_with_kwarg(*args, **kwargs)
```
JonRowe, line 263: This will also need to be wrapped to pass our build, until we release RSpec 4 we support pre keyword rubies (sorry)

igor-drozdov: @JonRowe ah, got it, I've changed the code, and now all the new examples are evaluated conditionally 👍

igor-drozdov: Ruby 2.0 failed because the kwarg doesn't have a default value, I changed the condition to run only when required-kwarg are available. Hope it will pass now

JonRowe:
Thanks!

JonRowe:
Released in 3.12.6

PR #1567, opened 2024-01-25
Remove mentions of legacy methods

pirj:
Follow-up to #1365

### DIFF
--- a/lib/rspec/mocks/example_methods.rb
+++ b/lib/rspec/mocks/example_methods.rb
@@ -326,9 +326,8 @@ module RSpec
       end
 
       # Used to specify a message that you expect or allow an object
-      # to receive. The object returned by `receive` supports the same
-      # fluent interface that `should_receive` and `stub` have always
-      # supported, allowing you to constrain the arguments or number of
+      # to receive. The object returned by `receive` supports a fluent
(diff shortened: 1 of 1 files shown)


lib/rspec/mocks/example_methods.rb
```diff
@@ -326,9 +326,8 @@ def allow_any_instance_of(klass)
       end
 
       # Used to specify a message that you expect or allow an object
-      # to receive. The object returned by `receive` supports the same
-      # fluent interface that `should_receive` and `stub` have always
-      # supported, allowing you to constrain the arguments or number of
+      # to receive. The object returned by `receive` supports a fluent
+      # interface, allowing you to constrain the arguments or number of
```
JonRowe, line 330: This is fine, are you leaving this open to collect more?

pirj: No, the build was afwully ted and I decided to fix builds first, but it also was red on main

And I’ll to sync all changes from main yo 4-0-dev to incorporate all build fixes you recently did, and to align the two branches

JonRowe: As its just docs you can just merge this then fix it.

Issue #1568, opened 2024-01-25
`stub_const` doesn't clear `Class#subclasses`


GCorbel:
### Subject of the issue
When `stub_const` is used with a class that is a subclass of another, the class is still listed in subclasses of this parent after each spec.

### Your environment
* Ruby version: 3.1.4
* rspec-mocks version: 3.12.6

### Steps to reproduce

```
begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rspec", "3.12.0" # Activate the gem and version you are reporting the issue against.
  gem 'rspec-mocks', '3.12.6'
end

puts "Ruby version is: #{RUBY_VERSION}" # Ruby version is: 3.1.4

class Something; end

describe 'Test' do
  before(:each) do
    class A < Something; end
    stub_const('B', Class.new(Something))
  end

  it 'something' do
    puts Something.subclasses # => [B, A]
  end

  it 'something else' do
    puts Something.subclasses # => [B, B, A]
    # Only one occurence of B should be listed
  end
end
```

### Expected behavior

`Something.subclasses` always have to return `[B, A]` and be the same for each spec.

### Actual behavior

`Something.subclasses` still return every stubbed classes.

GCorbel:
It seems that doing a garbage collection after `RSpec::Mocks.teardown` fix the issue.

```
class Something; end

describe 'Test' do
  before(:each) do
    class A < Something; end
    stub_const('B', Class.new(Something))
  end

  after(:each) do
    RSpec::Mocks.teardown
    GC.start
  end

  it 'something' do
    pp Something.subclasses # => [B, A]
  end

  it 'something else' do
    pp Something.subclasses # => [B, A]
  end
end
```

GCorbel:
I tried on a Rails project and doing a garbage collection is not enough. Defined classes without `stub_const` are correctly removed and those that use `stub_const` aren't.

pirj:
What do you mean "correctly removed"?

```ruby
describe 'Test', order: :defined do
  context 'with A' do
    before(:each) do
      class A < Something; end
    end

    it 'something' do
      pp Something.subclasses # => [A]
    end
  end

  context 'presumably without A' do
    it 'something else' do
      pp Something.subclasses # => [A] 💥
    end
  end
end
```

`A` won't go anywhere.

[...]

PR #1570, opened 2024-02-20
Exclude stubbed classes from subclasses after teardown

GCorbel:
This fix https://github.com/rspec/rspec-mocks/issues/1568.

Stubbed classes are excluded from parent subclasses after each spec.

The original issue : 

```
begin
  require "bundler/inline"
rescue LoadError => e
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rspec", "3.12.0" # Activate the gem and version you are reporting the issue against.
  gem 'rspec-mocks', '3.12.6'
end

puts "Ruby version is: #{RUBY_VERSION}" # Ruby version is: 3.1.4

class Something; end

describe 'Test' do
  before(:each) do
    class A < Something; end
    stub_const('B', Class.new(Something))
  end

  it 'something' do
    puts Something.subclasses # => [B, A]
  end

  it 'something else' do
    puts Something.subclasses # => [B, B, A]
    # Only one occurence of B should be listed
  end
end
```

Now, `Something.subclasses` always return `[B, A]`.

### DIFF
--- a/lib/rspec/mocks.rb
+++ b/lib/rspec/mocks.rb
@@ -101,27 +101,6 @@ module RSpec
       end
     end
 
-    @excluded_subclasses = []
-
-    def self.excluded_subclasses
-      @excluded_subclasses.select(&:weakref_alive?).map do |ref|
-        ref.__getobj__
-      rescue RefError
-        nil
-      end.compact
-    end
-
-    def self.exclude_subclass(constant)
-      @excluded_subclasses << WeakRef.new(constant)
-    end
-
-    module ExcludeClassesFromSubclasses
-      def subclasses
-        super - RSpec::Mocks.excluded_subclasses
-      end
-    end
-    Class.prepend(ExcludeClassesFromSubclasses)
-
     class << self
       # @private
       attr_reader :space
@@ -134,10 +113,11 @@ module RSpec
 
     # To speed up boot time a bit, delay loading optional or rarely
     # used features until their first use.
-    autoload :AnyInstance,      "rspec/mocks/any_instance"
-    autoload :ExpectChain,      "rspec/mocks/message_chain"
-    autoload :StubChain,        "rspec/mocks/message_chain"
-    autoload :MarshalExtension, "rspec/mocks/marshal_extension"
+    autoload :AnyInstance,                          "rspec/mocks/any_instance"
+    autoload :ExpectChain,                          "rspec/mocks/message_chain"
+    autoload :StubChain,                            "rspec/mocks/message_chain"
+    autoload :MarshalExtension,                     "rspec/mocks/marshal_extension"
+    autoload :ExcludeStubbedClassesFromSubclasses,  "rspec/mocks/exclude_stubbed_classes_from_subclasses"
 
     # Namespace for mock-related matchers.
     module Matchers
--- a/lib/rspec/mocks.rb
+++ b/lib/rspec/mocks.rb
@@ -113,11 +113,11 @@ module RSpec
 
     # To speed up boot time a bit, delay loading optional or rarely
     # used features until their first use.
-    autoload :AnyInstance,                          "rspec/mocks/any_instance"
-    autoload :ExpectChain,                          "rspec/mocks/message_chain"
-    autoload :StubChain,                            "rspec/mocks/message_chain"
-    autoload :MarshalExtension,                     "rspec/mocks/marshal_extension"
-    autoload :ExcludeStubbedClassesFromSubclasses,  "rspec/mocks/exclude_stubbed_classes_from_subclasses"
+    autoload :AnyInstance, "rspec/mocks/any_instance"
+    autoload :ExpectChain, "rspec/mocks/message_chain"
+    autoload :StubChain, "rspec/mocks/message_chain"
+    autoload :MarshalExtension, "rspec/mocks/marshal_extension"
+    autoload :ExcludeStubbedClassesFromSubclasses, "rspec/mocks/exclude_stubbed_classes_from_subclasses"
 
     # Namespace for mock-related matchers.
     module Matchers
@@ -125,10 +125,10 @@ module RSpec
       # just a "tag" for rspec-mock matchers detection
       module Matcher; end
 
-      autoload :HaveReceived,        "rspec/mocks/matchers/have_received"
-      autoload :Receive,             "rspec/mocks/matchers/receive"
+      autoload :HaveReceived, "rspec/mocks/matchers/have_received"
+      autoload :Receive, "rspec/mocks/matchers/receive"
       autoload :ReceiveMessageChain, "rspec/mocks/matchers/receive_message_chain"
-      autoload :ReceiveMessages,     "rspec/mocks/matchers/receive_messages"
+      autoload :ReceiveMessages, "rspec/mocks/matchers/receive_messages"
     end
   end
 end
--- a/lib/rspec/mocks.rb
+++ b/lib/rspec/mocks.rb
@@ -101,18 +101,18 @@ module RSpec
       end
     end
 
-    @@excluded_subclasses = []
+    @excluded_subclasses = []
 
     def self.excluded_subclasses
-      @@excluded_subclasses.select(&:weakref_alive?).map do |ref|
+      @excluded_subclasses.select(&:weakref_alive?).map do |ref|
         ref.__getobj__
       rescue RefError
         nil
-      end
+      end.compact
     end
 
     def self.exclude_subclass(constant)
-      @@excluded_subclasses << WeakRef.new(constant)
+      @excluded_subclasses << WeakRef.new(constant)
     end
 
     module ExcludeClassesFromSubclasses
--- a/lib/rspec/mocks.rb
+++ b/lib/rspec/mocks.rb
@@ -101,6 +101,23 @@ module RSpec
       end
     end
 
+    @@excluded_subclasses = []
+
+    def self.excluded_subclasses
+      @@excluded_subclasses.select(&:weakref_alive?).map(&:__getobj__)
+    end
+
+    def self.exclude_subclass(constant)
+      @@excluded_subclasses << WeakRef.new(constant)
+    end
+
+    module ExcludeClassesFromSubclasses
+      def subclasses
+        super - RSpec::Mocks.excluded_subclasses
+      end
+    end
+    Class.prepend(ExcludeClassesFromSubclasses)
+
     class << self
       # @private
       attr_reader :space
--- a/lib/rspec/mocks.rb
+++ b/lib/rspec/mocks.rb
@@ -104,7 +104,11 @@ module RSpec
     @@excluded_subclasses = []
 
     def self.excluded_subclasses
-      @@excluded_subclasses.select(&:weakref_alive?).map(&:__getobj__)
+      @@excluded_subclasses.select(&:weakref_alive?).map do |ref|
+        ref.__getobj__
+      rescue RefError
+        nil
+      end
     end
 
     def self.exclude_subclass(constant)
--- a/lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
+++ b/lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
@@ -36,7 +36,7 @@ module RSpec
 
           @excluded_subclasses ||= []
           @excluded_subclasses.select(&:weakref_alive?).map do |ref|
-            begin # rubocop:disable Style/RedundantBegin
+            begin
               ref.__getobj__
             rescue ::WeakRef::RefError
               nil
--- a/lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
+++ b/lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
@@ -4,48 +4,51 @@ module RSpec
     #
     # @private
     class ExcludeStubbedClassesFromSubclasses
-      def self.enable!
-        return unless RUBY_VERSION >= "3.1"
-        return if Class.respond_to?(:subclasses_with_rspec_mocks)
+      class << self
+        def enable!
+          return unless RUBY_VERSION >= "3.1"
+          return if Class.respond_to?(:subclasses_with_rspec_mocks)
 
-        require 'weakref'
+          Class.class_eval do
+            def subclasses_with_rspec_mocks
+              subclasses_without_rspec_mocks - RSpec::Mocks::ExcludeStubbedClassesFromSubclasses.excluded_subclasses
+            end
 
-        mod_exclude_subclasses = Module.new do
-          def excluded_subclasses
-            @excluded_subclasses ||= []
-            @excluded_subclasses.select(&:weakref_alive?).map do |ref|
-              begin
-                ref.__getobj__
-              rescue ::WeakRef::RefError
-                nil
-              end
-            end.compact
-          end
-
-          def exclude_subclass(constant)
-            @excluded_subclasses ||= []
-            @excluded_subclasses << ::WeakRef.new(constant)
+            alias subclasses_without_rspec_mocks subclasses
+            alias subclasses subclasses_with_rspec_mocks
           end
         end
-        extend(mod_exclude_subclasses)
 
-        Class.class_eval do
-          def subclasses_with_rspec_mocks
-            subclasses_without_rspec_mocks - RSpec::Mocks::ExcludeStubbedClassesFromSubclasses.excluded_subclasses
+        def disable!
+          @excluded_subclasses = []
+
+          if Class.respond_to?(:subclasses_with_rspec_mocks)
+            Class.class_eval do
+              undef subclasses_with_rspec_mocks
+              alias subclasses subclasses_without_rspec_mocks # rubocop:disable Lint/DuplicateMethods
+              undef subclasses_without_rspec_mocks
+            end
           end
-
-          alias subclasses_without_rspec_mocks subclasses
-          alias subclasses subclasses_with_rspec_mocks
         end
-      end
 
-      def self.disable!
-        return unless Class.respond_to?(:subclasses_with_rspec_mocks)
+        def excluded_subclasses
+          require 'weakref' unless defined?(::WeakRef)
 
-        Class.class_eval do
-          undef subclasses_with_rspec_mocks
-          alias subclasses subclasses_without_rspec_mocks # rubocop:disable Lint/DuplicateMethods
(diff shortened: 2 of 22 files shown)


lib/rspec/mocks.rb
```diff
@@ -101,6 +101,23 @@ def self.with_temporary_scope
       end
     end
 
+    @@excluded_subclasses = []
+
+    def self.excluded_subclasses
+      @@excluded_subclasses.select(&:weakref_alive?).map(&:__getobj__)
```
pirj, line 107: This could be compacted to prevent endless growth.

lib/rspec/mocks.rb
```diff
@@ -101,6 +101,23 @@ def self.with_temporary_scope
       end
     end
 
+    @@excluded_subclasses = []
+
+    def self.excluded_subclasses
```
pirj, line 106: Is it necessary to getobj?
There’s a potential race condition, and the weakref itself should quack just like the object itself (including the === probably to allow to remove from the list.
https://stackoverflow.com/questions/69185508/ruby-weakref-has-implicit-race-condition

pirj (commented):
It’s nice that you’ve found a reliable solution.
I’m skeptical, however, to include it for everyone.
There might be some performance impact, and since this can introduce flakiness due to the GC race condition and an exception is not reassuring.

Still, this should be kept as a reference for those who experience issues with `subclasses`.

lib/rspec/mocks.rb
GCorbel: Yes, otherwise the substraction wasn't working.

I changed to rescue RefError.

lib/rspec/mocks.rb
GCorbel: I added `.compact`.

JonRowe (changes requested):
Just placing a block on this as I don't want it merged without final say so @pirj I'm concerned about the require.

lib/rspec/mocks.rb
```diff
@@ -101,6 +101,27 @@ def self.with_temporary_scope
       end
     end
 
+    @excluded_subclasses = []
+
+    def self.excluded_subclasses
+      @excluded_subclasses.select(&:weakref_alive?).map do |ref|
```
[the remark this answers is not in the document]
GCorbel: You mean with  ? 

```
diff --git a/lib/rspec/mocks.rb b/lib/rspec/mocks.rb
index a82055dd..ad697b59 100644
--- a/lib/rspec/mocks.rb
+++ b/lib/rspec/mocks.rb
@@ -104,11 +104,7 @@ module RSpec
     @excluded_subclasses = []
 
     def self.excluded_subclasses
-      @excluded_subclasses.select(&:weakref_alive?).map do |ref|
-        ref.__getobj__
-      rescue RefError
-        nil
-      end.compact
+      @excluded_subclasses
     end
 
     def self.exclude_subclass(constant)
@@ -117,7 +113,11 @@ module RSpec
 
     module ExcludeClassesFromSubclasses
       def subclasses
-        super - RSpec::Mocks.excluded_subclasses
+        super - RSpec::Mocks.excluded_subclasses.select(&:weakref_alive?).map do |ref|
+          ref.__getobj__
+        rescue RefError
+          nil
+        end.compact
       end
     end
     Class.prepend(ExcludeClassesFromSubclasses)
```

GCorbel:
> I’m skeptical, however, to include it for everyone. There might be some performance impact, and since this can introduce flakiness due to the GC race condition and an exception is not reassuring.

I understand your worries. I would like to find a better solution but wasn't able. 

Maybe we can use a config as `excluded_stubbed_const_from_subclasses` or enable it by default and do a config as `thread_safe_garabage_collection` ?

GCorbel:
The more I think, the more I'm in favor of a config flag. Maybe a warning message that say the flag have to be switched should be displayed when there is an error.

pirj:
I support your idea of opting in for this with a flag.

pirj:
We could conditionally require weakref if this option is on.

GCorbel:
Should we have the option on by default ?

pirj:
No. But we’re working in RSpec 4, and we can think about enabling it there unless performance considerations come up.

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
GCorbel, line 1: I took `RSpec::Mocks::MarshalExtension` for example.

GCorbel (commented):
I refactored, added a config flag and specs. I tried to respect the existing code. Let me know what you think.

GCorbel:
I change to run specs only for Ruby 3.1 and ahead as in https://github.com/rails/rails/blob/v7.0.8/activesupport/lib/active_support/ruby_features.rb

GCorbel:
I fixed the code for Ruby <= 2.4

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
```diff
@@ -0,0 +1,53 @@
...
+          alias subclasses subclasses_with_rspec_mocks
+        end
+      end
+
+      def self.disable!
+        return unless Class.respond_to?(:subclasses_with_rspec_mocks)
+
+        Class.class_eval do
+          undef subclasses_with_rspec_mocks
+          alias subclasses subclasses_without_rspec_mocks
```
GCorbel: Rubocop triggers this warning : 

```
lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb:47:11: W: Lint/DuplicateMethods: Method RSpec::Mocks::ExcludeStubbedClassesFromSubclasses::Class#subclasses is defined at both lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb:38 and lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb:47.
          alias subclasses subclasses_without_rspec_mocks
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

It's wrong because the method is undefined at line 46. I disabled this warning.

GCorbel:
The CI pass except for "ruby-head" which is not to my commit

GCorbel:
I fixed a flap on specs. Now the CI should pass.

GCorbel:
I finally succeeded to install Ruby 3.4-dev locally and confirm that the error on the CI is not related to the PR. It also fail on the main branch.

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
```diff
@@ -0,0 +1,53 @@
...
+    #
+    # @private
+    class ExcludeStubbedClassesFromSubclasses
+      def self.enable!
+        return unless RUBY_VERSION >= "3.1"
+        return if Class.respond_to?(:subclasses_with_rspec_mocks)
+
+        require 'weakref'
+
+        mod_something = Module.new do
```
pirj, line 13: Is “something” intentional wording?

lib/rspec/mocks.rb
```diff
@@ -113,10 +113,11 @@ class << self
 
     # To speed up boot time a bit, delay loading optional or rarely
     # used features until their first use.
-    autoload :AnyInstance,      "rspec/mocks/any_instance"
-    autoload :ExpectChain,      "rspec/mocks/message_chain"
-    autoload :StubChain,        "rspec/mocks/message_chain"
-    autoload :MarshalExtension, "rspec/mocks/marshal_extension"
```
pirj, line 119: I recall we wanted to adjust our style not to align by default, to avoid churn. I’d rather have inconsistently aligned neighbour lines, but no pressure.

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
```diff
@@ -0,0 +1,53 @@
...
+              end
+            end.compact
+          end
+
+          def exclude_subclass(constant)
+            @excluded_subclasses ||= []
+            @excluded_subclasses << ::WeakRef.new(constant)
+          end
+        end
+        RSpec::Mocks.extend(mod_something)
```
pirj, line 30: Will we keep extending on every `enable!` (intermingled with disable!)?

spec/rspec/mocks/exclude_stubbed_classes_from_subclasses_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+          described_class.disable!
+        end
+
+        describe '.enable!' do
+          it 'does not extends Class when it has been enabled' do
+            described_class.enable!
+
+            expect {
+              described_class.enable!
+            }.not_to(change { Class.respond_to?(:subclasses_with_rspec_mocks) })
```
pirj, line 18: I don’t quite understand this. We enable the e feature, the Class is altered, and we expect the second call to enable not to … remove the method added by previous enable?

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
pirj, line 30: On another thought, can we use this class to keep the list instead of RSpec::Mocks?

spec/rspec/mocks/exclude_stubbed_classes_from_subclasses_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+          it 'excludes stubbed classes from subclasses' do
+            ::RSpec::Mocks.space.reset_all
+            RSpec::Mocks.configuration.exclude_stubbed_classes_from_subclasses = true
+
+            subclass = Class.new(TestClass)
+            stub_const('TestSubClass', subclass)
+
+            ::RSpec::Mocks.space.reset_all
+            expect(TestClass.subclasses.map(&:object_id)).not_to include(subclass.object_id)
```
pirj, line 40: Is there a reason to write this differently from how the opposite is written below?

expect(TestClass.subclasses).to include(subclass)

pirj (commented):
Please accept my apologies for the wait. The code is mostly good, thanks!

I’d love to hear what @JonRowe says.
Do we need a benchmark to measure the overhead in `reset` calls?

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
GCorbel: No, it wasn't. I changed.

lib/rspec/mocks.rb
GCorbel: I changed and applied the rule for other lines in this file.

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
GCorbel: No, if we call `enable!` multiple times, it extends only once because of `return if Class.respond_to?(:subclasses_with_rspec_mocks)`.

spec/rspec/mocks/exclude_stubbed_classes_from_subclasses_spec.rb
GCorbel: I agree this wasn't making sense. I changed.

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
GCorbel: I agree and changed for that.

spec/rspec/mocks/exclude_stubbed_classes_from_subclasses_spec.rb
GCorbel: I changed to be similar to other tests.

GCorbel:
> Please accept my apologies for the wait. The code is mostly good, thanks!

No problem, it's open source. You don't even have to reply.

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
pirj: The problem I’m thinking of is when you enable!, disable!, and then enable! again when Class is not augmented, but the ExcludeStubbed… is.
Do you see what I mean?
You could add a described_class.disable! in between two enable! calls in the corresponding spec, and see how the `.once` expectation fails. I did not test this, just an assumption. I’ll be happy if it’s just my imagination.

What if we just define those two methods on the ExcludeStubbed immediately and don’t ‘extend’ it at all?

GCorbel: ok, my bad. I take a look.

GCorbel: I refactored the code. Doing and `extend` was now useless. I also changed and added missing specs.

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
```diff
@@ -0,0 +1,56 @@
...
+            rescue ::WeakRef::RefError
+              nil
+            end
+          end.compact
+        end
+
+        def exclude_subclass(constant)
+          require 'weakref' unless defined?(::WeakRef)
+
+          @excluded_subclasses ||= []
```
pirj, line 50: This is in a few places. Might be it belongs to `enable!`, too?

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
```diff
@@ -0,0 +1,56 @@
...
+            Class.class_eval do
+              undef subclasses_with_rspec_mocks
+              alias subclasses subclasses_without_rspec_mocks
+              undef subclasses_without_rspec_mocks
+            end
+          end
+        end
+
+        def excluded_subclasses
+          require 'weakref' unless defined?(::WeakRef)
```
pirj, line 35: Ok to move this line into `enable!`?

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
```diff
@@ -0,0 +1,56 @@
...
+
+          @excluded_subclasses ||= []
+          @excluded_subclasses.select(&:weakref_alive?).map do |ref|
+            begin # rubocop:disable Style/RedundantBegin
```
pirj, line 39: It slipped my mind, isn’t it possible to squash that so that rescue is on the map itself?

pirj: Ah, nevermind, Ruby 1.8 probably doesn’t support that. Do you mind making a note to indicate that?

pirj (approved):
Looks solid, thank you! 🙌 

Just a few optional, subjective and cosmetic notes - please feel free to ignore.

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
GCorbel: I'm not sure if it's a problem but now we can do `RSpec::Mocks::ExcludeStubbedClassesFromSubclasses.excluded_subclasses` without `ExcludeStubbedClassesFromSubclasses.enable!` first. If we move this line to `enable!`, it will raise an error.

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
GCorbel: I removed the comment because the CI was failing because of that (I don't have the same behavior locally with the same version of Ruby/Rubocop).

lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
GCorbel: As you wish, I can do : 


```
diff --git a/lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb b/lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
index bb72101e..129ff0d2 100644
--- a/lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
+++ b/lib/rspec/mocks/exclude_stubbed_classes_from_subclasses.rb
@@ -8,6 +8,8 @@ module RSpec
         def enable!
           return unless RUBY_VERSION >= "3.1"
           return if Class.respond_to?(:subclasses_with_rspec_mocks)
+          @excluded_subclasses = []
+          require 'weakref' unless defined?(::WeakRef)
 
           Class.class_eval do
             def subclasses_with_rspec_mocks
@@ -32,9 +34,6 @@ module RSpec
         end
 
         def excluded_subclasses
-          require 'weakref' unless defined?(::WeakRef)
-
-          @excluded_subclasses ||= []
           @excluded_subclasses.select(&:weakref_alive?).map do |ref|
             begin
               ref.__getobj__
@@ -45,9 +44,6 @@ module RSpec
         end
 
         def exclude_subclass(constant)
-          require 'weakref' unless defined?(::WeakRef)
-
-          @excluded_subclasses ||= []
           @excluded_subclasses << ::WeakRef.new(constant)
         end
       end
diff --git a/spec/rspec/mocks/exclude_stubbed_classes_from_subclasses_spec.rb b/spec/rspec/mocks/exclude_stubbed_classes_from_subclasses_spec.rb
index 3b5ded94..16454922 100644
--- a/spec/rspec/mocks/exclude_stubbed_classes_from_subclasses_spec.rb
+++ b/spec/rspec/mocks/exclude_stubbed_classes_from_subclasses_spec.rb
@@ -63,6 +63,10 @@ if RUBY_VERSION >= '3.1'
         end
 
         describe '.excluded_subclasses' do
+          before do
+            described_class.enable!
+          end
+
           it 'returns excluded subclasses' do
             subclass = Class.new
             described_class.exclude_subclass(subclass)
@@ -75,14 +79,12 @@ if RUBY_VERSION >= '3.1'
             described_class.exclude_subclass(subclass)
 
             subclass = nil
-
             GC.start
 
             expect(described_class.excluded_subclasses).to eq([])
           end
 
           it 'does not return excluded subclasses that raises a ::WeakRef::RefError' do
-            require 'weakref'
             subclass = double(:weakref_alive? => true)
             described_class.instance_variable_set(:@excluded_subclasses, [subclass])
 
```

GCorbel:
@JonRowe can you give a look at this request ?

JonRowe:
I'd like to apologise for leaving this hanging for so long, its been on my review list for ages but I just haven't been able to get around to it, I had a couple of concerns about weak ref and other such behaviour but following on from the discussion in #1568 I'm going to close this. Sorry to have wasted your time.

PR #1572, opened 2024-04-14
Fix and_wrap_original failing with ruby 3.2

Geesu:
This spec only fails in ruby 3.2, it succeeds with ruby 3.1. The failure in 3.2:
```
     ArgumentError:
       wrong number of arguments (given 1, expected 0; required keyword: my_object)
     Shared Example Group: "using and_wrap_original" called from ./spec/rspec/mocks/and_wrap_original_spec.rb:121
     # ./spec/rspec/mocks/and_wrap_original_spec.rb:27:in `initialize'
```


Geesu:
Closing as my spec was incorrect, it should have been:

`allow(MyService).to receive(:call_now).and_wrap_original do |blk, **args|`

PR #1581, opened 2024-07-13
Fix build for Ruby 3.4.0 preview 1

JonRowe:
Ruby head is currently Ruby 3.4 and has a couple of changes which issue warnings this PR:

- Removes some warnings about block paramters being unused, even though ironically, they are
- Relaxes some checks around formatting


JonRowe:
Cucumber is failing on head potentially because of hash changes? I'm going to merge this and I will probably add a 3.4.0-preview1 matrix entry

PR #1586, opened 2024-08-19
Bug: Classes don't follow the double docs WRT #dup

jcoleman:
While mocks don't get copied on an object that's dup'd, that is not true for classes.


JonRowe:
Migrated to the monorepo

Issue #1592, opened 2024-09-12
An unexpected args error confuses minitest when it looks for the source line


onlynone:
### Subject of the issue

We're using `rspec-expectations`, `rspec-mocks` with `minitest`. We're using the `minitest_integration` with both of them. For the most part everything works well. But sometimes failures cause minitest's location logic to not find where the problem really occurred.

For example, we have a test like:

```
class ExampleTest
  test "something" do
    expect(a_real_object).to receive(:a_method).with("these", "args")
    do_something
  end
end
```

And if `do_something` eventually results in a call to `a_real_object.a_method`, but with different arguments. We'll get failure output like:

```
Failure:
ExampleTest#test_something [/Users/user/.gem/ruby/2.7.8/gems/rspec-mocks-3.12.5/lib/rspec/mocks/proxy.rb:217]:
```

But that source location isn't very helpful. I traced it down to the way [minitest looks for a probable failure line](https://github.com/minitest/minitest/blob/v5.11.3/lib/minitest.rb#L823). It looks at the backtrace of the exception and grabs the line just after a line matching the regex `/in .(assert|refute|flunk|pass|fail|raise|must|wont)/` (note, this is based on the version of minitest I'm using, but the most recent version does essentially the same thing).

And the backtrace looks like:

```
"/Users/user/.gem/ruby/2.7.8/gems/rspec-support-3.12.0/lib/rspec/support.rb:102:in `block in <module:Support>'",
"/Users/user/.gem/ruby/2.7.8/gems/rspec-support-3.12.0/lib/rspec/support.rb:111:in `notify_failure'",
"/Users/user/.gem/ruby/2.7.8/gems/rspec-mocks-3.12.5/lib/rspec/mocks/error_generator.rb:348:in `notify'",
"/Users/user/.gem/ruby/2.7.8/gems/rspec-mocks-3.12.5/lib/rspec/mocks/error_generator.rb:332:in `__raise'",
"/Users/user/.gem/ruby/2.7.8/gems/rspec-mocks-3.12.5/lib/rspec/mocks/error_generator.rb:55:in `raise_unexpected_message_args_error'",
"/Users/user/.gem/ruby/2.7.8/gems/rspec-mocks-3.12.5/lib/rspec/mocks/message_expectation.rb:555:in `raise_unexpected_message_args_error'",
[...]

PR #1593, opened 2024-09-17
Ensure backtraces start at user code, not rspec-mocks code

onlynone:
This will ensure all exceptions generated by rspec-mocks have the first line of their backtraces pointing to the user's code where the error happened.

Addresses #1592


pirj:
HunkGenerator on 2.2 and 2.5 is flaky.

pirj:
Do you need some help with JRuby support?

JonRowe:
We already handle this in rspec-core and as per #1592 I don't want to add the maintenance burden of this, it would duplicate logic we have elsewhere.

